PR #8, opened 2023-04-06
try condition lint on diff

rboston628:
This will add a job to check that the .py files were updated, and will only run the linters and pytest in the case that some .py files were updated.

If no .py files differ from branch `next`, then neither the linter nor pytest will run.

If any .py files do differ from branch `next`, then the linter will run on only those changed files, and the pytests will run.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -20,7 +20,7 @@ jobs:
 
   lint:
     needs: [check-lint]
-    if: check-lint.outputs.OUTPUT == 1
+    if: jobs.check-lint.outputs.OUTPUT == 1
     runs-on: ubuntu-latest
     defaults:
       run:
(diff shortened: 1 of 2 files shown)


.github/workflows/actions.yml
```diff
@@ -27,23 +41,21 @@ jobs:
...
+        black ${{ needs.check-lint.outputs.output1 }} --check
     
     - name: Type-check with myy
       run: |
         conda install mypy
         mypy src/
 
   tests:
+    needs: [check-lint]
+    if: needs.check-lint.outputs.output1 != ''
```
walshmm, line 58: Not sure we want to prevent tests from running, what if a configuration file changed that caused a failure?

peterfpeterson:
Why don't you just install pre-commit (in your `environment.yml`) and run `pre-commit run --all`? It'd be less moving parts to synchronize as far as configuration.

rboston628:
> Why don't you just install pre-commit (in your `environment.yml`) and run `pre-commit run --all`? It'd be less moving parts to synchronize as far as configuration.

We considered it, weren't sure how it handled diffs when run on github, so went with resolving the diff by hand

walshmm:
Closed in favor of pre-commit.ci

PR #12, opened 2023-04-14
Conda build

rboston628:
Setup actions for automatic build of conda package


conda.recipe/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - versioningit
+    - pydantic
+    - mantidworkbench
+    - qtpy
+    - pytest
+
+  run:
+    - pydantic
+    - mantidworkbench
+    - qtpy
```
peterfpeterson, line 43: mantidworkbench requires qtpy and you shouldn't include pytest as a dependency for the person installing your package

rboston628:
The git tree is hopelessly tangled, rebasing cannot fix it.

PR #18, opened 2023-04-20
Fix deployments and wheel

walshmm:


src/snapred/main.py
```diff
@@ -46,34 +40,37 @@ def __init__(self, parent):
...
                     background-color: {};
                 }}
                 QToolButton:hover {{
                     background-color: {}
                 }}
-            '''.format(colorNormal, colorHover))
+            """.format(
+                    colorNormal, colorHover
+                )
+            )
```
rboston628, line 68: I know this is trivial, but could we use an f-string?
```
            btn.setStyleSheet(
                f"""
                QToolButton {{
                    background-color: {colorNormal};
                }}
                QToolButton:hover {{
                    background-color: {colorHover)}
                }}
            '''
```
Also, why does one end in a `;` and the other doesn't?

rboston628 (approved):

PR #23, opened 2023-05-04
change ServiceFactory to expect delimited path

walshmm:
This will allow us to specify specific actions in each service, so that its not restricted to just executing a recipe.

Ex. The calibration service would like to save its calibration it just created upon the user approving it, the ui would then specify
`calibration/recipe` to execute the initial recipe
`calibration/save` to save the current result to the calibration index.


walshmm:
[moving debugging of pytest githubactions segfault elsewhere]

rboston628 (approved):

PR #24, opened 2023-05-08
set pytest to generate tests

walshmm:


rboston628 (approved):
Looks like these caused tests to pass

PR #27, opened 2023-05-10
new crystalographic DAO and test

rboston628:


walshmm:
A few things:
- probably dont need to pr every task if youre the only one working on a story
- give your branches a descriptive name
- generally dont need to write tests for the dao package (I should really add them to the exclusion list).  A test for the custom validator would make sense, else we are just testing pydantic works.

PR #28, opened 2023-05-11
Ingest Crystallographic Info

rboston628:
See EWM 774


environment.yml
```diff
@@ -20,6 +20,7 @@ dependencies:
 - setuptools
 - python-build
 - boa
+- pycifrw
```
walshmm, line 23: where is this used? ctrl+f doesnt reveal anything

rboston628: This is needed in order for LoadCIF to work.  See output on failing test here: https://github.com/neutrons/SNAPRed/actions/runs/4995107184/jobs/8946555614#step:6:161

walshmm: sounds good

src/snapred/backend/service/CrystallographicInfoService.py
```diff
@@ -0,0 +1,29 @@
+from typing import Any, Dict
+
+from snapred.backend.data.DataFactoryService import DataFactoryService
+from snapred.backend.recipe.CrystallographicInfoRecipe import CrystallographicInfoRecipe
+from snapred.meta.Singleton import Singleton
+
+
+@Singleton
+class IngestionService:
```
walshmm, line 9: Could you refactor this to be in line with the new way we do services?
i.e. rename the orchestrate method, decorate it with FromString, and register the former orchestrate method in init to "", and have IngestionService inherit from Service

tests/unit/backend/service/test_ExtractionService.py
rboston628, line 1: This test didn't do anything, so I removed it

PR #40, opened 2023-06-05
patch to crystallographic info, d^4

rboston628:
Includes the Lorentz factor d^4 in calculation of intensity.

Also refactors CrystallographicInfo as discussed with Michael


walshmm (approved):

walshmm (changes requested):
Tests need to be updated, not the biggest fan of the thin wrapper but I get its to minimize refactoring.

PR #48, opened 2023-06-26
Add callbacks to mantid snapper

walshmm:
Pr is a little contaminated with innocuous alignandfocuspowder changes

### DIFF
--- a/src/snapred/backend/recipe/algorithm/AlignAndFocusReductionAlgorithm.py
+++ b/src/snapred/backend/recipe/algorithm/AlignAndFocusReductionAlgorithm.py
@@ -16,7 +16,7 @@ class AlignAndFocusReductionAlgorithm(PythonAlgorithm):
         self.mantidSnapper = MantidSnapper(self, name)
 
     def PyExec(self):
-        reductionIngredients = ReductionIngredients.parse_raw(self.getProperty("ReductionIngredients"))
+        reductionIngredients = ReductionIngredients.parse_raw(self.getProperty("ReductionIngredients").value)
         # run the algo
         self.log().notice("Execution of AlignAndFocusReductionAlgorithm START!")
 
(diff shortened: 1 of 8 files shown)


tests/unit/backend/recipe/algorithm/test_MantidSnapper.py
```diff
@@ -18,7 +20,7 @@ def test_snapper_fake_algo():
         return_of_algo = "return of algo"
         mantidSnapper = MantidSnapper(parentAlgorithm=None, name="")
         test = mantidSnapper.fakeFunction("test", fakeOutput=return_of_algo)
-        assert test == return_of_algo
+        assert str(test.__class__) == str(callback(return_of_algo.__class__).__class__)
```
rboston628, line 23: I think this test is less stringent than the old one.  The old one ensured the value was the same, whereas this is apparently checking that a callback is returned.

src/snapred/backend/recipe/algorithm/AlignAndFocusReductionAlgorithm.py
```diff
@@ -29,15 +29,21 @@ def PyExec(self):
         # raw_data = self.loadEventNexus(Filename=rawDataPath, OutputWorkspace="raw_data")
         self.mantidSnapper.LoadNexus("loading vanadium file...", Filename=vanadiumFilePath, OutputWorkspace="vanadium")
 
-        self.mantidSnapper.CreateGroupWorkspace(
+        self.mantidSnapper.CustomGroupWorkspace(
```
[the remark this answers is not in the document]
walshmm: I think it was a typo, `CreateGroupWorkspace` is not an algo.
for ref: https://docs.mantidproject.org/nightly/algorithms/categories/AlgorithmIndex.html

tests/unit/backend/recipe/algorithm/test_MantidSnapper.py
walshmm: Thats because an unpopulated callback is now returned instead.  This is mostly so all types are treated uniformly as opposed to workspace names being an exception. 

In order to actually test the value, the algo would have to run and populate that callback.

rboston628 (approved):

PR #56, opened 2023-07-11
Extract DetectorPeakPredictor algorithm. Add peak tail coefficient.

bolotovskyr:
In accordance with the DRY principle, this PR creates a reusable algorithm for detector peak prediction. Correspondingly, `PurgeOverlappingPeaksAlgorithm` and `FitMultiplePeaksAlgorithm` have been refactored to use the new peak prediction code. This PR also takes care of [Story 1930](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=1930).


src/snapred/backend/dao/DetectorPeak.py
```diff
@@ -0,0 +1,9 @@
+from pydantic import BaseModel
+
+
+class DetectorPeak(BaseModel):
+    """Class to hold diffraction peak position and limits in d-spacing"""
+
+    position: float
+    limLeft: float
+    limRight: float
```
walshmm, line 9: Id replace these two with a `Limit[float]`

src/snapred/backend/recipe/algorithm/FitMultiplePeaksAlgorithm.py
```diff
@@ -28,22 +29,22 @@ def PyExec(self):
...
         peakType = fitPeakIngredients.PeakType
 
-        beta_0 = instrumentState.gsasParameters.beta[0]
-        beta_1 = instrumentState.gsasParameters.beta[1]
-        FWHMMultiplierLeft = instrumentState.fwhmMultiplierLimit.minimum
-        FWHMMultiplierRight = instrumentState.fwhmMultiplierLimit.maximum
-        L = instrumentState.instrumentConfig.L1 + instrumentState.instrumentConfig.L2
+        instrumentState.gsasParameters.beta[0]
+        instrumentState.gsasParameters.beta[1]
+        instrumentState.instrumentConfig.L1 + instrumentState.instrumentConfig.L2
```
walshmm, line 34: if these values are not used please remove them.  it looks like the formatter removed the assignments which is indicative of this.

src/snapred/backend/dao/DetectorPeak.py
bolotovskyr: Yes, I thought about it.  Actually there is more to the story: we want `limLeft <= position <= limRight`.  So maybe we need an extended version of the `Limit` class?

walshmm: well, these variables would become `positionLimit`.  What's the additional usecase?  You can add a validator to this object to ensure that position satisfies this requirement. [See this example.](https://github.com/neutrons/SNAPRed/blob/db74e5964e9b2527d662d7277cdd64a56c474b21/src/snapred/backend/dao/FitMultiplePeaksIngredients.py#L2)

walshmm: Which I suppose you could then extract out to some sorted of `BoundedValue[float]` generic object

src/snapred/backend/recipe/algorithm/FitMultiplePeaksAlgorithm.py
bolotovskyr: Done.

src/snapred/backend/dao/DetectorPeak.py
bolotovskyr: Done.

PR #58, opened 2023-07-18
Use MantidSnapper to call nested DetectorPeakPredictor algorithm

bolotovskyr:
Adding missing `setChild(True)` calls - a leftover from PR #56.
To test:
```
pytest -s tests/unit/backend/recipe/algorithm/test_FitMultiplePeaksAlgorithm.py
pytest -s tests/unit/backend/recipe/algorithm/test_PurgeOverlappingPeaksAlgorithm.py
```


src/snapred/backend/recipe/algorithm/DiffractionSpectrumWeightCalculator.py
```diff
@@ -5,7 +5,6 @@
 from mantid.kernel import Direction
 
 from snapred.backend.dao.DetectorPeak import DetectorPeak
-from snapred.backend.recipe.algorithm.DetectorPeakPredictor import DetectorPeakPredictor
```
walshmm, line 8: Is this imported somewhere else?  Otherwise it may not get registered with Mantid.  I suggest just adding a [GenericRecipe](https://github.com/neutrons/SNAPRed/blob/next/src/snapred/backend/recipe/GenericRecipe.py#L55) for it so the linter doesnt remove it.  At least until we get Mantid to inspect our Algo folder.

walshmm: err though I'm not sure GenericRecipe is used in next (I did some refactoring in my other branch) yet so might still have an import issue there

walshmm: the other option is to add the import back somewhere and commit with the `-n` flag to skip the linting hooks

walshmm (approved):

PR #63, opened 2023-07-20
Add test for starting GUI in CI

peterfpeterson:
A couple of very big items
* Add `argparse` with various options including `--headcheck` which is intended for use in CI
* Default disable mantid's network capabilities (can be turned on with command line options)
* Clarify error messages associated with misconfigured system settings and missing files
* Remove the registration of `ExtractionService` so the GUI can be started

~~Do not merge before #64~~

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -8,3 +8,28 @@ Installation
 
 
 Under Construction!
+
+Application settings
+--------------------
+
+Settings for the application are shipped with the package in the `application.yml <https://github.com/neutrons/SNAPRed/blob/next/src/snapred/resources/application.yml>`_ file.
+There is also a `test application.yml <https://github.com/neutrons/SNAPRed/blob/next/tests/resources/application.yml>`_ file used in tests.
+Most of what is included in the settings is a parameterization of options for the instrument.
+It may be desireable to override some of the settings which can be done via environment injection.
+For example
+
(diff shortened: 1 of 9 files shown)


peterfpeterson:
Do not merge this yet

docs/source/usage.rst
```diff
@@ -8,3 +8,28 @@ Installation
...
+   instrument:
+     home: ~/snapred-testing/
+
+Can be used to override the ``instrument.home`` directory to ``~/snapred-testing/``.
+At this time, only the ``instrument.home`` and ``samples.home`` directories will expand ``~/`` to a directory.
+This file can be supplied via command line injection
+
+.. code-block:: sh
+
+   env=myenv.yml snapred
```
walshmm, line 35: Sorry for not catching this sooner, but why is the `env` including the file extension?  The environment file is derived from the environment and will always be a `.yml`.  Not that we have other things dependent on `env` right now, but potentially.

peterfpeterson: It can also be called as `env=myenv snapred` to the same effect. The previous code always added on the `.yml` extension and looked into the resource directory. This way allows for a varient of specifying a relative or absolute file path.

walshmm: Ah ok sounds good.

walshmm (approved):

PR #64, opened 2023-07-20
Modify values in test.yml to reduce mock object use

peterfpeterson:
By modifying the settings, there is no need to side-step the LocalDataService in the tests. It is now correctly configured to use information that are in the repository.

The important thing to look at is the change to `test.yml`.

### DIFF
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -3,6 +3,8 @@ import glob
 import hashlib
 import json
 import os
+from errno import ENOENT as NOT_FOUND
+from pathlib import Path
 from typing import Any, Dict, List, Tuple
 
 import h5py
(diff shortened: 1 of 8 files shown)


src/snapred/backend/data/LocalDataService.py
```diff
@@ -35,35 +37,58 @@
...
     reductionParameterCache: Dict[str, Any] = {}
     iptsCache: Dict[str, Any] = {}
     stateIdCache: Dict[str, str] = {}
-    dataPath = Config["instrument.home"]
-    instrumentConfigPath: str = dataPath + Config["instrument.config"]
     instrumentConfig: InstrumentConfig  # Optional[InstrumentConfig]
 
-    def __init__(self) -> None:
-        self.instrumentConfig = self.readInstrumentConfig()
+    def __init__(self, verifyPaths: bool = True) -> None:
```
walshmm, line 51: I would change this verifyPath to a value on the application.yml and set it to false in test.yml

walshmm (approved):

PR #67, opened 2023-07-24
Update calculation of pixel grouping dependent parameters

bolotovskyr:
- Update the math used for calculation of pixel grouping-dependent parameters: [Story 1929](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=1929)
- Reduce the algorithm input level from `CalibrationState` to `InstrumentState`, b/c passing the whole calibration state is unnecessary
- In the testing script, use a helper method for getting the IDF path. For the full instrument, use `SNAP_Definition.xml` instead of `SNAP_Definition_2011-09-07.xml`. The latter file isn't valid as of 2018 and apparently causes a reverse group ordering problem
- Remove the code bits that reverse group ordering when testing calculation results against reference data.  
- Change `json` string variable suffixes from `_str(s)` to `_json`.

### DIFF
--- a/src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
+++ b/src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
@@ -157,8 +157,14 @@ class PixelGroupingParametersCalculationAlgorithm(PythonAlgorithm):
                 "Loading instrument...", Workspace=ws_name, FileName=idf, RewriteSpectraMap=False
             )
         else:  # full instrument
+            # self.mantidSnapper.LoadInstrument(
+            #     "Loading instrument...", Workspace=ws_name, InstrumentName="SNAP", RewriteSpectraMap="False"
+            # )
             self.mantidSnapper.LoadInstrument(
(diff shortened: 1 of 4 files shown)


src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
```diff
@@ -130,9 +152,20 @@ def LoadSNAPInstrument(self, ws_name):
...
+                "Loading instrument...", Workspace=ws_name, FileName=idf, RewriteSpectraMap=False
+            )
+        else:  # full instrument
+            # self.mantidSnapper.LoadInstrument(
+            #     "Loading instrument...", Workspace=ws_name, InstrumentName="SNAP", RewriteSpectraMap="False"
+            # )
+            self.mantidSnapper.LoadInstrument(
+                "Loading instrument...",
+                Workspace=ws_name,
+                FileName="/opt/anaconda/envs/mantid-dev/instrument/SNAP_Definition.xml",
```
rboston628, line 166: Did we solve the issue with this definition being outdated?

rboston628 (dismissed):

rboston628 (approved):

PR #70, opened 2023-07-25
Import bugfix

peterfpeterson:
This fixes the issue reported by Malcom trying to use snapred by importing rather than from the GUI. The "simple" test to run is
```py
from snapred.backend.api.InterfaceController import InterfaceController
from snapred.backend.dao.SNAPRequest import SNAPRequest

request = SNAPRequest(path="/api")
interfaceController = InterfaceController()
result = interfaceController.executeRequest(request=request)
print('code:', result.responseCode)
assert result.responseCode == 200
print('data:', result.responseData)
assert result.responseData is not None
```

### DIFF
--- a/src/snapred/backend/api/InterfaceController.py
+++ b/src/snapred/backend/api/InterfaceController.py
@@ -19,13 +19,8 @@ class InterfaceController:
         # execute the recipe
         # return the result
         try:
-            # ServiceFactory doesn't always return an instance so it needs to be fixed here
+            # leaving this a separate line makes stack traces make more sense
             service = self.serviceFactory.getService(request.path)
-            if inspect.isclass(service):
-                try:
-                    service = service()
-                except Exception as e:  # noqa BLE001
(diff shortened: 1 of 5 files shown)


src/snapred/backend/api/InterfaceController.py
```diff
@@ -1,33 +1,39 @@
...
 
     def executeRequest(self, request: SNAPRequest) -> SNAPResponse:
-        result = None
-        response = None
         # execute the recipe
         # return the result
         try:
-            result = self.serviceFactory.getService(request.path).orchestrateRecipe(request)
-
+            # ServiceFactory doesn't always return an instance so it needs to be fixed here
```
walshmm, line 22: I think the instantiation code should get extracted from ServiceDirectory to a class method and then propagated to its [get method](https://github.com/neutrons/SNAPRed/blob/next/src/snapred/backend/service/ServiceDirectory.py#L33)

walshmm (approved):

PR #72, opened 2023-07-27
Diffraction calibration on grouped spectra with log binning

rboston628:
For [EWM1965](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=1965)


src/snapred/backend/dao/state/PixelGroupingParameters.py
```diff
@@ -4,6 +4,7 @@
 
 
 class PixelGroupingParameters(BaseModel):
+    groupID: int
```
bolotovskyr, line 7: From the design standpoint, I am concerned that `groupID` is going to be "grouped" in one class with physical parameters: `dmin`/`dmax`, `2Theta`.  @rboston628, I thought you were going to keep those things separate, i.e. create a dictionary with the group ID being the key and the physical parameters being the value.

rboston628:
The failing test is because the recent changes to `PDCalibration` haven't bubbled up to the CI yet.
See [mantid PR 35837](https://github.com/mantidproject/mantid/pull/35837)

src/snapred/backend/dao/state/PixelGroupingParameters.py
rboston628: I can see if there's a way to avoid grouping the group ID in with the physical parameters.  Following on of @walshmm 's comments, we don't want to specify the group ID as a value to a key, i.e. `"groupID": 2` instead of just `2`.  I'll look around for another way to include it, that doesn't require yet another DAO.

src/snapred/backend/recipe/algorithm/FitMultiplePeaksAlgorithm.py
```diff
@@ -35,28 +36,34 @@ def PyExec(self):
...
             peakCenters = []
             peakLimits = []
-            for peak in reducedList[index]:
+            for peak in reducedList[groupID]:
                 peakCenters.append(peak.position.value)
                 peakLimits.extend([peak.position.minimum, peak.position.maximum])
 
             self.mantidSnapper.ExtractSingleSpectrum(
-                "Extract Single Spectrm...", InputWorkspace=wsName, OutputWorkspace="ws2fit", WorkspaceIndex=index
+                "Extract Single Spectrm...",
```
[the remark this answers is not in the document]
rboston628: got it, thanks

src/snapred/backend/dao/state/PixelGroupingParameters.py
bolotovskyr: Please also see @walshmm's earlier comment on [Story 1532](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=1532): "So there will be a map of (group,data) pairs.."

src/snapred/backend/dao/GroupPeakList.py
```diff
@@ -0,0 +1,13 @@
...
+
+from snapred.backend.dao.DetectorPeak import DetectorPeak
+
+
+class GroupPeakList(BaseModel):
+    """Class to hold diffraction peak positions of one group"""
+
+    peaks: List[DetectorPeak]
+    groupID: int
+    spectrumNumber: Optional[int]
```
bolotovskyr, line 13: After grouping, there is just one spectrum per group, correct? Then why do we need both `groupID` and `spectrumNumber`? Perhaps a comment would help.

src/snapred/backend/recipe/algorithm/GroupByGroupCalibration.py
```diff
@@ -0,0 +1,257 @@
...
+        if self.isLite:
+            self.outputFilename = (
+                f'{self.stateFolder}SNAP{self.runNumber}_calib_geom_{date.today().strftime("%Y%m%d")}.lite.h5'
+            )
+        else:
+            self.outputFilename = (
+                f'{self.stateFolder}SNAP{self.runNumber}_calib_geom_{date.today().strftime("%Y%m%d")}.h5'
+            )
+
+        # fdrom the instrument state, read the overall min/max TOF
```
bolotovskyr, line 48: from

src/snapred/backend/recipe/algorithm/GroupByGroupCalibration.py
```diff
@@ -0,0 +1,257 @@
...
+        self.TOFMin: float = ingredients.instrumentState.particleBounds.tof.minimum
+        self.TOFMax: float = ingredients.instrumentState.particleBounds.tof.maximum
+        self.TOFBin: float = -0.001
```
bolotovskyr, line 51: A comment would help as to why -0.001 value was chosen and why it's negative.

src/snapred/backend/dao/GroupPeakList.py
rboston628: My understanding is that `spectrumNumber` and `groupID` are not guaranteed to be related to one another.  But actually, I don't think we even need `spectrumNumber`, since the algorithms never call for it

src/snapred/backend/recipe/algorithm/GroupByGroupCalibration.py
rboston628: It is negative as a flag to `Rebin` to use logarithmic binning.  I'm not sure why 0.001, except that it was hard-coded in Malcolm's script and there does not seem to be any other data object with it.  @mguthriem

mguthriem: The choice of 0.001 is somewhat arbitrary and should be considered preliminary. It should also not be hard-coded, but should be pulled out. I think it probably makes sense to just store this alongside the other values as particleBounds.tof.binning

src/snapred/backend/recipe/algorithm/GroupByGroupCalibration.py
rboston628: thanks

bolotovskyr (approved):

PR #73, opened 2023-07-27
Give CustomGroupWorkspace an optional InputWorkspace

peterfpeterson:
Give CustomGroupWorkspace an option to take instrument from a supplied workspace. Since all current use of the Algorithm load the data in some way before calling, re-use the supplied workspace.

### DIFF
--- a/src/snapred/backend/recipe/algorithm/CustomGroupWorkspace.py
+++ b/src/snapred/backend/recipe/algorithm/CustomGroupWorkspace.py
@@ -45,7 +45,9 @@ class CustomGroupWorkspace(PythonAlgorithm):
 
         for grpIndx, focusGroup in enumerate(focusGroups):
             self.mantidSnapper.LoadDetectorsGroupingFile(
-                InputFile=focusGroup.definition, InputWorkspace=donorWorkspace, OutputWorkspace=focusGroup.name
+                InputFile=focusGroup.definition,
+                InputWorkspace=donorWorkspace,
+                OutputWorkspace=focusGroup.name,
             )
             self.mantidSnapper.executeQueue()
 
--- a/src/snapred/backend/recipe/algorithm/CustomGroupWorkspace.py
+++ b/src/snapred/backend/recipe/algorithm/CustomGroupWorkspace.py
@@ -1,8 +1,14 @@
 import json
 
-from mantid.api import AlgorithmFactory, PythonAlgorithm
+from mantid.api import AlgorithmFactory, MatrixWorkspaceProperty, PropertyMode, PythonAlgorithm
 from mantid.kernel import Direction
-from mantid.simpleapi import CreateWorkspace, GroupWorkspaces, LoadDetectorsGroupingFile, LoadInstrument
+from mantid.simpleapi import (
+    CreateWorkspace,
+    DeleteWorkspace,
+    GroupWorkspaces,
+    LoadDetectorsGroupingFile,
+    LoadInstrument,
+)
 
 from snapred.backend.dao.StateConfig import StateConfig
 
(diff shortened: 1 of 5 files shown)


src/snapred/backend/recipe/algorithm/CustomGroupWorkspace.py
```diff
@@ -24,19 +34,31 @@ def PyExec(self):
...
-        )
+        donorWorkspace = self.getProperty("InputWorkspace").value
+        loadEmptyInstrument = bool(not donorWorkspace)
+        if loadEmptyInstrument:
+            donorWorkspace = "idf"  # name doesn't matter
+            CreateWorkspace(OutputWorkspace=donorWorkspace, DataX=1, DataY=1)
+            LoadInstrument(
+                Workspace=donorWorkspace,
+                Filename="/SNS/SNAP/shared/Calibration/Powder/SNAPLite.xml",
+                MonitorList="-2--1",
```
bolotovskyr, line 45: I think we concluded earlier that `MonitorList` doesn't do anything as input and that it's unclear what "-2--1" means.

src/snapred/backend/recipe/algorithm/CustomGroupWorkspace.py
```diff
@@ -1,8 +1,14 @@
 import json
 
-from mantid.api import AlgorithmFactory, PythonAlgorithm
+from mantid.api import AlgorithmFactory, MatrixWorkspaceProperty, PropertyMode, PythonAlgorithm
 from mantid.kernel import Direction
-from mantid.simpleapi import CreateWorkspace, GroupWorkspaces, LoadDetectorsGroupingFile, LoadInstrument
+from mantid.simpleapi import (
```
bolotovskyr, line 5: Is there a reason why we have to import from `simpleapi` here and not use `MantidSnapper`?

peterfpeterson: I'm suspicious that this is older than `MantidSnapper`. I'll update the imports.

walshmm: Pete is correct

src/snapred/backend/recipe/algorithm/CustomGroupWorkspace.py
rboston628: I think it means pixels -2 to -1 are monitors.  The SNAPLite definition has two monitors in it.

rboston628: Do we want to always and only ever load the SNAPLite instrument?

bolotovskyr: > I think it means pixels -2 to -1 are monitors.

OK, maybe in the old days that property was used to set up monitor pixels. But now `MonitorList` is an _output_ property.  The C++ implementation touches it only at the end of `LoadInstrument::exec()`:
```
// Set the monitors output property
setProperty("MonitorList", (ws->getInstrument())->getMonitors());
```
In other words, I don't see how us passing that list as input would affect things.

rboston628: Removing it is probably a good idea, since this might presumably be used on non-SNAP instruments at some point

peterfpeterson: Related: we have a ton of places where the instrument is hard-coded as "SNAP" even though there is an abstraction for instrument in `Config`. That refactor should be a separate PR.

rboston628 (approved):

PR #79, opened 2023-08-01
makes importing things from modules cleaner

walshmm:
This is mostly a proof of concept, the code in the `dao` `__init__.py` will need to be propagated to the other inits


walshmm:
> I highly recommend against merging this. There is a bunch of unintended consequences associated with this. When you import, the top indentation level +1 of a file is executed. This will run a bunch of things you don't mean to (e.g. all the singletons get kicked off before they need to).

Just to add this to git from slack.
This is a shallow import, and this will mostly be used to import dataobjects more easily.  
Do you have an example of singletons being misconfig'd due to early imports? There shouldnt be any race conditions, otherwise theres a fundamental flaw with their implementation.  If its memory, I think we've kept them very light.  They are probably not the bottleneck for startup time.

src/snapred/backend/dao/calibration/Calibration.py
```diff
@@ -1,6 +1,7 @@
 from datetime import datetime
 
 from pydantic import BaseModel
+
 from snapred.backend.dao.state.InstrumentState import InstrumentState
```
peterfpeterson, line 5: should this be shorter? e.g. 
```py
from snapred.backend.dao.state import InstrumentState
```

src/snapred/backend/dao/calibration/CalibrationRecord.py
```diff
@@ -1,6 +1,7 @@
 from typing import Optional
 
 from pydantic import BaseModel
+
 from snapred.backend.dao.ReductionIngredients import ReductionIngredients
```
peterfpeterson, line 5: should this be shorter? e.g.
```py
from snapred.backend.dao import ReductionIngredients
```

PR #83, opened 2023-08-07
Enable mypy in pre-commit pre-push stage

peterfpeterson:
*DO NOT MERGE INTO NEXT UNTIL ISSUES ARE FIXED*

This turns mypy on for pre-push stage so pre-commit.ci will not timeout during its runs, but the developer can still see what is happening with the code.

To try this out
```sh
pre-commit run mypy --hook-stage=pre-push  --all
```

At the time of the last push, this results in
```sh
Found 35 errors in 15 files (checked 187 source files)
```

References
----------
* [Using mypy with an existing codebase](https://mypy.readthedocs.io/en/stable/existing_code.html)
* [pydantic with mypy](https://docs.pydantic.dev/latest/integrations/mypy/)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,8 +27,8 @@ repos:
     - id: black
       args: ['--line-length=120']
       exclude: "tests/cis_tests/.*"
-# disable for now
-#- repo: https://github.com/pre-commit/mirrors-mypy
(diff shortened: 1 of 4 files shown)


rboston628:
It's been a fun year, but I'm going to close this one out

PR #92, opened 2023-08-18
Diffraction calibration test panel ewm1841

walshmm:


src/snapred/ui/widget/WorkflowNode.py
darshdinger, line 1: My PR will need to be updated once this PR is merged. I use these workflow methods so it will need refactoring.

darshdinger (approved):
Looks good. I know that it will break the test panel calibration initialization check but other than that, I'm unsure if the workflow methods are used else where. They may be employed within the FitMultiplePeaks flow as well on the front end.

PR #96, opened 2023-08-24
Raw vanadium correction

rboston628:
**Description of work**

Implements a script to perform an unfocused normalization correction.  This is sometimes called a "raw vanadium correction," but it could use many other types of materials.

The work implements a [prototype script](https://github.com/neutrons/prototype_SNAPRed_scripts/blob/main/ewm780/raw_vanadium%20(1).py) from @mguthriem into SNAPRed.  

The main change is that this is not limited to the pre-defined vanadium cylinder, but can rather take any sample defined through the CalibrantSample DAO, and use mantid algorithms to set this as the sample.

**To test:**

Firstly, run the unit tests.

Within workbench, create reduction ingredients and a calibrant sample.

Setup the algorithm, and ensure the output has been properly corrected.

**Link to EWM item**

Work for [EWM780](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=780)


src/snapred/backend/recipe/algorithm/UnfocusedNormalizationCorrection.py
darshdinger, line 1: image
In testing this algo, I believe that this gives the correct output but I'm unsure of the issue found within the parameter sent to set the filename. It gives an error of file not being found but isn't this script supposed to produce an output correction file?

darshdinger (changes requested):

src/snapred/backend/recipe/algorithm/UnfocusedNormalizationCorrection.py
```diff
@@ -0,0 +1,209 @@
...
+        self.mantidSnapper = MantidSnapper(self, name)
+
+    def chopIngredients(self, ingredients: Ingredients) -> None:
+        stateConfig = ingredients.reductionState.stateConfig
+        self.liteMode: bool = stateConfig.isLiteMode
+        self.vanadiumRunNumber: str = ingredients.runConfig.runNumber
+        self.vanadiumBackgroundRunNumber: str = stateConfig.emptyInstrumentRunNumber
+        self.geomCalibFile: str = stateConfig.geometryCalibrationFileName
+        # iPrm['calibrationDirectory'] + sPrm['stateID'] +'/057514/'+ f'RVMB{VRun}
+        self.rawVFile: str = stateConfig.rawVanadiumCorrectionFileName
```
darshdinger, line 36: Might be an issue here.

darshdinger (changes requested):

src/snapred/backend/recipe/algorithm/UnfocusedNormalizationCorrection.py
darshdinger: Okay, found it. This file is defined within src.tests.resources.inputs.ReductionParameters.json -> rawVCorrFileName. Unsure as to why this is being invoked at all.

src/snapred/backend/recipe/algorithm/UnfocusedNormalizationCorrection.py
rboston628: What did you run to create this error?

darshdinger: Let me know if I'm missing something.

mguthriem: something looks to be hardwired to look for the prototype output unfocused vanadium correction: "RVMB57573.lite.nxs". I don't know why this would be looked for, since the purpose of the algo is to create this file from scratch?

darshdinger: ^^^^^^^^^^Yep @rboston628 this is what I ran into as well^^^^^^^^^^

rboston628: "RVMB57573.lite.nxs" is coming from the `rawVanadiumCorrectionFilename` [inside the `StateConfig` DAO](https://github.com/neutrons/SNAPRed/blob/4a64ab88c67747fbf8e07ac56a23027862a5f9ea/src/snapred/backend/dao/StateConfig.py#L19).  From what I understood, this is supposed to be a background run performed under analogous circumstances as the data, and the purpose of this algorithm is to subtract off the background data represented by this file.

mguthriem: `UnfocusedNormalizationCorrection` should be taking two input (raw neutron) calibration datasets (measured in a specific instrument state) and then using these to create a calibration dataset. It should also then save this calibration dataset to disk in the corresponding state calibration folder. 

In my prototype, I called these output files RVMB12345.nxs or RVM12345.lite.nxs (depending on lite mode), where 12345 was the run number of the input raw neutron dataset.

Subsequently, the RVMB file is loaded during reduction, is `DiffractionFocussed` (using exactly the same parameters as the sample data being reduced), `StripPeaksAndSmoothed` and, finally, used to *divide* the data.

This division normalizes for the fact that the total flux of neutrons generated (and measured) is different at different wavelengths.

rboston628:
This PR was opened before the hackathon, and its commit history got badly polluted since then.  I'm going to close and reopen a cleaner branch.

PR #99, opened 2023-08-30
New lite data algo, service and recipe integration: For EWM 759

darshdinger:
New lite data creation algo, service and recipe integration. For EWM 759


src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
```diff
@@ -0,0 +1,90 @@
...
+                superID = self.getSuperID(nativeID=eventID, xdim=superPixel[0], ydim=superPixel[0])
+
+                # clear detector IDs and then add the new modified super ID
+                event.clearDetectorIDs()
+                event.addDetectorID(superID)
+
+        self.mantidSnapper.DeleteWorkspace("Cleaning up input workspace...", Workspace=ws)
+        self.mantidSnapper.executeQueue()
+
+        self.setProperty("OutputWorkspace", outputWorkspaceName)
```
walshmm, line 62: you may need to update the associated instrument stored on the output workspace, otherwise it may still think its the full SNAP instrument as opposed to the lite version.

walshmm: otherwise I think everything else is shaping up pretty good.

walshmm: you should just be able to use
 self.mantidSnapper.[LoadInstrument](https://docs.mantidproject.org/nightly/algorithms/LoadInstrument-v1.html)("Loading instrument...",, Workspace=outputWorkspace, Filename="/SNS/SNAP/shared/Calibration/Powder/SNAPLite.xml", RewriteSpectraMap=False)

darshdinger: Should be added. Thank you!

src/snapred/backend/recipe/LiteDataRecipe.py
rboston628, line 1: You don't need this file, since it is already registered as a GenericRecipe

src/snapred/backend/service/LiteDataService.py
rboston628, line 1: I don't think we need a service for this task

src/snapred/backend/service/ServiceFactory.py
```diff
@@ -10,6 +10,7 @@
 from snapred.backend.service.ConfigLookupService import ConfigLookupService
 from snapred.backend.service.CrystallographicInfoService import CrystallographicInfoService
 from snapred.backend.service.FitMultiplePeakService import FitMultiplePeaksService
+from snapred.backend.service.LiteDataService import LiteDataService
```
rboston628, line 13: If the service is not needed, remove this line

src/snapred/backend/service/ServiceFactory.py
```diff
@@ -37,6 +38,7 @@ def __init__(self):
         self.serviceDirectory.registerService(FitMultiplePeaksService)
         self.serviceDirectory.registerService(VanadiumFocussedReductionService)
         self.serviceDirectory.registerService(SmoothDataExcludingPeaksService)
+        self.serviceDirectory.registerService(LiteDataService)
```
rboston628, line 41: If the service is not needed, remove this line

tests/unit/backend/recipe/test_LiteDataRecipe.py
rboston628, line 1: This is not needed, as it is registered as a generic algo

tests/unit/backend/service/test_LiteDataService.py
rboston628, line 1: I don't think this is needed, since I don't think the service is needed

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
```diff
@@ -0,0 +1,99 @@
...
+        self.log().notice("Lite Data Creation START!")
+        # define super pixel dimensions
+        superPixel = [8, 8]
+
+        # get output workspace name
+        outputWorkspaceName = self.getProperty("OutputWorkspace").value
+
+        # load input workspace
+        inputWorkspace = self.getProperty("InputWorkspace").value
+        self.mantidSnapper.mtd[inputWorkspace]
```
rboston628, line 29: I don't think this line does anything.  It is returning a pointer to the workspace, but this pointer is not retained in an l-value, so nothing occurs.

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
```diff
@@ -0,0 +1,99 @@
...
+                superID = self.getSuperID(nativeID=event, xdim=superPixel[0], ydim=superPixel[0])
+
+                # clear detector IDs and then add the new modified super ID
+                spectrum.clearDetectorIDs()
+                spectrum.addDetectorID(superID)
+
+        self.mantidSnapper.LoadInstrument(
+            "Loading instrument...",
+            Workspace=outputWorkspace,
+            Filename="/SNS/SNAP/shared/Calibration/Powder/SNAPLite.xml",
```
rboston628, line 65: I'm worried this ties this algorithm to only be able to run on analysis.  Is there a way to specify this filename, for instance to point at `tests/resources/inputs/calibration/fakeSNAPLite.xml`?

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
```diff
@@ -0,0 +1,99 @@
...
+        for spec in range(numSpec):
+            # get event list per workspace
+            spectrum = outputWorkspace.getSpectrum(workspaceIndex=spec)
+            spectrumData.append(spectrum)
+
+        # allocate array of event IDs
+        for spectrum in spectrumData:
+            eventIDs = spectrum.getDetectorIDs()
+            for event in eventIDs:
+                superID = self.getSuperID(nativeID=event, xdim=superPixel[0], ydim=superPixel[0])
```
rboston628, line 56: `ydim=superPixel[1]`

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
```diff
@@ -0,0 +1,99 @@
...
+    def PyInit(self):
+        self.declareProperty("InputWorkspace", defaultValue="", direction=Direction.Input)
+        self.declareProperty("OutputWorkspace", defaultValue="", direction=Direction.Output)
+        self.setRethrows(True)
+        self.mantidSnapper = MantidSnapper(self, name)
+
+    def PyExec(self):
+        self.log().notice("Lite Data Creation START!")
+        # define super pixel dimensions
+        superPixel = [8, 8]
```
rboston628, line 22: Do we really want this hard-coded?  Is there a way to either read this from the instrument definition, or to input this some other way?

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+        # self.mantidSnapper.DeleteWorkspace("Cleaning up input workspace...", Workspace=ws)
+        self.mantidSnapper.executeQueue()
+
+        self.setProperty("OutputWorkspace", outputWorkspaceName)
+        return outputWorkspaceName
+
+    def getSuperID(self, nativeID, xdim, ydim):
+        # Constants
+        Nx, Ny = 256, 256  # Native number of horizontal and vertical pixels
```
rboston628, line 76: My same concern about hard-coded values.

src/snapred/backend/recipe/LiteDataRecipe.py
darshdinger: I registered it to the GenericRecipe. I guess you're right, I think I may need to remove it from the generic recipe file.

src/snapred/backend/service/LiteDataService.py
darshdinger: The task explicitly called for the creation of a service. This might be a Michael question.

tests/unit/backend/recipe/test_LiteDataRecipe.py
darshdinger: Recipe creation for this task was within the story so again, maybe a Michael question.

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
darshdinger: The purpose of the algo is to run on the analysis machine. I understand what you are getting at but it needs to implement a new instrument definition. Are all the values contained within fakeSNAPLite.xml the same as the ones within SNAPLite.xml? In that case, this change would be fine.

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
darshdinger: Thank you for clearing this up for me.

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
darshdinger: After examination, these instrument definition files are not the same. The one within the inputs directory is explicitly for testing purposes only as seen in the comments. There are other details included as well within SNAPLite.xml that are missing within fakeSNAPLite.xml.

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
darshdinger: There are no signifiers for super pixel classification within the Instrument defintion.

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
darshdinger: Again, no instance of this within the instrument definition file.

<?xml version='1.0' encoding='UTF-8'?>

<instrument xmlns="http://www.mantidproject.org/IDF/1.0"
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="http://www.mantidproject.org/IDF/1.0 http://schema.mantidproject.org/IDF/1.0/IDFSchema.xsd"
            name="SNAP" valid-from   ="2018-05-01 00:00:01"
                        valid-to     ="2100-01-31 23:59:59"
		        last-modified="2019-05-03 11:30:00">
  
  
  
  <defaults>
    <length unit="metre"/>
    <angle unit="degree"/>
    <reference-frame>
      <along-beam axis="z"/>
      <pointing-up axis="y"/>
      <handedness val="right"/>
      <theta-sign axis="x"/>
    </reference-frame>
  </defaults>

  
  <component type="moderator">
    <location z="-15.0"/>
  </component>
  <type name="moderator" is="Source"/>

  
  <component type="sample-position">
    <location y="0.0" x="0.0" z="-5.0"/>
  </component>
  <type name="sample-position" is="SamplePos"/>

  
  <idlist idname="only-monitor">
    <id val="-1"/>
  </idlist>
  <component type="only-monitor" idlist="only-monitor">
    <location/>
  </component>
  <type is="monitor" name="only-monitor">
    <component type="monitor">
      <location x="0.0" y="0.0" z="15" name="monitor1"/>
    </component>
  </type>

  
  <component type="East">
    <location >
      <parameter name="roty">
        <logfile id="det_arc2" eq="180.0+value"/>
      </parameter>
      <parameter name="r-position">
        <logfile id="det_lin2" eq="0.5+value" />
      </parameter>
      <parameter name="t-position">
        <logfile id="det_arc2" />
      </parameter>
    </location>
  </component>

  <type name="East">
    <component type="Column1">
      <location/>
    </component>
  </type>

  <type name="Column1">
    <component type="panel" idstart="0" idfillbyfirst="y" idstepbyrow="1" >
      <location name="bank1">  
        <trans x="0.0" y="1.0" />
      </location>
    </component>
    <component type="panel" idstart="8" idfillbyfirst="y" idstepbyrow="1" >
      <location name="bank2">  
        <trans x="1.0" y="0.0" />
      </location>
    </component>
  </type>


<type name="panel" is="rectangular_detector" type="pixel"
    xpixels="1" xstart="0.0" xstep="+0.001"
    ypixels="8" ystart="0.0" ystep="+0.001" >
  <properties/>
</type>


<type is="detector" name="pixel">
  <cuboid id="pixel-shape">
    <left-front-bottom-point  y="0.0" x="0.0" z="0.0"/>
    <left-front-top-point     y="1.0" x="0.0" z="0.0"/>
    <left-back-bottom-point   y="0.0" x="0.0" z="-0.0001"/>
    <right-front-bottom-point y="0.0" x="1.0" z="0.0"/>
  </cuboid>
  <algebra val="pixel-shape"/>
</type>



<type is="monitor" name="monitor">
  <cylinder id="some-shape">
    <centre-of-bottom-base p="0.0" r="0.0" t="0.0"/>
    <axis y="0.0" x="0.0" z="1.0"/>
    <radius val="1.00"/>
    <height val="1.00"/>
  </cylinder>
  <algebra val="some-shape"/>
</type>

</instrument>

src/snapred/backend/service/LiteDataService.py
rboston628: @walshmm is a service necessary here?

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
rboston628: You quoted a fake instrument I made the unit testing. But, it does get at what I needed.  If you look at the original [SNAP definition](https://github.com/neutrons/SNAPRed/blob/next/tests/resources/inputs/pixel_grouping/SNAP_Definition.xml), it has the following:
``` xml
<!-- Rectangular Detector Panel -->
<type name="panel" is="rectangular_detector" type="pixel"
    xpixels="256" xstart="-0.078795" xstep="+0.000618"
    ypixels="256" ystart="-0.078795" ystep="+0.000618" >
  <properties/>
</type>
```
and then in the [SNAPLite definition](https://github.com/neutrons/SNAPRed/blob/next/tests/resources/inputs/pixel_grouping/SNAPLite_Definition.xml), this has been reduced to:
``` xml
<!-- Rectangular Detector Panel -->
<type name="panel" is="rectangular_detector" type="pixel"
    xpixels="32" xstart="-0.076632" xstep="+0.004944"
    ypixels="32" ystart="-0.076632" ystep="+0.004944" >
  <properties/>
</type>

```
where 256/32 = 8.  So both of these numbers could be derived from instrument definitions.  Lemme check with @mguthriem if there is any reason why that might ever be desirable.

darshdinger: Okay! Gotcha. Yeah, I'm unsure in a practicle sense that these values would change as the instrument's physical parameters won't change unless new upgraded detectors are used or different grouping methods are employed. Regardless, this is still hard coded but removed by an extra level of abstraction. You could point to these values within the instrument definition file, in specific, the local one (fakeSNAPLite.xml) but this would require a refactor of the instrument defintion file to update those values in case some fundamental change occurs to the instrument.

rboston628: @mguthriem suggests the use of lite mode is going to be specific to SNAP, so this is probably not worth worrying over at the moment.

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
rboston628: there is a SNAPLite.xml defined inside the test resources

src/snapred/backend/recipe/LiteDataRecipe.py
rboston628: I think it should stay as a generic recipe, since it follows the same process

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
darshdinger: I've changed the path to /tests/resources/inputs/pixel_grouping/SNAPLite_Definition.xml

rboston628 (approved):

PR #110, opened 2023-09-15
Create simple algorithm to load data for other algoritrhms

rboston628:
**Description of work**

Algorithms frequently need to use actual neutron scattering data.

It is undesirable for them to handle their own loading.  It would be better to pass this off somewhere.  Algorithms and recipes should not be calling services to do this.  At the recipe level, an algorithm is needed which will load scattering data, before the workspace is passed to an algorithm.

This is a first step.

**To test:**

The following algorithms call some form of "Load" function with nexus data:
- `AlignAndFocusReductionAlgorithm`
- `CalculateOffsetDIFC`
- `CalibrationReductionAlgorithm`
- `GroupByGroupCalibration`
- `ReductionAlgorithm`
- `VanadiumFocussedReductionAlgorithm`

In each, replace the Load function with a call to `RaidPantry`, passing if you wish the name of the Load function used in the `LoaderType` argument.  Then re-run unit tests and verify everything worked.

**Link to EWM item**
None, I did this for fun, following discussion in hackathon.

### DIFF
--- a/src/snapred/backend/recipe/algorithm/RaidPantry.py
+++ b/src/snapred/backend/recipe/algorithm/RaidPantry.py
@@ -1,0 +1,69 @@
+import json
+from typing import Dict, List, Tuple
+
+import numpy as np
+from mantid.api import (
+    AlgorithmFactory,
+    PythonAlgorithm,
+)
+from mantid.kernel import (
+    Direction,
+    FileAction,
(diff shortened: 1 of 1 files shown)


rboston628:
Apparently I did this twice... the is handled better in PR #158

PR #120, opened 2023-09-27
Updated CalibrantSample classes to match SetSample

dlcaballero16:
**Description of work**

Updated CalibrantSample classes to match SetSample
 - Geometry and Material classes had members updated to match what is being used in mantid's SetSample() function

**To test:**

On the analysis cluster, run mantid from SNAPRed and run a script that is able to call mantid's SetSample() using Geometry and Material objects created from a CalibrantSamples() object. SetSample() should not fail when called.

Link to EWM item
[Story 2436](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=2436)


src/snapred/backend/dao/state/CalibrantSample/Geometry.py
```diff
@@ -1,26 +1,26 @@
...
 from pydantic import BaseModel, root_validator
 
 
 class Geometry(BaseModel):
     """Class to hold Geometry data for Calibrant Samples
-    form: string must be 'cylinder' or 'sphere'
-    radius: single float (cm)
-    illuminated height: single float (cm)
-    total height: single float (cm) (not used for sphere)"""
+    Shape: string must be 'Cylinder' or 'Sphere'
```
walshmm, line 8: I would prefer that member variables start with a lowercase letter, unless this is somehow required to be in lockstep with mantid

dlcaballero16: So did some testing and it looks like those are not case sensitive to mantid, will update to be lowercase

walshmm (approved):

PR #121, opened 2023-09-28
pin check-wheel-contents so it doesnt install pydantic 2

walshmm:
**Description of work**
for some reason check-wheel-contents overwrites our pydantic 1 requirement.  Need to pin it.

**To test:**

check that the readthedocs build passes!


dlcaballero16 (approved):
Changes look good to me

PR #129, opened 2023-10-09
update to diffcal script for entire process

rboston628:
## Description of work

This updates the CIS script for testing the diffraction calibration.  The previous script only ran the offset pixel calibration. This will run the pixel calibration until it converges, and then run the group calibration.  It will also then call the recipe, which should do all of this.

## Explanation of work

This is only editing a CIS script.  It only adds some more lines to the script, or rearranges things to be more streamlined.

Make sure to toggle the `Config._config['cis_mode']` flag to True to see the dirty dishes.

### Dev testing

This is a CIS test, no edits made to SNAPRed, so no dev tests to run.

### CIS testing

This is a CIS test, run the new script and inspect the results.

## Link to EWM item

following on work for 
[EWM#1126](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=1126)


mguthriem (approved):
Approved

PR #133, opened 2023-10-11
fixes to calibrant sample DAOs for mantid API

rboston628:
## Description of work

This came out of PR #96, which needs to interact with mantid's `SetSamples` algorithm.  This adds properties to the DAOs to easily generate the needed dictionaries, and also runs some tests to ensure these work with mantid.

I changed uses of `List[float]` where the size of the list was known (3 or 6, say), and replaced with an appropriate `Tuple`.

The `Material` DAO would fail setting in mantid for single-element materials.  This is because mantid can use the fact that it is a single element to predict the number density (see [L333 in mantid's `MaterialBuilder`](https://github.com/mantidproject/mantid/blob/c1daec527052e9b88866bece07e6fadf35d6030c/Framework/Kernel/src/MaterialBuilder.cpp#L333)), leading to an over-determination of densities.  For this reason, single-element materials can only have _one_ of `packingFraction` and `massDensity`.

In `Atom`, it seemed redundant to refer to `atom_type` and `atom_coordinates`.  These were changed to `symbol` (to better reflect what it is), and `coordinates`.

I also put in camel case where snake case was used.

## Explanation of work

The main changes are the addition of this to `Geometry`:
``` python
    @property
    def geometryDictionary(self) -> Dict[str, Any]:
        ans = {
            "Shape": self.shape,
            "Radius": self.radius,
            "Center": list(self.center),
        }
        if self.shape == "Cylinder":
            ans["Height"] = self.height
            ans["Axis"] = list(self.axis)
        return ans
```
and this to `Material`:
``` python
    @property
    def materialDictionary(self) -> Dict[str, Any]:
        ans = {
            "ChemicalFormula": self.chemicalFormula,
        }
        if self.packingFraction is not None:
            ans["PackingFraction"] = self.packingFraction
        if self.massDensity is not None:
            ans["MassDensity"] = self.massDensity
        return ans
```

### Dev testing

The unit tests added for `Geometry`, `Material`, `Crystallography`, and `Atom` will test all of this new behavior.

### CIS testing

This does not need a CIS test, as it only involves internals.

If you want to run in workbench:
``` python
# import mantid algorithms, numpy and matplotlib
import numpy as np

import sys
sys.path.append("/SNS/users/4rx/SNAPRed")
from snapred.backend.dao.state.CalibrantSample.Geometry import Geometry
from snapred.backend.dao.state.CalibrantSample.Material import Material

# test that these can be used to set the sample in mantid
# run SetSample with the dictionary to set shape
# then get output XML of the sample shape
# note that the XML converts from cm to m
sampleWS = CreateWorkspace(
    DataX=1,
    DataY=1,
)

cylinder = Geometry(
    shape="Cylinder",
    radius=2,
    height=5,
)
sphere = Geometry(
    shape="Sphere",
    radius=3,
)

# test setting with a cylinder, compare output XML
# will have center-of-bottom-base, axis, height, and radius
SetSample(InputWorkspace=sampleWS, Geometry=cylinder.geometryDictionary)
ref = f"""
<type name="userShape">
    <{cylinder.shape.lower()} id="sample-shape">
        <centre-of-bottom-base x="0" y="{-cylinder.height*0.005}" z="0"/>
        <axis x="0" y="1" z="0"/>
        <height val="{cylinder.height*0.01}"/>
        <radius val="{cylinder.radius*0.01}"/>
    </{cylinder.shape.lower()}>
</type>
"""
ans = sampleWS.sample().getShape().getShapeXML()
ref = ref.replace("\n", "").replace(" ", "")
ans = ans.replace("\n", "").replace(" ", "")
assert ref == ans
# test setting with a sphere, compare output XML
# will have center and radius
SetSample(
    InputWorkspace=sampleWS,
    Geometry=sphere.geometryDictionary,
)
ref = f"""
<type name="userShape">
    <{sphere.shape.lower()} id="sphere">
        <center x="0" y="0" z="0"/>
        <radius val="{sphere.radius*0.01}"/>
    </{sphere.shape.lower()}>
</type>
"""
ans = sampleWS.sample().getShape().getShapeXML()
ref = ref.replace("\n", "").replace(" ", "")
ans = ans.replace("\n", "").replace(" ", "")
assert ref == ans

vanadium = Material(
            packingFraction=0.3,
            chemicalFormula="V",
        )
SetSample(
    InputWorkspace=sampleWS,
    Material=vanadium.materialDictionary,
)
material = sampleWS.sample().getMaterial()
assert material.chemicalFormula()[0][0].symbol == "V"
assert material.packingFraction == vanadium.packingFraction

vanadiumBoron = Material(
            packingFraction=0.3,
            massDensity=0.3,
            chemicalFormula="V B",
        )
SetSample(
    InputWorkspace=sampleWS,
    Material=vanadiumBoron.materialDictionary,
)
material = sampleWS.sample().getMaterial()
assert material.chemicalFormula()[0][0].symbol == "V"
assert material.chemicalFormula()[0][1].symbol == "B"
assert material.packingFraction == vanadium.packingFraction
```

This will only duplicate the unit tests for Material and Geometry.

In addition to this, run the `calibrant_samples_script.py` in the CIS tests.  This has been updated with tests of behavior.

## Link to EWM item

This work is related  to: 

[EWM#780](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=780)

### DIFF
--- a/src/snapred/backend/dao/state/CalibrantSample/Atom.py
+++ b/src/snapred/backend/dao/state/CalibrantSample/Atom.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Tuple
 
 from pydantic import BaseModel, validator
 
@@ -7,8 +7,8 @@ class Atom(BaseModel):
     """Class containing atomic paramaters used in Calibrant Samples
     atomic parameters:
 
-        atom type: string of chemical symbol
-        atom coordinates:
+        symbol: string of chemical symbol
+        coordinates:
 
             x coordinate: float in range [-1, 1]
             y coordinate: float in range [-1, 1]
@@ -17,18 +17,18 @@ class Atom(BaseModel):
         site occupation factor: float in range [0, 1]
         adp: positive float, default is 0.01"""
 
-    atom_type: str
-    atom_coordinates: List[float]
-    site_occupation_factor: float
+    symbol: str
+    coordinates: Tuple[float, float, float]
+    siteOccupationFactor: float
     adp: float = 0.1
 
-    @validator("atom_coordinates", allow_reuse=True)
+    @validator("coordinates", allow_reuse=True)
     def validate_atom_coordinates(cls, v):
-        if not all(-1 <= val <= 1 for val in v) and not len(v) == 3:
-            raise ValueError("atom coordinates must be 3 values (x, y, z) all in range [-1, 1]")
+        if not all(-1 <= val <= 1 for val in v):
+            raise ValueError("atom coordinates (x, y, z) must be all in range [-1, 1]")
         return v
 
-    @validator("site_occupation_factor", allow_reuse=True)
+    @validator("siteOccupationFactor", allow_reuse=True)
     def validate_site_occupation_factor(cls, v):
         if v < 0 or v > 1:
             raise ValueError("Site occupation factor must be a value in range [0, 1]")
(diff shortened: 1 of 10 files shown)


dlcaballero16:
You'll also want to update calibrant_samples_script.py in cis_tests. On L11. the Atom object needs to use the new member names

dlcaballero16 (approved):

dlcaballero16 (approved):

PR #135, opened 2023-10-13
Load outside of algos and cache

rboston628:
## Description of work

There should be segregation between loading and transforming data.  Many algorithms load their own data, which blurs the distinction between the data service and the algorithm layer.

This PR is a first step in removing data loading from algorithms and passing it to an external service.

For the moment, it does this by using a separate algorithm.  In the future, this algorithm should be replaced by a service.

### CIS testing

``` python
```

## Link to EWM item

[EWM#2481](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=2481)

### DIFF
--- a/src/snapred/backend/recipe/algorithm/RaidPantry.py
+++ b/src/snapred/backend/recipe/algorithm/RaidPantry.py
@@ -1,0 +1,81 @@
+import json
+from typing import Dict, List, Tuple
+
+import numpy as np
+from mantid.api import (
+    AlgorithmFactory,
+    PythonAlgorithm,
+)
+from mantid.kernel import (
+    Direction,
+    FileAction,
+    FileProperty,
+    StringListValidator,
+)
+
+from snapred.backend.recipe.algorithm.MantidSnapper import MantidSnapper
+
+name = "RaidPantry"
+
+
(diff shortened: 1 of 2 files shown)


src/snapred/backend/recipe/algorithm/RaidPantry.py
```diff
@@ -0,0 +1,81 @@
...
+    def PyExec(self) -> None:
+        runNumber = self.getProperty("RunNumber").value
+        if self.alreadyLoaded(runNumber):
+            return
+        filename = self.getProperty("Filename").value
+        outWS = self.getProperty("Workspace").value
+        loaderType = self.getProperty("LoaderType").value
+        if not self.mantidSnapper.mtd.doesExist(outWS):
+            if loaderType == "":
+                ws = loader = self.mantidSnapper.Load(
```
walshmm, line 66: why must a loader type be specified for load?

rboston628:
This is handled better now by Fetch Groceries in PR #158

rboston628:
See also PR #110

PR #136, opened 2023-10-13
raw vanadium

rboston628:
## Description of work

One of the steps of calibration is to compare to raw vanadium data.  This raw vanadium data must be processed to normalize by current, remove the background, and then account for absorption.

The work implements a [prototype script](https://github.com/neutrons/prototype_SNAPRed_scripts/blob/main/ewm780/raw_vanadium%20(1).py) from @mguthriem into SNAPRed.  

## Explanation of work

This implemented the above-mentioned prototype.  The technique involves: loading a vanadium file, loading a background measurement, processing both to be normalized, subtracting off the background from the vanadium, then calculating the absorption of neutron into the vanadium, dividing that out.

The original prototype used hard-coded values for a cylinder.  This uses a `CalibrantSamples` DAO inside the `SetSamples` method.  The hard-coded numbers of the prototype do not exactly correspond to a vanadium cylinder.

The prototype uses the `NumberOfSlices` and `NumberOfAnnuli` properties of the `CylinderAbsorption` method.  They define a mesh for the calculation, and are not part of the geometry.  Are those necessary? 

A lot of the code to review is fiddling with the `CalibrantSamples` and its subclasses.  The goal was to make its use in mantid as natural as possible, which included overwriting the `dict` method of `Geometry` and `Material`.  

There was a change made to `test_PixelGroupingParameters` to reflect a change in an error message thrown inside a negative test.

### Dev testing

This adds many tests to the DAOs.  The unit tests verify that everything works as expected and the algo will execute.

The `lazy_raw_vanadium` test in the cis_tests folder should show in the final result only a single peak at 70, of height less than 200.

### CIS testing

Run the `lazy_raw_vanadium` script in the cis_tests folder.  This should give a clear demonstration of what the algorithm does.

Run the `raw_vanadium_proto` script in the cis_tests folder.  This should duplicate the behavior of the original prototype.  [Run this prototype for comparison](https://github.com/neutrons/prototype_SNAPRed_scripts/blob/main/ewm780/reece_raw_vanadium.py).

## Link to EWM item

[EWM#780](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=780)

Also carried over into
[EWM#2711](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=2711)

Continuing work from PR #96

### DIFF
--- a/src/snapred/backend/dao/state/CalibrantSample/CalibrantSamples.py
+++ b/src/snapred/backend/dao/state/CalibrantSample/CalibrantSamples.py
@@ -16,7 +16,7 @@ class CalibrantSamples(BaseModel):
     date: Optional[str] = None
     geometry: Geometry
     material: Material
-    crystallography: Crystallography
+    crystallography: Optional[Crystallography]
 
     @validator("date", pre=True, always=True, allow_reuse=True)
     def set_datetime(cls, v):
--- a/src/snapred/backend/recipe/algorithm/RawVanadiumCorrection.py
+++ b/src/snapred/backend/recipe/algorithm/RawVanadiumCorrection.py
@@ -4,6 +4,9 @@ from typing import Any, Dict, Tuple
 import numpy as np
 from mantid.api import (
     AlgorithmFactory,
+    ITableWorkspaceProperty,
+    MatrixWorkspaceProperty,
+    PropertyMode,
     PythonAlgorithm,
 )
 from mantid.kernel import Direction
@@ -20,11 +23,24 @@ name = "RawVanadiumCorrection"
 class RawVanadiumCorrection(PythonAlgorithm):
     def PyInit(self):
         # declare properties
-        self.declareProperty("InputWorkspace", defaultValue="", direction=Direction.Input)
-        self.declareProperty("BackgroundWorkspace", defaultValue="", direction=Direction.Input)
+        self.declareProperty(
+            MatrixWorkspaceProperty("InputWorkspace", "", Direction.Input, PropertyMode.Mandatory),
+            doc="Workspace containing the raw vanadium data",
+        )
+        self.declareProperty(
+            MatrixWorkspaceProperty("BackgroundWorkspace", "", Direction.Input, PropertyMode.Mandatory),
+            doc="Workspace containing the raw vanadium background data",
+        )
+        self.declareProperty(
+            ITableWorkspaceProperty("CalibrationWorkspace", "", Direction.Input, PropertyMode.Mandatory),
+            doc="Table workspace with calibration data: cols detid, difc, difa, tzero",
+        )
+        self.declareProperty(
+            MatrixWorkspaceProperty("OutputWorkspace", "", Direction.Output, PropertyMode.Optional),
+            doc="Workspace containing corrected data; if none given, the InputWorkspace will be overwritten",
+        )
         self.declareProperty("Ingredients", defaultValue="", direction=Direction.Input)
         self.declareProperty("CalibrantSample", defaultValue="", direction=Direction.Input)
-        self.declareProperty("OutputWorkspace", defaultValue="vanadiumrawcorr_out", direction=Direction.Output)
         self.setRethrows(True)
         self.mantidSnapper = MantidSnapper(self, name)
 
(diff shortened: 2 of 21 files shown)


rboston628:
See [comments on previous PR](https://github.com/neutrons/SNAPRed/pull/96#discussion_r1331974045)

tests/unit/backend/recipe/algorithm/test_PixelGroupingParametersCalculationAlgorithm.py
```diff
@@ -341,7 +341,7 @@ def test_local_wrong_idf():
                 groupingFile=groupingFile,
                 referenceParametersFile=referenceParametersFile,
             )
-        assert "File 'junk' does not exist" in str(excinfo.value)
```
rboston628, line 344: @darshdinger this line here suddenly started causing problems for me locally this morning.  This test checks the error being returned  by checking the error string.  For some reason, this morning, with no changes made to the algo, this stared throwing a different error message when run locally.  This change made the test pass locally.  Not sure what happens elsewhere.

darshdinger (dismissed):
Looks good! As long as tests pass, should be good to merge. I will add other failing tests to EWM as a defect for another day.

src/snapred/backend/dao/state/CalibrantSample/CalibrantSamples.py
```diff
@@ -16,7 +16,7 @@ class CalibrantSamples(BaseModel):
     date: Optional[str] = None
     geometry: Geometry
     material: Material
-    crystallography: Crystallography
+    crystallography: Optional[Crystallography]
```
walshmm, line 19: why is this optional?  I'd like to avoid optional if possible.

src/snapred/backend/recipe/algorithm/RawVanadiumCorrection.py
```diff
@@ -0,0 +1,219 @@
...
+from snapred.backend.dao.ingredients import ReductionIngredients as Ingredients
+from snapred.backend.dao.state.CalibrantSample.CalibrantSamples import CalibrantSamples
+from snapred.backend.recipe.algorithm.MakeDirtyDish import MakeDirtyDish
+from snapred.backend.recipe.algorithm.MantidSnapper import MantidSnapper
+
+name = "RawVanadiumCorrection"
+
+
+# TODO: Rename so it matches filename
+class RawVanadiumCorrection(PythonAlgorithm):
```
walshmm, line 24: I'd prefer if algo names were appended with `Algo` at least.  We dont adhere to this everywhere, but I'd like to try.

walshmm:
When this does get merged, @rboston628 can you work with @darshdinger to create a method in the CalibrationService to consume this algorithm?

src/snapred/backend/dao/state/CalibrantSample/CalibrantSamples.py
rboston628: These are supposed to integrate easily with mantid.  Inside mantid, the sample is set using the appropriately named [`SetSample` algorithm](https://docs.mantidproject.org/nightly/algorithms/SetSample-v1.html).  However, this is only able to set the geometry and material properties of the sample (plus its environment and container).  It can't set the crystal structure of the container.  I think [`SetUB`](https://docs.mantidproject.org/nightly/algorithms/SetUB-v1.html) is the only algorithm that can let a user set the crystal structure, but so far we haven't called it.  I think it simplifies data handling having the crystallography here, but it's also only occasionally (and so far never) needed.

rboston628: It looks like the [`CrystalStructure`](https://docs.mantidproject.org/nightly/concepts/CrystalStructureAndReflections.html) data object, mixed with `ReflectionGenerator` object would also be able to use these, but so far we haven't worked with those, either, and their use seems pretty distinct from setting the calibration sample.

src/snapred/backend/recipe/algorithm/RawVanadiumCorrection.py
rboston628: Okie dokie

src/snapred/backend/dao/state/CalibrantSample/CalibrantSamples.py
mguthriem: We should not use `SetUB` for calibrant samples as that's specifically for single-crystal materials. I think CrystalStruture is the way to go for our powder samples.

darshdinger (approved):

PR #139, opened 2023-10-20
Updates to diffraction calibration

rboston628:
## Description of work

Diffraction calibration is one of the major epics of Phase 2 of SNAPRed.

It consists of two pieces:
1. pixel calibration (using offsets)
2. group calibration (aligning peaks)

This makes refactoring changes to both.

Firstly, the algorithm names were changed to be more descriptive.

Secondly, the algorithms were changed to no longer load their own data.  The neutron data and grouping information should be loaded by Fetch Groceries (PR #158) from within the `"calibration/diffraction"` service.

The workspace names can be passed to the Recipe using the `groceries` argument, like so:
``` python
rx = DiffractionCalibrationRecipe()
groceries = {
  "inputWorkspace": inWS,
  "groupingWorkspace": groupingWS,
  "outputWorkspace": outWS,
  "calibrationTable": calTable,
}
res = rx.executeRecipe(ingredients, groceries)
```
The "outputWorkspace" and "calibrationTable" keys are optional, and if used the output data from diffraction calibration will have these names.  Otherwise default names will be used, and returned with the same keys in `res`.

### CIS testing

Run `new_diffcal_script.py`.

``` python
import numpy as np
import json

import sys
sys.path.append("/SNS/users/4rx/SNAPRed")
from snapred.backend.dao.request.DiffractionCalibrationRequest import DiffractionCalibrationRequest
from snapred.backend.service.CalibrationService import CalibrationService
from snapred.backend.log.logger import snapredLogger
from snapred.meta.redantic import list_to_raw_pretty
snapredLogger._level = 20


from typing import List

from snapred.meta.redantic import list_to_raw_pretty

# User inputs ######################################################################
runNumber = "58882"  # 58409'
cifPath = "/SNS/SNAP/shared/Calibration/CalibrantSamples/Silicon_NIST_640d.cif"
groupingFile = "/SNS/SNAP/shared/Calibration/Powder/PixelGroupingDefinitions/SNAPFocGrp_Column.lite.xml"
peakThreshold = 0.05
offsetConvergenceLimit = 0.1

# SET TO TRUE TO STOP WASHING DISHES
Config._config['cis_mode'] = True
####################################################################################

diffcalRequest = DiffractionCalibrationRequest(
    runNumber = runNumber,
    cifPath = cifPath,
    focusGroupPath = groupingFile,
    convergenceThreshold = offsetConvergenceLimit,
    peakIntensityThreshold = peakThreshold,
)

calibrationService = CalibrationService()
res = calibrationService.diffractionCalibration(diffcalRequest)
print(res)
assert False
```

## Link to EWM item

[EWM#2794](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=2794)

### Additional PRs in this project

This work touched very many files, so many of the changes were broken off to keep individual PRs small.  Other PRs in this work:
- PR #150 
- PR #142 
- PR #149 
- PR #148 
- PR #152 
- PR #154 
- PR #157 
- PR #158 
- PR #163


src/snapred/backend/dao/ingredients/GroceryListItem.py
```diff
@@ -0,0 +1,25 @@
+from typing import Literal, Optional
+
+from pydantic import BaseModel, root_validator
+
+from snapred.backend.dao import RunConfig
+
+
+class GroceryListItem(BaseModel):
+    """Class to hold the ingredients necessary for pixel grouping parameter calculation."""
```
walshmm, line 9: comment probably needs updating

src/snapred/backend/dao/state/FocusGroup.py
```diff
@@ -12,3 +12,26 @@ class FocusGroup(BaseModel):
     dMax: List[float]
     dMin: List[float]
     definition: str
+
+    @root_validator(pre=True, allow_reuse=True)
```
walshmm, line 16: unfortunately these values will be going away in favor of pixelgroupingparameters, just a heads up.

rboston628: That will change a lot.  How would that square with the other PR #154

walshmm: I'm not sure these values are actually being used, it should be noted that they are about equivalent to pixelgroupingparameters.  The rational for their removal is the separation of static focusgroup data like name and definitionfile from the calculated values.

walshmm:
Somewhat related to the work done here and worth mentioning is this [workspacenamegenerator](https://github.com/neutrons/SNAPRed/compare/next...workspace_name_generator)(not quite ready yet) I've written up to help standardize our interaction and expectations with workspaces.

src/snapred/backend/service/CalibrationService.py
```diff
@@ -157,7 +159,25 @@ def diffractionCalibration(self, request: DiffractionCalibrationRequest):
             convergenceThreshold=convergenceThreshold,
         )
 
-        return DiffractionCalibrationRecipe().executeRecipe(ingredients)
+        # get the needed input data
+        groceryList = [
```
walshmm, line 163: I dont think the consumer should have to specify the loader implementation, it cares not the means of getting the data, it just wants data.  

I would put this responsibility in the data layer, where it would pass back a wrapper containing workspace name and a grocerylistItem, then upon interaction with the wrapper, MantidSnapper would FetchGroceries using the parameters provided in the wrapper and return the workspace.

rboston628: The user doesn't **have** to specify the loader.  The default behavior is to call the mantid "Load" algorithm, which will figure it out.  Once mantid runs "Load", the correct loader will be stored here; maybe one day in the future, when all wars are over, we can also write SNAPRed to remember which loader was used and use it again.  
But further, the **dev** should know which loader to use, and this lets the dev specify the one to use.

src/snapred/backend/dao/ingredients/GroceryListItem.py
rboston628: Good catch

src/snapred/backend/dao/state/FocusGroup.py
rboston628: I can guarantee these values are actually being used, inside the diffraction calibration process.  Changing them will require some additional changes to this workflow.

src/snapred/backend/service/CalibrationService.py
walshmm: User =\/= Consumer. 
Consumer is more generalistic and in this case refers to code that consumes this recipe.

I am suggesting that the implementation details of data loading functionality shouldnt live in the service layer but the data layer.   If the datatypes or loaders change we'd have to change this in every service that used it, it should be abstracted behind an api so services dont have to bear that burden.  This breaks form from the architecture.

rboston628: Can I request we make this a phase 3 (or 4, or 5) fix, rather than a requirement right now?  Because that's going to take me at least a week to work out

src/snapred/backend/recipe/DiffractionCalibrationRecipe.py
```diff
@@ -1,56 +1,96 @@
...
+        """
         self.runNumber = ingredients.runConfig.runNumber
         self.threshold = ingredients.convergenceThreshold
-        pass
 
-    def executeRecipe(self, ingredients: Ingredients) -> Dict[str, Any]:
+    def unbagGroceries(self, groceryList: Dict[str, Any]):
+        """
+        Checkout the workspace names needed for this recipe.
+        It is necessary to provide the followign keys:
```
dlcaballero16, line 29: following is misspelled

src/snapred/backend/recipe/DiffractionCalibrationRecipe.py
```diff
@@ -1,56 +1,96 @@
...
+        - "inputWorkspace": the raw neutron data
+        - "groupingWorkspace": a grouping workspace for focusing the data
+        It is optional to provide the following keys:
+        - "outputWorkspace": a name for the final output workspace; otherwise default is used
+        - "calibrationTable": a name for the fully caliibrated DIFC table; otherwise a default is used
```
dlcaballero16, line 34: calibrated is misspelled

rboston628: thanks

src/snapred/backend/recipe/DiffractionCalibrationRecipe.py
rboston628: got it

src/snapred/backend/service/CalibrationService.py
walshmm: I'd rather not put it off too long, but it doenst have to be part of the pr.

walshmm:
rebasing after [This](https://github.com/neutrons/SNAPRed/pull/168) pr should get it to pass

walshmm (approved):

PR #148, opened 2023-10-25
Fixes to crystallographic ingestion

rboston628:
## Description of work

A few things came up.

1. The algo to load crystal data had the dumb word "Ingest" hanging around, which was entirely an artifact of Malcolm's word choice in a story description.  That was removed.
2. The tests of the  CrystallographicIngestionAlgorithm used a very low d-spacing (0.1) which made the test very, very slow.
3. It was noted that CrystallographicIngestionAlgorithm called `LoadCIF`, which was sometimes slow, to generate a mantid `CrystalStructure` object.  However, SNAPRed had recently implemented the `Crystallography` DAO which is supposed to capture much of the same behavior as `CrystalStructure`.  Therefore, this algorithm should be able to use *either* a CIF path *or* a `Crystallography` to create the `CrystalStructure` that is used in this algorithm to generate the crystal data.
4. While it is easy to initialize a `CrystalStructure` in mantid using the SNAPRed `Crystallography`, going the other was was not easy.  This enables the ability to set `Crystallography` using a `CrystalStructure` (and `Atom` using a scatterer string).

## Explanation of work

In the DAOs, the `__init__` method was overridden.  If position arguments are given (`*args`), then it will call the BaseModel constructor using kwargs set from the args properties.

### CIS testing

``` python
```

## Link to EWM item


dlcaballero16 (dismissed):

src/snapred/backend/dao/state/CalibrantSample/Atom.py
```diff
@@ -22,6 +22,20 @@ class Atom(BaseModel):
     siteOccupationFactor: float
     adp: float = 0.1
 
+    @validate_arguments
+    def __init__(self, *args: StrictStr, **kwargs):
+        if args:
+            print(args)
```
dlcaballero16: Assuming this print is for testing, probably should remove this

dlcaballero16 (approved):

src/snapred/backend/dao/state/CalibrantSample/Atom.py
rboston628: Good catch

PR #153, opened 2023-10-31
Added fixes and tests for Save/LoadGroupingDefinition

dlcaballero16:
Fixed LoadGroupingDefinition.py by updating the input checks Added cis test for SaveGroupingDefinition

## Description of work
This is to fix the defect related to SaveGroupingDefintion and LoadGroupingDefinition

## Explanation of work

In LoadGroupingDefinition.py, input validation now correctly checks for InstrumentName and InstrumentFilename being set. A warning instead of an exception is now thrown if InstrumentDonor is set when it is not required. A lite mode check is now done to set the instrument file name.

### Dev testing

Make sure unit tests pass.

### CIS testing

Run the new script included called save_grouping_definition_with_xml.py on analysis.

GUI Testing

In mantid workbench, open the SNAPRed GUI. Then open the test panel. In the test panel, give it a run number like 58882, and select a sample file. For the grouping file, select SNAPFocGroup_Column.lite.hdf and then hit continue. After some processing, there should be no errors. If the file does not show up in the selection box, application.yml needs to be changed.

## Link to EWM item

[EWM Defect #2621](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=2621)

[EWM Task #2779](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=2779)

### DIFF
--- a/src/snapred/backend/recipe/algorithm/LoadGroupingDefinition.py
+++ b/src/snapred/backend/recipe/algorithm/LoadGroupingDefinition.py
@@ -7,8 +7,6 @@ from snapred.backend.log.logger import snapredLogger
 from snapred.backend.recipe.algorithm.MantidSnapper import MantidSnapper
 from snapred.meta.Config import Config
 
-name = "LoadGroupingDefinition"
-
 logger = snapredLogger.getLogger(__name__)
 
 
@@ -56,7 +54,7 @@ class LoadGroupingDefinition(PythonAlgorithm):
         )
 
         self.setRethrows(True)
-        self.mantidSnapper = MantidSnapper(self, name)
+        self.mantidSnapper = MantidSnapper(self, __name__)
 
         # define supported file name extensions
         self.supported_calib_file_extensions = ["H5", "HD5", "HDF"]
@@ -91,9 +89,7 @@ class LoadGroupingDefinition(PythonAlgorithm):
         grouping_file_name = self.getProperty("GroupingFilename").value
         output_ws_name = self.getProperty("OutputWorkspace").value
         file_extension = pathlib.Path(grouping_file_name).suffix.upper()[1:]
-        isLite = False
-        if ".lite" in grouping_file_name:
-            isLite = True
+        isLite: bool = ".lite" in grouping_file_name
         if isLite and self.getProperty("InstrumentFilename").value == "":
             self.setProperty("InstrumentFilename", Config["instrument.lite.definition.file"])
         if file_extension in self.supported_calib_file_extensions:
--- a/src/snapred/backend/recipe/algorithm/LoadGroupingDefinition.py
+++ b/src/snapred/backend/recipe/algorithm/LoadGroupingDefinition.py
@@ -1,4 +1,5 @@
 import pathlib
+from datetime import datetime
 
 from mantid.api import AlgorithmFactory, MatrixWorkspaceProperty, PropertyMode, PythonAlgorithm
 from mantid.kernel import Direction
@@ -114,11 +115,12 @@ class LoadGroupingDefinition(PythonAlgorithm):
             preserve_donor = True if instrument_donor else False
             if not instrument_donor:
                 # create one from the instrument definition file
-                instrument_donor = self.mantidSnapper.LoadEmptyInstrument(
+                instrument_donor = datetime.now().ctime() + "_idf"
+                self.mantidSnapper.LoadEmptyInstrument(
                     "Loading instrument definition file...",
                     Filename=self.getProperty("InstrumentFilename").value,
                     InstrumentName=self.getProperty("InstrumentName").value,
-                    OutputWorkspace="idf",
+                    OutputWorkspace=instrument_donor,
                 )
             self.mantidSnapper.LoadDetectorsGroupingFile(
                 "Loading grouping definition from detectors grouping file...",
--- a/src/snapred/backend/recipe/algorithm/LoadGroupingDefinition.py
+++ b/src/snapred/backend/recipe/algorithm/LoadGroupingDefinition.py
@@ -3,9 +3,13 @@ import pathlib
 from mantid.api import AlgorithmFactory, MatrixWorkspaceProperty, PropertyMode, PythonAlgorithm
 from mantid.kernel import Direction
 
+from snapred.backend.log.logger import snapredLogger
 from snapred.backend.recipe.algorithm.MantidSnapper import MantidSnapper
+from snapred.meta.Config import Config
 
 name = "LoadGroupingDefinition"
+
+logger = snapredLogger.getLogger(__name__)
 
 
 class LoadGroupingDefinition(PythonAlgorithm):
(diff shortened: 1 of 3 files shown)


src/snapred/backend/recipe/algorithm/LoadGroupingDefinition.py
```diff
@@ -3,10 +3,14 @@
 from mantid.api import AlgorithmFactory, MatrixWorkspaceProperty, PropertyMode, PythonAlgorithm
 from mantid.kernel import Direction
 
+from snapred.backend.log.logger import snapredLogger
 from snapred.backend.recipe.algorithm.MantidSnapper import MantidSnapper
+from snapred.meta.Config import Config
 
 name = "LoadGroupingDefinition"
```
rboston628, line 10: We should move away from the use of this variable.  Can you delete this line (L10), and replace it anywhere it is used with `__name__`, as you did in L12?

src/snapred/backend/recipe/algorithm/LoadGroupingDefinition.py
```diff
@@ -73,19 +77,25 @@ def validateInput(self) -> None:
...
+                logger.warn("InstrumentDonor will only be used if GroupingFilename is in XML format.")
 
     def PyExec(self) -> None:
         self.validateInput()
         grouping_file_name = self.getProperty("GroupingFilename").value
         output_ws_name = self.getProperty("OutputWorkspace").value
         file_extension = pathlib.Path(grouping_file_name).suffix.upper()[1:]
+        isLite = False
+        if ".lite" in grouping_file_name:
+            isLite = True
```
rboston628, line 96:
``` python
isLite: bool = ".lite" in grouping_file_name
```

src/snapred/backend/recipe/algorithm/LoadGroupingDefinition.py
```diff
@@ -73,19 +77,25 @@ def validateInput(self) -> None:
         if file_extension in self.supported_calib_file_extensions:
             instrument_name = self.getProperty("InstrumentName").value
             instrument_file_name = self.getProperty("InstrumentFilename").value
-            absent_input_properties_count = sum(not prop for prop in [instrument_name, instrument_file_name])
-            if absent_input_properties_count == 0 or absent_input_properties_count == 2:
-                raise Exception("Either InstrumentName or InstrumentFilename must be specified, but not both.")
+            if instrument_name + instrument_file_name != "":
+                if not (instrument_name == "" or instrument_file_name == ""):
+                    raise Exception("Either InstrumentName or InstrumentFilename must be specified, but not both.")
+
```
rboston628, line 83: I think the original method of counting was cleaner than these string comparisons.  Why is an instrument donor not being considered a possibility for the case of H5/HD5/HDF files?

dlcaballero16: The instrument donor is only used in the case of an XML file

rboston628:
I ran the following script (edited from the CIS test in this PR) and found it to run and match the workspace
``` python
# import mantid algorithms, numpy and matplotlib
import numpy as np
import pathlib
import os.path

from snapred.backend.recipe.algorithm.LoadGroupingDefinition import LoadGroupingDefinition as LoadingAlgo
from snapred.backend.recipe.algorithm.SaveGroupingDefinition import SaveGroupingDefinition as SavingAlgo
from snapred.meta.Config import Config, Resource

localDir = '~/tmp/'
PGDDir = '/SNS/SNAP/shared/Calibration/Powder/PixelGroupingDefinitions/'
groupingFile = PGDDir + "SNAPFocGroup_Column.xml"

Config._config["cis_mode"] = True # <<-- change to True and rerun

output_file_name = pathlib.Path(groupingFile).stem + ".hdf"
outputFilePath = os.path.join(localDir, output_file_name)
# save the input grouping file in the calibration format
savingAlgo = SavingAlgo()
savingAlgo.initialize()
savingAlgo.setProperty("GroupingFilename", groupingFile)
savingAlgo.setProperty("OutputFilename", outputFilePath)
savingAlgo.setProperty("InstrumentName", "SNAP")
assert savingAlgo.execute()

loadingAlgo = LoadingAlgo()
loadingAlgo.initialize()
loadingAlgo.setProperty("GroupingFilename", outputFilePath)
loadingAlgo.setProperty("InstrumentName", "SNAP")
loadingAlgo.setProperty("OutputWorkspace", "_impossible_name")
assert loadingAlgo.execute()

loadingAlgo2 = LoadingAlgo()
loadingAlgo2.initialize()
loadingAlgo2.setProperty("GroupingFilename", groupingFile)
loadingAlgo2.setProperty("InstrumentName", "SNAP")
loadingAlgo2.setProperty("OutputWorkspace", "_another_weird_name")
assert loadingAlgo2.execute()

assert CompareWorkspaces(
    Workspace1 = "_impossible_name",
    Workspace2 = "_another_weird_name",
)
```

src/snapred/backend/recipe/algorithm/LoadGroupingDefinition.py
rboston628: As far as I understand, all three methods can be used regardless of H5 or XML.  In the case of H5, the [LoadDiffCal](https://docs.mantidproject.org/nightly/algorithms/LoadDiffCal-v1.html) algorithm can be given an instrument donor through the `InputWorkspace` property, like so:
``` python
self.mantidSnapper.LoadDiffCal(
    "Loading grouping definition from calibration file...",
    Filename=grouping_file_name,
    InstrumentName=self.getProperty("InstrumentName").value,
    InstrumentFilename=self.getProperty("InstrumentFilename").value,
    InputWorkspace=self.getPropertyValue("InstrumentDonor"),
    MakeGroupingWorkspace=True,
    MakeCalWorkspace=False,
    MakeMaskWorkspace=False,
    WorkspaceName=output_ws_name,
)
```

dlcaballero16: Ah ok. The call to LoadDiffcal below on line 96 was not making use of the InputWorkpsace property

rboston628:
I also discovered that if there is a workspace named "idf", then the loading algo will delete it, which is probably not desired.  This can be fixed by changing the temp filename at L121 from "idf" to something less likely to cause collision

src/snapred/backend/recipe/algorithm/SaveGroupingDefinition.py
```diff
@@ -123,6 +128,10 @@ def PyExec(self) -> None:
             f"Cleanup the zero calibration workspace {cal_ws_name}",
             Workspace=cal_ws_name,
         )
+        self.mantidSnapper.WashDishes(
+            f"Cleanup the grouping workspace {grouping_ws_name}",
+            Workspace=grouping_ws_name,
+        )
```
rboston628, line 134: This _could_ point to the user's grouping workspace, in which case we'd wipe it.  This should check how the grouping workspace was set first,

rboston628 (dismissed):
Everything works fine, and using the h5py method the saving runs in about 10s.  This is provisional a new update that will use an improved mantid method.

rboston628:
H5py repeatedly fails at basically its only job, which is creating hdf files, throwing inexplicable and confusing errors that cause unit tests to fail.  Has been removed, will approve now in anticipation of better mantid solution.

rboston628 (approved):

PR #155, opened 2023-11-02
[Defect] Fixed LiteDataCreationAlgo.

darshdinger:
## Description of work
Upon testing of the new LiteDataCreationAlgo, the algo executed successfully. Upon inspection of the output workspace, it was found the the updating of the instrument definition worked properly but the reduction of the number of histograms (reduction by a factor of 64) did not occur. The input and output workspaces for a given test run had the same number of histograms.

Solution suggested by Andrei:

suggested fix:
a) create a grouping file reflecting the 8x8 grouping of SNAPLite
b) use GroupDetectors, using this grouping information to create the the necessary output file.

Malcolm and Andrei helped to find the working solution to this issue.

**UPDATE:**
Malcolm requested to use `LoadGroupingDefinition` from within SNAPRed. This has now replaced the on the fly calculation.

## Explanation of work
There were a few different implementations of the grouping schema but the one from the prototype script functioned correctly. In this new implementation, `GroupDetectors` is used to achieve the remapping.

### Dev testing
For Dev testing, please run `pytest` to insure all tests pass.

### CIS testing
1. `git pull fix-Lite-Data-Creation-Algo`
2. `python -m workbench`
3. Open script `lite_data_creation.py` within `SNAPRed/tests/cis_tests/` directory
4. Run within the workbench

``` python
# Use this script to test LiteDataCreationAlgo.py
import numpy as np

from snapred.backend.recipe.algorithm.LiteDataCreationAlgo import LiteDataCreationAlgo
from snapred.backend.dao.RunConfig import RunConfig
from snapred.backend.data.DataFactoryService import DataFactoryService

#User input ###########################
runNumber = "47278"
#######################################

dataFactoryService = DataFactoryService()

runConfig = dataFactoryService.getRunConfig(runNumber)
ipts = runConfig.IPTS
run = RunConfig(runNumber=runNumber,IPTS=ipts)

LDCA = LiteDataCreationAlgo()
LDCA.initialize()
LDCA.setProperty("InputWorkspace", f"{runNumber}_raw")
LDCA.setProperty("Run", run.json())
LDCA.setProperty("OutputWorkspace", f"{runNumber}_lite")
LDCA.execute()
```

## Link to EWM item

[EWM# 2782](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=2782)


mguthriem:
Have just tested this script. It works as expected and seems performant. Only a couple of comments: 1) it would be useful to add a setting that allows automatic deletion of the input (non-lite) workspace, leaving only the output (lite) workspace in memory. 2) after execution, does does the run config need to be modified in anyway, so indicate that a lite file has been created?

darshdinger:
> Have just tested this script. It works as expected and seems performant. Only a couple of comments: 1) it would be useful to add a setting that allows automatic deletion of the input (non-lite) workspace, leaving only the output (lite) workspace in memory. 2) after execution, does does the run config need to be modified in anyway, so indicate that a lite file has been created?

1. Yeah! I can update the script to accept a flag that deletes the non-lite workspace once the execution is finished. So that should be a simple feature to add. 

2. I think some of the ingredients (which is kinda like RunConfig) [as they are pydantic objects] have a Boolean value called “isLiteMode” and that can be set but RunConfig doesn’t have that. We can add something like that if we need to if it would be helpful in differentiating between lite and non-lite files.

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
```diff
@@ -4,94 +4,89 @@
...
-                # clear detector IDs and then add the new modified super ID
-                spectrum.clearDetectorIDs()
-                spectrum.addDetectorID(superID)
-
-        self.mantidSnapper.LoadInstrument(
-            "Loading instrument...",
-            Workspace=outputWorkspace,
-            Filename=Resource.getPath("/inputs/pixel_grouping/SNAPLite_Definition.xml"),
-            RewriteSpectraMap=False,
+        self.mantidSnapper.mtd[groupingWorkspaceName]
```
rboston628, line 63: The linter "fixed" this line by removing its assignment.  You can just delete the line.

src/snapred/backend/service/LiteDataService.py
rboston628, line 1: Is there an associated recipe?  Given the simplicity of the algorithm after all loading is removed, I actually think it would be better to remove the algo, and just have a recipe that makes the two mantid calls that are left (GroupDetectors and CompressEvents)

tests/unit/backend/recipe/algorithm/test_LiteDataCreationAlgo.py
rboston628, line 1: It looks like this used to test if the lite conversion was successful, and now it doesn't.  Could you do some sort of test that the number of spectra has been reduced by a factor of 64, or that data in the superpixels matches data in the ubspixels?

rboston628 (changes requested):

src/snapred/backend/recipe/algorithm/LiteDataCreationAlgo.py
```diff
@@ -4,94 +4,89 @@
...
-            InputWorkspace=inputWorkspace,
-            OutputWorkspace=outputWorkspaceName,
+        groupingWorkspaceName = f"{self.runNumber}_lite_grouping_ws"
+
+        # load grouping map
+        self.mantidSnapper.LoadGroupingDefinition(
+            "Loading lite grouping information...",
+            GroupingFilename=str(Config["instrument.lite.map.file"]),
+            InstrumentFilename=str(Config["instrument.native.definition.file"]),
+            OutputWorkspace=groupingWorkspaceName,
```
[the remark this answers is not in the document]
mguthriem: I'm not sure about this comment. The grouping file (the map from full to lite) is a special one for doing the lite conversion. If this were loaded outside of the algo, then we would need an extended recipe that knew to look for it when the request is made to generate a lite version. 

It's true that, at the moment, every run that SNAPRed loads will need to be converted to lite. So it would make sense to persist the lite grouping workspace rather than load it everytime. I just don't know how SNAPRed would know when to load and when to delete if it's done outside of the LiteDataCreation algo??

src/snapred/backend/service/LiteDataService.py
darshdinger: It could be done this way but it would be better to leave it as an algo, this will need front end integration as part of phase 2 and containing this logical within an algo makes more sense for orchestration.

dlcaballero16 (dismissed):

dlcaballero16 (approved):

PR #159, opened 2023-11-10
issue template from EWM story template

rboston628:
## Description of work

Sometimes it is easier to write story descriptions in markdown.  Github's markdown also allows for embedding equation with $\LaTeX$ such as

$DIFC_{prev} = (1-|dX|)^{-offset}$

for directly typing syntax-highlighted code, such as

``` python
x = [1,2,3]
y = [4,5,6]
xy = {key:value for key,value in zip(x,y)}
print(xy)
```
and for directly embedding graphs like this one:
Screenshot from 2023-10-23 12-19-46

which are all useful in describing the stories.  The issues can also be tracked directly to PRs that should close them.

All of this points to the u

sefulness of using github issues as a means of communicating needed work.

## Explanation of work

The issue template directly implements the story template for EWM stories developed by Michael and Malcolm.

### Dev testing
N/A

### CIS testing
N/A

## Link to EWM item
N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/workitem.md
+++ b/.github/ISSUE_TEMPLATE/workitem.md
@@ -1,0 +1,39 @@
+---
+name: Work Item
+about: Description of EWM work item
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+## Overview
+### PR's
+Where the pr's related to the story go.
+### I/O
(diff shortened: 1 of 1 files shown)


rboston628:
This idea was torpedoed, as it requires not only linking to the issue but duplicating the same writeup in EWM.  Maybe copy/paste from the rtf rendered markdown could save it?

PR #168, opened 2023-11-21
fix setuptools for deploy/pip install

walshmm:
## Description of work

Fixes setuptools so gui tests and deploys can pip install.

## Explanation of work

I added some missing props and removed the duplicate license members

### Dev testing

ci passes

## Link to EWM item

[EWM#3115](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3115)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,6 +8,8 @@ license = { text = "GPL3.0" }
 
 [tools.setuptools.dynamic.dependecies]
 
+[project.scripts]
+snapred = "snapred.__main__:main"
 
 [build-system]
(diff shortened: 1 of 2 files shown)


peterfpeterson (approved):
This looks good. It feels like python-build is starting to change its policy on whether `pyproject.toml` or `setup.cfg` wins when information is specified in both. We should probably make the effort to move entirely to `pyproject.toml`.

PR #169, opened 2023-11-21
Update RawVanadiumCorrectionAlgorithm

bolotovskyr:
## Description of work

Refactors `RawVanadiumCorrectionAlgorithm` according to the EWM story linked below.
## Explanation of work

- `ApplyDiffCal` step is removed from the algorithm. Correspondingly, `CalibrationWorkspace` is removed as input property. 
- Unit test `test_RawVanadiumCorrection` and CIS tests `lazy_raw_vanadium` and `raw_vanadium_proto` are adjusted so they do not set `CalibrationWorkspace` property on the algorithm
- `NormaliseByCurrent` is moved to the end of processing steps in `chopNeutronData`

### Dev testing
pytest -s tests/unit/backend/recipe/algorithm/test_RawVanadiumCorrection.py

### CIS testing
pytest -s tests/cis_tests/lazy_raw_vanadium.py
pytest -s tests/cis_tests/raw_vanadium_proto.py

## Link to EWM item

[EWM#2989](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=2989)


src/snapred/backend/recipe/algorithm/RawVanadiumCorrectionAlgorithm.py
```diff
@@ -29,10 +28,6 @@ def PyInit(self):
             MatrixWorkspaceProperty("BackgroundWorkspace", "", Direction.Input, PropertyMode.Mandatory),
```
rboston628, line 28: I'm wondering if "InputWorkspace" should be changed to "VanadiumWorkspace"?

bolotovskyr: I am not sure. Yes, it would make the property name more meaningful. But on the other hand,  `InputWorkspace`/`OutputWorkspace` is a name pattern used in many `Mantid` algorithms. Also, to be consistent, if we went with `VanadiumWorkspace`, shouldn't we then change `OutputWorkspace` to `CorrectedVanadiumWorkspace`?

rboston628:
I tested both of the raw vanadium scripts in workbench, and they seemed to work correctly, after I made some minor fixes leftover from my own messing around with trying to set the test environment.  

While I was adding commits, I added a category to every algorithm to prevent workbench warnings from being generated, and I removed every instance of an algorithm having a `name` variable equal to its name used to initialize it (replaced with `__name__`).  These are just minor changes to keep the codebase current.

This is what the raw vanadium correction from the workbench test looks like:
image

rboston628 (dismissed):

rboston628 (approved):

rboston628 (approved):

PR #172, opened 2023-12-04
Advance calibration quality accessment

bolotovskyr:
## Description of defect

Calibration assessment is facilitated by examining calibration metrics. Currently there are 2 problems with it:

1. The metrics and their errors are output as columns into a table workspace. When comparing metrics from different calibrations, the data in separate table workspaces can't be plotted in a single plot, so quantitative comparative assessment of quality is difficult. 
2. The user needs to be able to load any earlier calibration, not just the most recent one.

## Explanation of work

Convert the table workspace into matrix workspaces as follows: 

`X`:  `twoThetaAverage`  | `Y`: `sigmaAverage` | `E`: `sigmaStandardDeviation`
`X`:  `twoThetaAverage`  | `Y`: `strainAverage` | `E`: `strainStandardDeviation`

Support the following workflow:

1. Per user's GUI request, the backend returns an index of all existent calibration records corresponding to the instrument state which is associated with the current run ID.
2. The user selects a calibration record  (version) to be examined
3. The backend loads the specified calibration record and (a) generates the matrix workspaces (see above) and (b) loads persistent workspaces referenced by the calibration record

### Dev testing

```
pytest -s tests/unit/backend/data/test_DataFactoryService.py
pytest -s tests/unit/backend/service/test_CalibrationService.py
```

### CIS testing
?

<!--``` python
```
-->

<!-- GUI
If testing should instead be done from the GUI, explain
 - how to access the feature in the GUI
 - what buttons to click
 - what output should be generated in both test and fail.  state the SPECIFIC text
 - what will the CIS see?  link to screenshots of both success and failure, if possible
-->

## Link to EWM item
<!-- LINK TO THE EWM HERE -->
[EWM#2770](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=2770)

<!--
Inside the EWM, paste a link to this PR in a comment there
Link to any other relevant context, such as related mantid PRs, related SNAPRed PRs, related issues, etc.
-->

### CODE REVIEW

File: src/snapred/backend/data/LocalDataService.py [hunk 1]
```diff
@@ -448,6 +448,16 @@ def readCalibrationRecord(self, runId: str, version: str = None):
             record = parse_file_as(CalibrationRecord, latestFile)
         return record
 
+    def readCalibrationRecords(self, runId: str) -> CalibrationRecordList:
```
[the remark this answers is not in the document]
bolotovskyr: My understanding from talking to Malcolm is that he expects to select a subset from an available list of strings, e.g.
Run 1234 12/06/2023
Run 56789 12/07/2023
etc.

So that means the backend should be able to:
(1) produce such a list given a single run number (without loading/creating any workspaces)
(2) load/generate all relevant workspaces on user's request, where the request might contain more than one run id

src/snapred/backend/service/CalibrationService.py
```diff
@@ -344,10 +346,24 @@ def _collectMetrics(self, focussedData, focusGroup, pixelGroupingParam):
...
+            raise ValueError(f"No calibration record found for run {runId}, version {version}.")
+        GenerateCalibrationMetricsWorkspaceRecipe().executeRecipe(
+            CalibrationMetricsWorkspaceIngredients(calibrationRecord=calibrationRecord)
+        )
+        for ws_name in calibrationRecord.workspaceNames:
+            path = self.dataFactoryService.getCalibrationDataPath(
+                calibrationRecord.runNumber, calibrationRecord.version
+            )
+            fullPath = os.path.join(path, ws_name + ".nxs")
+            self.dataFactoryService.loadWorkspace(fullPath, ws_name)
```
walshmm, line 362: The part where it finds the path should probably live in the data layer.

I would condense these into a `loadCalibrationDataWorkspace` method.

tests/unit/backend/service/test_CalibrationService.py
```diff
@@ -287,8 +316,68 @@ def test_assessQuality(
...
+    def test_readQuality_mock_metrics_workspace_ingredients(
+        self, mockCalibrationMetricsWorkspaceIngredients  # noqa: ARG002
+    ):
+        run = MagicMock()
+        version = MagicMock()
+        calibRecord = CalibrationRecord.parse_raw(Resource.read("inputs/calibration/CalibrationRecord.json"))
+        self.instance.dataFactoryService.getCalibrationRecord = MagicMock(return_value=calibRecord)
+        with pytest.raises(Exception) as excinfo:  # noqa: PT011
+            self.instance.readQuality(run, version)
+        assert "The input table is empty" in str(excinfo.value)
```
walshmm, line 337: The test name should reflect that it is expecting an Exception. (negative case)

bolotovskyr: I've changed the name of that test. I've also split `test_readQuality` into 2 tests. It actually had 3 tests (i.e. 3 `readQuality` calls), of which 2 were testing for an exception. However, upon closer examination, one of the expected exceptions was from `pydantic` validation, so I removed that test as unnecessary.

walshmm (approved):

PR #173, opened 2023-12-06
minimal change should auto switch between vals based on guide

walshmm:
## Description of work

Depending on whether or not the instrument has a guide in place, the delTh needs to switch between two configured values.  This should be the most minimal solution to enable such.

## Explanation of work

I use the guideStat value stored on DetectorState to decide which delTh to return.

### Dev testing

pytests pass as normal

### CIS testing

TODO

## Link to EWM item

[EWM#3225](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3225)

### DIFF
--- a/src/snapred/backend/dao/state/DetectorState.py
+++ b/src/snapred/backend/dao/state/DetectorState.py
@@ -7,6 +7,7 @@ class DetectorState(BaseModel):
     arc: Tuple[float, float]
     wav: float
     freq: float
+    # Expected to only be 0 or 1, potentially should be a bool
     guideStat: int
     # two additional values that don't define state, but are useful
     lin: Tuple[float, float]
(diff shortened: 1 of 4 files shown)


src/snapred/backend/dao/state/DetectorState.py
```diff
@@ -7,6 +7,7 @@ class DetectorState(BaseModel):
     arc: Tuple[float, float]
     wav: float
     freq: float
+    # Expected to only be 0 or 1, potentially should be a bool
     guideStat: int
```
rboston628, line 11:
``` suggestion
    guideStat: Literal[0, 1]
```

rboston628: This way it's limited to the allowed values, but maybe if we want more possibilities we can easily expand it

walshmm: Did you one better, added an enum.  Through happenstance found out the valid values are apparently 1-2, the story is now updated.

rboston628 (approved):

PR #174, opened 2023-12-08
Refactor of PGP and Load Grouping

rboston628:
## Description of work

While working on Normalization Calibration backend, I ended up interacting with `PixelGroupingParametersCalculationAlgorithm`.  This calculation needs a grouping workspace, and was originally setup to load its own using `LoadGroupingDefinition`.  This is undesired behavior that needed to be refactored.

The algorithm also required an instrument filename.  After adding necessary sample logs, the instrument properties do not update, which causes invalid values in the calculation.  This was previously fixed by re-loading the instrument, requiring this file.  However, this should be unnecessary.  I refactored, for the moment, with another algorithm call which does not need to load anything.  However, when ~~[Mantid PR #37524](https://github.com/mantidproject/mantid/pull/36524)~~ new [mantid PR #36539](https://github.com/mantidproject/mantid/pull/36539) is merged into main, it will be possible to use that simpler method.

Refactoring above led to`LoadGroupingDefinition`, which needed some additional validations.  

Both of these algorithms also had unit tests that were notoriously slow.  Since the changes to the algorithm required updating the tests, I also tried to fix the speed.

In both cases, the slow tests were due to loading grouping files defined on the full SNAP resolution.  This is unnecessary for unit tests.  I wrote faster tests using only the test instrument and SNAPLite which run locally almost instantly.   The full SNAP tests run on analysis.

## Explanation of work

Pretty simple, removing calls to loading algorithms and removing inputs related to loading.

Then changing the tests to cover a bunch of cases on the test instrument.

### Dev testing

Check that the unit tests run, and run **faster**.

Open on analysis.  Run all unit tests.

Also open the GUI test panel.  Try running calibration.  I used run 58810, with convergence threshold 1, peak threshold 0.05, with silicon and the column grouping in lite mode.  It should run, ~~until it fails when trying to save (analysis issue, not problem with SNAPRed)~~ TO COMPLETION

### CIS testing

This is purely internal and does not require CIS testing.

## Link to EWM item

Came out of work done in 
[EWM#3293](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3293)


src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
```diff
@@ -44,64 +55,56 @@ def chopIngredients(self, instrumentState: InstrumentState) -> None:
         self.delLOverL = instrumentState.instrumentConfig.delLOverL
         self.L = instrumentState.instrumentConfig.L1 + instrumentState.instrumentConfig.L2
         self.delL = self.delLOverL * self.L
-        self.delTheta = instrumentState.delTh
-        self.grouping_ws_name = "pgpca_grouping_ws"
-        self.grouped_ws_name = "pgpca_grouped_ws"
-        self.resolution_ws_name = "pgpca_resolution_ws"
-        self.partial_resolution_group_ws_name = "pgpca_partial_resolution_group_ws"
+        self.delTheta = instrumentState.instrumentConfig.delThWithGuide
```
walshmm, line 58:
```suggestion
        self.delTheta = instrumentState.delTh
```
please change this back, needs to switch depending on guide status.

rboston628:
Note: the tests pass, but this change would actually break any service using pixel grouping parameters.  The service layer needs to be updated to include these changes.

rboston628:
I have fixed the services.  Calibration runs in the GUI test panel, indicating that PGP successfully ran, populated ingredients, and finished a diffraction calibration.

tests/unit/backend/recipe/algorithm/test_SaveGroupingDefinition.py
rboston628, line 1: These tests were already broken, but no one realized because they only run on analysis.

tests/unit/backend/recipe/test_CalibrationReductionRecipe.py
rboston628, line 1: This test is still broken, and only runs on analysis.  It has been broken for a very long time.

rboston628:
Fixed an issue with calPath inside diffcal recipe, now completes as follows:
image

rboston628:
Also checked assessment workflow.

Begins here:
Screenshot from 2023-12-12 16-24-59

Reaches this screen after successful calibration:
Screenshot from 2023-12-12 16-25-29

Will fail when trying to load the CIF path, because it is looking in Calibration_dynamic:
Screenshot from 2023-12-12 16-25-33

Fixing this is way outside the scope of this work.

src/snapred/backend/recipe/DiffractionCalibrationRecipe.py
```diff
@@ -47,13 +48,14 @@ def restockShelves(self, calibrationWS: str, maskWS: str = ""):
...
+        from datetime import date
+
         from mantid.simpleapi import SaveDiffCal
 
         from snapred.backend.data.LocalDataService import LocalDataService
 
-        lds = LocalDataService()
-        calibrationPath = lds._getCalibrationDataPath(self.runNumber)
-        filename = calibrationPath + "/difcal.h5"
+        LocalDataService()
```
walshmm, line 57:
```suggestion

```

walshmm (approved):

PR #176, opened 2023-12-11
Exception handlers and Workbench Text

walshmm:
## Description of work

This adds a decorator to wrap state validation interactions, such that if an exception occurs it will produce a relevant error message for the user. 

This also figues out why the text in workbench is red.  For some reason if you dont specify stdout in the stream object it goes to what I assume is stderror.

### Dev testing

Open workbench, see that the logs are no longer red.

### CIS testing

Open SNAPRed through workbench. Inspect log color.
Open Calibration Initialization window, attempt to proceed with runnumber `47582`
See that when it tries to init the state, that a pop up appears informing the user that this state is invalid. 
(The logs will have more details, i.e. guidestat is an invalid value.)


src/snapred/meta/decorators/ErrorHandler.py
```diff
@@ -0,0 +1,15 @@
...
+def StateExceptionHandler(func: callable):
+    @functools.wraps(func)
+    def inner(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except Exception as e:  # noqa: BLE001
+            raise StateValidationException(e)
+
+    return inner
+    # issubclass(func_annotations['request'].__args__[0], BaseModel)
```
dlcaballero16, line 15: if not needed, delete commented line

dlcaballero16 (approved):

PR #177, opened 2023-12-15
Grocery Service 🛒 🏪 

rboston628:
## Description of work

This completes the architecture for simply loading cached data within services by adding the additional `GroceryService`.  

The goal is to more easily facilitate acquiring and passing workspace arguments form the service layer.

It replaces the functionality of the `FetchGroceries` recipe with better dictionary methods for easier use inside client recipes.

Also included is the `GroceryListBuilder`, which can create individual list items or entire lists, and spit out a dictionary associating grocery items with names for use in client recipes or algorithms.

To use it, first build a list, then send the list to the service, and you will receive back a list of the items you requested, or a dictionary with their named keys.

The `FetchGroceriesRecipe` was reduced to just a wrapper on its algorithm.

The `GroceryService` has taken over the functionality of the `FetchGroceriesRecipe` and also of the `LocalDataWorkspaceService`.

## Explanation of work

Created the `GroceryListBuilder` for making more complicated grocery items.  It is capable of also adding the optional `propertyName` variable which allows for easier use inside dictionaries as, i.e. `{"InputWorkspace": groceries[0]}`.

The service has taken over the functionality `FetchGroceriesRecipe`, and provides some additional access to workspaces, such as a wrapper on `mtd` and a way to delete workspaces from the service layer.

The `LiteDataService` will load its own lite data map by a call to the `GroceryService`.  Its algorithm no longer loads the data map, as it is a transformation algorithm.  I could not remove the need to load the instrument.

The `DataFactoryService` was updated to use `GroceryService` methods, and the places using it updated accordingly.

### Dev testing

The unit tests for `GroceryListBuilder` should be pretty convincing, but in workbench you can try making some of your own.

In each CIS test script, please run at least the part that builds the grocery list.

### CIS testing

This can be tested from the GUI.

Open the test panel and run Diffraction Calibration on some data.  It should run to completion with no hiccups.  If there *are* hiccups, then the hiccups should be able about something like peak lists to PDCalibration, and not about loading data.

Then run Normalization Calibtation.  You will have to select Silicon as the material.  It should run to completion.

## Link to EWM item

[EWM#2481](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=2481)


src/snapred/backend/dao/ingredients/GroceryListItem.py
```diff
@@ -12,7 +10,7 @@ class GroceryListItem(BaseModel):
     Holds necessary information for a single item in grocery list
     """
```
walshmm, line 12: I would reccommend adding a 
```
def builder():
    return GroceryListBuilder()
```

So that code can just import GroceryListItem and do the following:
```
GroceryListItem.builder().nexus().native().using(item.runNumber).build()
```
w/o needing to import the builder separate.

rboston628: This creates a circular-import situation, as `GroceryListItem` then needs to import `GroceryListBuilder`, but `GroceryListBuilder` needs to import `GroceryListItem` :/

walshmm: if you move the import to the method does it still occur?

src/snapred/backend/dao/state/FocusGroupParameters.py
```diff
@@ -1,3 +1,5 @@
+# TODO delete
```
walshmm, line 1: Should this be deleted as part of this pr?

src/snapred/backend/service/CalibrationService.py
```diff
@@ -191,27 +193,17 @@ def diffractionCalibration(self, request: DiffractionCalibrationRequest):
...
+        self.groceryClerk.name("inputWorkspace").nexus(request.runNumber).useLiteMode(request.useLiteMode).add()
+        self.groceryClerk.name("groupingWorkspace").grouping(focusScheme).useLiteMode(
+            request.useLiteMode
+        ).fromPrev().add()
+        groceries = self.groceryService.fetchGroceryDict(self.groceryClerk.buildDict())
+        data = DiffractionCalibrationRecipe().executeRecipe(ingredients, groceries)
+
+        # save the calibration table
+        filename = f"{calpath}/SNAP_{request.runNumber}_difcal_{date.today().strftime('%Y%m%d')}.h5"
+        self.groceryService.writeDiffCalTable(filename, data["calibrationTable"])
```
walshmm, line 205: Saving of the diffcal table doesnt happen until after assessment.  Also paths generally wouldnt live here, that would be letting the implementation of the Data layer bleed into the service layer.  

I'm a bit uneasy being so specific as to mention nexus for similar reasons.

rboston628: I should probably change "nexus" to "neutrons" everywhere it occurs.  I was using it to indicate the workspace has neutron scattering data instead of a grouping definition (or other data).  I agree, confusing

src/snapred/backend/dao/state/FocusGroupParameters.py
rboston628: It had created some error last time I tried to delete it.  I'll try again.

src/snapred/backend/service/CalibrationService.py
```diff
@@ -74,7 +75,8 @@ def __init__(self):
         super().__init__()
         self.dataFactoryService = DataFactoryService()
         self.dataExportService = DataExportService()
-        self.groceryService = FetchGroceriesRecipe()  # TODO replace with service
+        self.groceryService = GroceryService()
+        self.groceryClerk = GroceryListItem.builder()
```
walshmm, line 79: Does clerk ever get reset or is it just building a bigger and bigger list of groceries every call?

walshmm: I see that it does reset [here](https://github.com/neutrons/SNAPRed/pull/177/files#diff-94ad68da3b204ecedce9e4af74b8fcb7cfbfb4e8f48d35f0bcc0e206eea4ad24R95), in my experience `build()` is terminal in the lifespan of a builder, youd just make a new one if you wanted to make a new object.

rboston628: The builder has the ability to create not just single items but also entire lists, and that is what leads it to this behavior.

The builder tracks two things, `_tokens` and `_list`.  `_tokens` is for building individual list items, and `_list` is for building an entire list.

The user can keep adding item specifications, then call either `build()` or `add()` to create the item, which also clears `_tokens`.  So a single builder can safely make an entire list of items using the `add()` method.  So the builder needs to be reusable.

If they use `add()`, then later a command `buildList()` or `buildDict()` will return a list of all the items made, and in the process clears `_list`.   It has to clear `_list`, or else the user could do the thing you're describing, with unintentional ever-increasing grocery lists of workspaces.

Because it has to be safely reusable, it might as well be reused.  That was my thought, anyway.

walshmm (approved):

PR #178, opened 2023-12-15
Fix AddSampleLog in PGP calculation

rboston628:
## Description of work

In `PixelGroupingParametersCalculationAlgorithm`, it was necessary to make a call to `ws->populateInstrumentParameters()` from within mantid.  The only algorithms which achieved this were loading algorithms.  As a temporary solution, a mantid algorithm which could load instrument parameters from a "file" (the file here being an empty XML string) was used as a kludge.

Since [Mantid PR #36539](https://github.com/mantidproject/mantid/pull/36539), it is now possible to use `AddSampleLog` directly to update the instrument parameters.

## Explanation of work

It does nothing but call `AddSampleLog` with the parameter set to update the instrument, and deletes the previous kludgey call.

### Dev testing

If PGP calculation tests run, then the instrument was successfully updated, giving the correct values in the comparison calculation.

### CIS testing

This is internal and does not need CIS testing.

## Link to EWM item

[EWM#3301](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3301)

Also related to 
[EWM#3280](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3280)

Useswork from [Mantid PR#36539](https://github.com/mantidproject/mantid/pull/36539)

### DIFF
--- a/src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
+++ b/src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
@@ -148,6 +148,11 @@ class PixelGroupingParametersCalculationAlgorithm(PythonAlgorithm):
         detectorState = instrumentState.detectorState
 
         # add sample logs with detector "arc" and "lin" parameters to the workspace
+        # NOTE after adding the logs, it is necessary to update the instrument to
+        #  factor in these new parameters, or else calculations will be inconsistent.
+        #  This is done with a call to `ws->populateInstrumentParameters()` from within mantid.
+        #  This call only needs to happen with the last log
+        logsAdded = 0
         for param_name in ["arc", "lin"]:
             for index in range(2):
                 self.mantidSnapper.AddSampleLog(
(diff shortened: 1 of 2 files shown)


src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
```diff
@@ -161,20 +161,11 @@ def loadNeededLogs(self, ws_name: str, instrumentState: InstrumentState):
...
-        # TODO remove the below after uncommenting above.
-        minimalXML = "<parameter-file>></parameter-file>"
-        self.mantidSnapper.LoadParameterFile(
-            "Calling an algorithm that includes a populateInstrumentParamters and no file load",
+        self.mantidSnapper.AddSampleLog(
+            "Update the instrument",
             Workspace=ws_name,
-            ParameterXML=minimalXML,
+            LogName="update_instrument",
+            UpdateInstrumentParameters=True,
```
peterfpeterson, line 168: You should get rid of these lines and put the `UpdateInstrumentParameters=True` in the `AddSampleLog` on L153. Calling it an extra time with a log that doesn't belong is silly.

peterfpeterson (changes requested):
You should change the `environment.yml` file to require a new enough mantid
```
- mantidworkbench>6.8.20231213
```

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
rboston628: It doesn't need to update the instrument parameters until after all four sample logs have been added, so it's kind of a waste of compute cycles to have it update the instrument within the for loop.

peterfpeterson (approved):

PR #179, opened 2023-12-15
Multi config key sub

walshmm:
## Description of work

This enables multilevel key substitution in the yaml.

What that means is if the supplied key terminates at a map for a given string, you may continue to add keys to the end of your lookup string to solve the true template value.

This is useful in the case that you want 2 roots for a given value uniformly across a set of keys, i.e. separate read and write directories.

## Explanation of work

As above, but see tests for concrete examples

### Dev testing

tests pass, try to make your own .

### CIS testing

none.

### DIFF
--- a/src/snapred/meta/Config.py
+++ b/src/snapred/meta/Config.py
@@ -135,18 +135,35 @@ class _Config:
             self.refresh(f"{env_name}.yml", clearPrevious)
 
     # method to regex for string pattern of ${key} and replace with value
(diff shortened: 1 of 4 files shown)


walshmm:
Syntactically it may be a good idea to prepend these sorts of keys with some symbol to make it more obvious these arent terminal value substitution.

ie. `$persistence`

rboston628 (approved):

PR #190, opened 2024-01-04
diffraction calibration: reference-pixel selection

ekapadi:
## Description of work

The first stage of the diffraction-calibration process involves the calculation and correction for the relative offsets of pixels in each detector group.  Ideally, each pixel should be measuring the same spectrum, although rescaled by a factor (DIFC) corresponding to the offset of that pixel.  Relative offsets  are calculated by taking the cross correlation of each pixel's spectrum with that of a reference pixel, and then calculating the shift of the cross-correlation peaks.  Since detector grouping assignment is arbitrary, it is important that the reference pixel be selected from within a detector group.  As it happens, this within-group selection is more important for the synthetic data used in the unit tests, than for data generated by the real instrument.  However, it should be noted that out-of-group selection might result in selection of a reference pixel currently assigned to _no_ group, in either case.

For synthetic test data used in the diffraction-calibration unit tests, the previous reference-pixel selection always resulted in out-of-group selection, as these groups included an _even_ number of pixels with non-consecutive workspace indices.  In addition, as each group only included detectors at exactly the _same_ physical offset, this caused significant changes in the behavior of the calibration process, in comparison to in-group reference pixel selection.  As noted previously, this should not be an issue on the real instrument.

## Explanation of work

The previous reference-pixel selection used `int(median(<group: pixel indices>))`: for even order groups containing non-consecutive indices this will select an out-of-group index.  The new selection method involves rounding the _effective_ _index_ of the _median_, and selecting the pixel at that index.

## To test

A new unit test has been added.  CIS-testing will also be required.

### Dev testing

A new unit test has been added to `PixelDiffractionCalibration` test class.  This test thoroughly covers a range of possible group-index sets, but by choice, does not test the empty set, which pathology should raise an exception elsewhere in the process.

### CIS testing

The current `tests/cis_tests/new_diffcal_script.py` is to be used for CIS testing.  No changes in behavior should be observed -- if there's any question, compare the output to that from the current `SNAPRed/next` branch, using the same script.

## Link to EWM item

Current defect number: [EWM#3336](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=3336)

Dependent story: [EWM#3464](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=3464)


ekapadi:
**Note that this PR does not yet deal with any aspect of masking, which does relate to reference-pixel selection, and additionally requires actually keeping track of the reference-pixel id from loop-to-loop** -- all of that's covered in the "diffraction-calibration: combine masks" PR (updated soon).

ekapadi:
**Note that this PR does not yet deal with any aspect of masking, which does relate to reference-pixel selection, and additionally requires actually keeping track of the reference-pixel id from loop-to-loop** -- all of that's covered in the "diffraction-calibration: combine masks" PR (updated soon).

tests/unit/backend/recipe/algorithm/test_PixelDiffractionCalibration.py
```diff
@@ -202,6 +204,19 @@ def test_reexecution_and_convergence(self):
...
 
+    def test_reference_pixel_selection(self):
+        """Test that the selected reference pixel is always a member of the detector group."""
+        algo = ThisAlgo()
+
+        # Test even and odd order groups;
+        #   test consecutive and non-consecutive groups.
+        # Do not test empty groups.
+        N_detectors = 10
+        dids = list(range(N_detectors))
```
rboston628, line 215: Can you add an extra test that includes a gap in `dids`?  For instance 
``` python
dids = [1, 2, 4, 5]
```
and make sure it doesn't return `3`?

rboston628: Sorry, and one more with e.g. `dids = [1, 2, 11]` and make sure it returns `2`?

rboston628 (approved):

rboston628 (approved):

PR #192, opened 2024-01-08
Diffraction calibration: combine masks

ekapadi:
## Description of work

During the diffraction-calibration process, masking information is accumulated for failing pixels.  Failing pixels are identified both during the offset-computation stage, and also during the diffraction-focussing optimization stage.  All of this masking information must be combined. 

## Explanation of work

Most of the implementation changes required to combine pixel masking information were in the _Mantid_ codebase, and have already been successfully merged. With respect to the _SNAPRed_ codebase, the present commit includes the following changes:

* Definition of 'MaskWorkspace' as an _optional_ `Output` parameter of both `PixelDiffractionCalibration` and `GroupDiffractionCalibration`.  As documented, the present `Output` parameter definition covers the use case of applying information from a specified workspace that already exists.
* Proper treatment of the 'MaskWorkspace' parameter in the `DiffractionCalibrationRecipe`: -- support for an _mandatory_ 'MaskWorkspace' `Output` parameter; if the workspace already exists, its masking information will be applied;  Note that this parameter _must_ be initialized in order to properly pass masking information between the diffraction-calibration processing stages; -- passing of the 'MaskWorkspace' parameter between the `PixelDiffractionCalibration` and `GroupDiffractionCalibration` processing steps;
* Generation of the default 'OutputWorkspace' and 'MaskWorkspace' names in the `GroceryList` using the new `WorkspaceNameGenerator` system.
* Implementation of a new system for generating synthetic data to be used for the diffraction-calibration tests.  Note that many of the previous versions of the tests, when masking information was examined after execution, were failing for almost every pixel.  This new synthetic data system can be generalized (although it has not yet been generalized) to allow _automatic_ use with almost any instrument and its associated grouping definitions.
* Synthetic-data preparation classes and routines and associated test-helper routines are now located at `tests/util` in the SNAPRed codebase.  **Please let me know if you think there's a better location for these routines.**
* **Additional notes:  I was not quite sure how to deal with the possibility of an _incoming_ 'MaskWorkspace' parameter in `DiffractionCalibrationRequest`, and it is also possible that the FetchGroceries system needs additional modification to deal with such a parameter.  If required, at this point due to the extent of this current PR, all of this should be added to a _new_ story.**

## To test
A number of new _unit_ tests have been added at `test_PixelDiffractionCalibration`, and `test_GroupDiffractionCalibration`.  In addition, _existing_ unit tests have been modified so that _masking_ is verified where appropriate.  Note that the new synthetic-data preparation provides more-realistic data, that passes all of the execution tests without _any_ pixels failing -- this then allows pixels and groups to be _selectively_ failed, in order to test masking behavior.  Note: additional tests may be required to actually test the synthetic-data and test-helper methods -- at present, these are tested _implicitly_ in that the associated tests will fail if the methods do not work as required.

### CIS testing
A modified version of `new_diffcal_script.py` is at `diffcal_masking_script.py`: this script adds sections specifically to allow convenient setup for testing masking behavior.

## Link to EWM item

[EWM#3464](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=3464)

### DIFF
--- a/src/snapred/backend/dao/ingredients/GroceryListItem.py
+++ b/src/snapred/backend/dao/ingredients/GroceryListItem.py
@@ -66,8 +66,7 @@ class GroceryListItem(BaseModel):
             case "diffcal_output" | "diffcal_table" | "diffcal_mask":
                 if v.get("runNumber") is None:
                     raise ValueError(f"diffraction-calibration {v['workspaceType']} output requires a run number")
-                if v.get("isOutput") == False:
-                    raise ValueError(f"diffraction-calibration {v['workspaceType']} output is special-order only")
-            case _:
-                raise ValueError(f"unrecognized 'workspaceType': '{v['workspaceType']}'")
+                if v.get("isOutput") is False:
+                    raise ValueError(f"diffraction-calibration {v['workspaceType']} is special-order only")
+            case _: raise ValueError(f"unrecognized \'workspaceType\': \'{v['workspaceType']}\'")
         return v
--- a/src/snapred/backend/dao/ingredients/GroceryListItem.py
+++ b/src/snapred/backend/dao/ingredients/GroceryListItem.py
@@ -10,7 +10,7 @@ class GroceryListItem(BaseModel):
     Holds necessary information for a single item in grocery list
     """
 
-    workspaceType: Literal["neutron", "grouping", "diffcal"]
+    workspaceType: Literal["neutron", "grouping", "diffcal", "diffcal_output", "diffcal_table", "diffcal_mask"]
     useLiteMode: bool  # indicates if data should be reduced to lite mode
     # optional loader:
     # -- "" tells FetchGroceries to choose the loader
@@ -27,6 +27,10 @@ class GroceryListItem(BaseModel):
     keepItClean: bool = True
     # name the property the workspace will be used for
     propertyName: Optional[str]
+    # flag to indicate if this is an _output_ workspace:
+    # an output workspace will not be loaded,
+    # it may or may not already exist in the ADS
+    isOutput: bool = False
 
     def builder():
         # NOTE this import is here to avoid circular dependencies -- don't bother trying to move it
@@ -36,22 +40,33 @@ class GroceryListItem(BaseModel):
 
     @root_validator(pre=True, allow_reuse=True)
     def validate_ingredients_for_groceries(cls, v):
-        if v["workspaceType"] == "neutron":
-            if v.get("runNumber") is None:
-                raise ValueError("Loading neutron data requires a run number")
-            if v.get("groupingScheme") is not None:
-                v["groupingScheme"] = None
-        if v["workspaceType"] == "grouping":
-            if v.get("groupingScheme") is None:
-                raise ValueError("you must specify the grouping scheme to use")
-            if v["groupingScheme"] == "Lite":
-                # the Lite grouping scheme reduces native resolution to Lite mode
-                v["instrumentPropertySource"] = "InstrumentFilename"
-                v["instrumentSource"] = str(Config["instrument.native.definition.file"])
-                v["useLiteMode"] = False  # the lite data map only works on native data
-            else:
-                if v.get("instrumentPropertySource") is None:
-                    raise ValueError("a grouping workspace requires an instrument source")
-                if v.get("instrumentSource") is None:
-                    raise ValueError("a grouping workspace requries an instrument source")
+        match v["workspaceType"]:
+            case "neutron":
+                if v.get("runNumber") is None:
+                    raise ValueError("Loading neutron data requires a run number")
+                if v.get("groupingScheme") is not None:
+                    v["groupingScheme"] = None
+            case "grouping":
+                if v.get("groupingScheme") is None:
+                    raise ValueError("you must specify the grouping scheme to use")
+                if v["groupingScheme"] == "Lite":
+                    # the Lite grouping scheme reduces native resolution to Lite mode
+                    v["instrumentPropertySource"] = "InstrumentFilename"
+                    v["instrumentSource"] = str(Config["instrument.native.definition.file"])
+                    v["useLiteMode"] = False  # the lite data map only works on native data
+                else:
+                    if v.get("instrumentPropertySource") is None:
+                        raise ValueError("a grouping workspace requires an instrument source")
+                    if v.get("instrumentSource") is None:
+                        raise ValueError("a grouping workspace requires an instrument source")
+            case "diffcal":
+                if v.get("runNumber") is None:
+                    raise ValueError(f"diffraction-calibration input table workspace requires a run number")            
+            # output (i.e. special-order) workspaces
+            case "diffcal_output" | "diffcal_table" | "diffcal_mask":
+                if v.get("runNumber") is None:
+                    raise ValueError(f"diffraction-calibration {v['workspaceType']} output requires a run number")
+                if v.get("isOutput") == False:
+                    raise ValueError(f"diffraction-calibration {v['workspaceType']} output is special-order only")
+            case _: raise ValueError(f"unrecognized \'workspaceType\': \'{v['workspaceType']}\'")
         return v
--- a/src/snapred/backend/dao/ingredients/GroceryListItem.py
+++ b/src/snapred/backend/dao/ingredients/GroceryListItem.py
@@ -68,5 +68,6 @@ class GroceryListItem(BaseModel):
                     raise ValueError(f"diffraction-calibration {v['workspaceType']} output requires a run number")
                 if v.get("isOutput") is False:
                     raise ValueError(f"diffraction-calibration {v['workspaceType']} is special-order only")
-            case _: raise ValueError(f"unrecognized \'workspaceType\': \'{v['workspaceType']}\'")
+            case _:
+                raise ValueError(f"unrecognized 'workspaceType': '{v['workspaceType']}'")
         return v
--- a/src/snapred/backend/dao/ingredients/GroceryListItem.py
+++ b/src/snapred/backend/dao/ingredients/GroceryListItem.py
@@ -61,12 +61,13 @@ class GroceryListItem(BaseModel):
                         raise ValueError("a grouping workspace requires an instrument source")
             case "diffcal":
                 if v.get("runNumber") is None:
-                    raise ValueError(f"diffraction-calibration input table workspace requires a run number")            
+                    raise ValueError("diffraction-calibration input table workspace requires a run number")
             # output (i.e. special-order) workspaces
             case "diffcal_output" | "diffcal_table" | "diffcal_mask":
                 if v.get("runNumber") is None:
                     raise ValueError(f"diffraction-calibration {v['workspaceType']} output requires a run number")
                 if v.get("isOutput") == False:
                     raise ValueError(f"diffraction-calibration {v['workspaceType']} output is special-order only")
-            case _: raise ValueError(f"unrecognized \'workspaceType\': \'{v['workspaceType']}\'")
+            case _:
+                raise ValueError(f"unrecognized 'workspaceType': '{v['workspaceType']}'")
         return v
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -90,6 +90,18 @@ class GroceryService:
         instr = "lite" if useLiteMode else "native"
         return Config["grouping.workspacename." + instr] + groupingScheme
 
+    def _createDiffcalWorkspaceName(self, runId: str):
+        return wng.diffCal().runNumber(runId).build()
+
+    def _createDiffcalOutputWorkspaceName(self, runId: str):
+        return wng.diffCalOutput().runNumber(runId).build()
+
+    def _createDiffcalTableWorkspaceName(self, runId: str):
+        return wng.diffCalTable().runNumber(runId).build()
+
+    def _createDiffcalMaskWorkspaceName(self, runId: str):
+        return wng.diffCalMask().runNumber(runId).build()
+        
     ## WRITING TO DISK
 
     def writeWorkspace(self, path: str, name: WorkspaceName):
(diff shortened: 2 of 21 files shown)


src/snapred/backend/recipe/algorithm/GroupDiffractionCalibration.py
```diff
@@ -106,6 +112,21 @@ def unbagGroceries(self):
...
+        else:
+            self.outputWStof = self.getPropertyValue("OutputWorkspace")
+
+        self.maskWS: str = ""
+        if self.getProperty("MaskWorkspace").isDefault:
+            self.maskWS = wng.diffCalMask().runNumber(self.runNumber).build()
+            self.setProperty("MaskWorkspace", self.maskWS)
+        else:
+            self.maskWS = self.getPropertyValue("MaskWorkspace")
+
```
walshmm, line 129: I think I lean in favor of just making these fields required.  They arent user facing, and the dependence on the input data to determine a good default name seems to indicate they should be supplied.

src/snapred/backend/recipe/algorithm/PixelDiffractionCalibration.py
```diff
@@ -72,15 +77,22 @@ def chopIngredients(self, ingredients: Ingredients) -> None:
...
+        # create a default name for output calibration table
         self.DIFCpixel: str = ""
         if self.getProperty("CalibrationTable").isDefault:
-            self.DIFCpixel = f"_DIFC_{self.runNumber}"
+            self.DIFCpixel = wng.diffCalTable().runNumber(self.runNumber).build()
             self.setProperty("CalibrationTable", self.DIFCpixel)
         else:
             self.DIFCpixel = self.getPropertyValue("CalibrationTable")
 
+        # create a default name for output mask workspace
```
walshmm, line 88: Same as previous comment on required outputworkspace names.  I dont think I've encountered a situation where we only need partial results from out algo's, and I would imagine we we have to refactor and split them if it is indeed required.

tests/util/diffraction_calibration_synthetic_data.py
```diff
@@ -0,0 +1,300 @@
...
+
+Peak = namedtuple("Peak", "centre sigma height")
+
+
+def random_seed(bits: int) -> int:
+    # Generate a random seed of 'bits' length.
+    return secrets.randbits(bits)
+
+
+class SyntheticData(object):
```
walshmm, line 47: Should this be a [test fixture](https://docs.pytest.org/en/6.2.x/fixture.html)?  Same to the other items in the utils module

tests/unit/backend/recipe/test_DiffractionCalibrationRecipe.py
```diff
@@ -25,87 +15,36 @@
...
 
 ThisRecipe: str = "snapred.backend.recipe.DiffractionCalibrationRecipe"
 PixelCalAlgo: str = ThisRecipe + ".PixelDiffractionCalibration"
 GroupCalAlgo: str = ThisRecipe + ".GroupDiffractionCalibration"
 
 
-class TestDiffractionCalibtationRecipe(unittest.TestCase):
+class TestDiffractionCalibrationRecipe(unittest.TestCase):
     def setUp(self):
-        self.fakeRunNumber = "555"
```
walshmm, line 36: Very much like what SytheticData is doing here :smile:

walshmm:
> Additional notes: I was not quite sure how to deal with the possibility of an incoming 'MaskWorkspace' parameter in DiffractionCalibrationRequest, and it is also possible that the FetchGroceries system needs additional modification to deal with such a parameter. If required, at this point due to the extent of this current PR, all of this should be added to a new story.

I dont think there should be an incoming mask for the Diffraction Calibration flow as a whole?

walshmm:
(separate work)
Though this has made me realize that perhaps we are not persisting the Diffraction Calibration correctly, as we are not using [SaveDiffCal](https://docs.mantidproject.org/nightly/algorithms/SaveDiffCal-v1.html) for the table, group, and mask, but are instead saving them separately.

src/snapred/backend/recipe/DiffractionCalibrationRecipe.py
```diff
@@ -95,11 +107,15 @@ def executeRecipe(self, ingredients: Ingredients, groceries: Dict[str, str]) ->
         groupedAlgo.setProperty("GroupingWorkspace", self.groupingWS)
         groupedAlgo.setProperty("Ingredients", ingredients.json())
         groupedAlgo.setProperty("PreviousCalibrationTable", self.calTable)
+        groupedAlgo.setProperty("MaskWorkspace", self.maskWS)
         groupedAlgo.setProperty("FinalCalibrationTable", self.calTable)
         try:
             groupedAlgo.execute()
+            # TODO: these output workspaces are *optional*: why are they included here without checking?
+            #   And, how to check if they're still default values?
```
rboston628, line 115: If you make them mandatory, then the mantid API will (errantly) demand they already exist in the ADS.  The way to ensure they exist is to add an additional method,
``` python
def validateInputs(self) -> Dict[str, str]:
    errors = {}
    if self.getProperty("OutputWorkspace").isDefault:
        errors["OutputWorkspace"] = "error message here"
    return errors
```
This will be called as part of the mantid algorithm lifecycle and will throw a runtime error including the key and the message.  Later algorithms in SNAPRed have examples of using this.  See for instance [`FetchGroceriesAlgorithm`](https://github.com/neutrons/SNAPRed/blob/04d31430dcd422cd164a4c0cde0e70515bae5e7c/src/snapred/backend/recipe/algorithm/FetchGroceriesAlgorithm.py#L75C1-L85C22).

src/snapred/backend/recipe/algorithm/GroupDiffractionCalibration.py
rboston628: I concur, they should simply be required.  See my comment above on how to do that.

tests/util/diffraction_calibration_synthetic_data.py
```diff
@@ -0,0 +1,300 @@
...
+    """_Fixed_ implementation of synthetic data to be used for all diffraction-calibration tests."""
+
+    RANDOM_SEED = 311379324803478887040360489933500222827
```
rboston628, line 50: `0xDEADBEEF`?

tests/util/diffraction_calibration_synthetic_data.py
```diff
@@ -0,0 +1,300 @@
...
+            InputFile=self.fakeGroupingFilePath,
+            InputWorkspace=rawWS,
+            OutputWorkspace=groupingWS,
+        )
+
+        focWS = mtd[groupingWS]
+        detectors = focWS.detectorInfo()
+        allXmins = [0] * 16
+        allXmaxs = [0] * 16
+        allDelta = [0] * 16
```
rboston628, line 250: this `16` should probably be replaced with `TEST_INSTRUMENT_RESOLUTION`

tests/util/diffraction_calibration_synthetic_data.py
```diff
@@ -0,0 +1,300 @@
...
+            OutputWorkspace=rawWS,
+            WorkspaceType="Histogram",
+            Function="User Defined",
+            UserDefinedFunction=functionString,
+            Xmin=overallDMin,
+            Xmax=overallDMax,
+            BinWidth=dBin_abs,
+            XUnit="dSpacing",
+            NumBanks=4,  # must produce same number of pixels as fake instrument
+            BankPixelWidth=2,  # each bank has 4 pixels, 4 banks, 16 total
```
rboston628, line 230: I'm not sure how to easily replace this 4 and this 2...

tests/unit/backend/recipe/test_DiffractionCalibrationRecipe.py
```diff
@@ -25,87 +15,36 @@
...
-                GroupPeakList(groupID=3, peaks=peakList, maxfwhm=0.01),
-                GroupPeakList(groupID=7, peaks=peakList, maxfwhm=0.02),
-                GroupPeakList(groupID=2, peaks=peakList, maxfwhm=0.02),
-                GroupPeakList(groupID=11, peaks=peakList, maxfwhm=0.02),
-            ],
-            convergenceThreshold=0.5,
-            calPath=Resource.getPath("outputs/calibration/"),
-            pixelGroup=fakeInstrumentState.pixelGroup,
-        )
+        self.syntheticInputs = SyntheticData()
```
rboston628, line 28: Should this be inside `setUpClass`?  Or is it better to re-run it before each test?

tests/unit/backend/recipe/algorithm/test_PixelDiffractionCalibration.py
```diff
@@ -1,153 +1,50 @@
...
-            calPath=Resource.getPath("outputs/calibration/"),
-            convergenceThreshold=1.0,
-            maxOffset=self.maxOffset,
-            pixelGroup=fakeInstrumentState.pixelGroup,
-        )
-
-        self.fakeRawData = f"_test_pixelcal_{self.fakeRunNumber}"
-        self.fakeGroupingWorkspace = f"_test_pixelcal_difc_{self.fakeRunNumber}"
-        self.makeFakeNeutronData(self.fakeRawData, self.fakeGroupingWorkspace)
+        inputs = SyntheticData()
```
rboston628, line 27: Should this be in `setUpClass`?

tests/unit/backend/recipe/algorithm/test_PixelDiffractionCalibration.py
```diff
@@ -191,17 +89,179 @@ def test_reexecution_and_convergence(self):
...
         assert x != 0.0
         assert x > 0.0
-        assert x <= self.maxOffset
+        assert x <= self.fakeIngredients.maxOffset
 
         # check that value converges
-        numIter = 5
+        # WARNING: testing for three iterations seems to be about the limit here.
+        #   At greater than 3 iterations, there are small oscillations about a limit value.
+        maxIter = 3
```
rboston628, line 97: The part `max(1.0e-4, allOffsets[-2])` was supposed to account for small oscillations about a limit value.  If the median offset is oscillating, but oscillations are smaller than this limited value, then it shouldn't be an issue if it runs 5 times.  Perhaps the `1.0e-4` needs to be adjusted?

src/snapred/backend/recipe/algorithm/GroupDiffractionCalibration.py
```diff
@@ -170,6 +183,9 @@ def PyExec(self) -> None:
         - PreviousCalibrationTable: str -- the name of the table workspace with previous DIFC values
         output:
         - OutputWorkspace: str -- the name of the diffraction-focussed d-spacing data after the final calibration
+        - MaskWorkspace: str -- the name of the mask workspace for detectors failing calibration (1.0 => dead-pixel, 0.0 => live-pixel)
+            when the mask workspace already exists, incoming masked values will be combined with any new masked values detected during
+            execution
```
rboston628, line 188: I believe indentations in this block are causing the docs build to fail

tests/util/diffraction_calibration_synthetic_data.py
ekapadi: I left you a separate message via slack about this.  **Yes, this should be a fixture, but that doesn't make sense right now, considering our mixed `unittest` / `pytest` test environment.**

src/snapred/backend/recipe/DiffractionCalibrationRecipe.py
ekapadi: This is an interesting issue.  `Output` parameters _may_ exist, or may not.  The Mantid documentation does mention that an `Output` workspace which already exists should be used (i.e. as "input") by an algorithm.  I believe that this is an alternative, and also correct way that this issue can be dealt with.  That is, as an _optional_ `Output` parameter.  This is what we did for the corresponding masking changes on the Mantid side (e.g. in `GetDetectorOffsets` and `PDCalibration`).  If the user has not specified the parameter, it is deleted at the end of execution.

tests/unit/backend/recipe/test_DiffractionCalibrationRecipe.py
ekapadi: **Tests may be run in parallel, although we generally do not do that.  I'm implementing all of my own tests as if this will be the case.**  We gain a _little_ bit of run time by caching this synthetic data, and cloning the workspaces that are required as input (that's done now in `test_GroupDiffractionCalibration`).  However, this adds a lot of complexity for very little gain, and makes things more difficult to maintain.  Here I'm adopting a compromise, and just regenerating the data.  As our test _instrument_ becomes more complicated, perhaps we should then change this.

tests/unit/backend/recipe/algorithm/test_PixelDiffractionCalibration.py
ekapadi: I simply "ran out of time" on this one.  I came up with a compromise setting that works.  As Malcolm mentioned, it's unrealistic here to try and fit data to > 0.5 of a bin or so.  (That means that 1.0e-4, which is 1.0e-4 of a bin, is completely artificial.)  There are major issues that happened during refactoring which changed all of our bin widths.  I'll also point out that the previous test data was over 50% masked during actual execution, and that `GetDetectorOffsets` returns _zero_ when it fails to find a fit.  My point here, is that the previous value wasn't actually _working_ at all.

src/snapred/backend/recipe/algorithm/GroupDiffractionCalibration.py
ekapadi: I'll try that and see if it solves the issue.

src/snapred/backend/recipe/algorithm/GroupDiffractionCalibration.py
ekapadi: Please see my comment below.  These don't necessarily have to be `Input` parameters, they could also be optional (or mandatory) `Output` parameters.

tests/unit/backend/recipe/algorithm/test_PixelDiffractionCalibration.py
ekapadi: **A key issue is that these tests are designed to be successfully run in parallel.**  We don't do this yet, but it's one of the better ways to reduce our test suite execution time.  Note that there are two versions of how this data is used:  (1) `test_PixelDiffractionCalibration` (here), and (2) `test_GroupDiffractionCalibration`, which does use `setUpClass` and needs to _clone_ the input data in many of the tests.  
It was beyond the scope of this story to attempt to _unify_ the entire diffraction-calibration test suite.  :)

src/snapred/backend/dao/ingredients/GroceryListItem.py
```diff
@@ -54,4 +58,20 @@ def validate_ingredients_for_groceries(cls, v):
                     raise ValueError("a grouping workspace requires an instrument source")
                 if v.get("instrumentSource") is None:
                     raise ValueError("a grouping workspace requries an instrument source")
+        # output (i.e. special-order) workspaces
+        if v["workspaceType"] == "diffcal_output":
+            if v.get("runNumber") is None:
+                raise ValueError("diffraction-calibration data output requires a run number")
+            if v.get("isOutput") == False:
+                raise ValueError("diffraction-calibration data output is special-order only")
```
walshmm, line 66: could these be moved to a helper method? Looks like they happen 3 time

src/snapred/backend/data/GroceryService.py
```diff
@@ -343,6 +352,14 @@ def fetchGroceryList(self, groceryList: List[GroceryListItem]) -> List[Workspace
...
                     item.instrumentSource = prev
                 res = self.fetchGroupingDefinition(item)
+            # for output (i.e. special-order) workspaces
+            elif item.workspaceType == "diffcal_output":
+                res = {"result": True, "workspace": self._createDiffcalOutputWorkspaceName(item.runNumber)}
+            elif item.workspaceType == "diffcal_table":
+                res = {"result": True, "workspace": self._createDiffcalTableWorkspaceName(item.runNumber)}
+            elif item.workspaceType == "diffcal_mask":
+                res = {"result": True, "workspace": self._createDiffcalMaskWorkspaceName(item.runNumber)}
+
```
walshmm, line 362: I would like to see this elif chain refactored to a dict of callables at some point but wont burden this pr with it.

src/snapred/backend/dao/ingredients/GroceryListItem.py
rboston628: I think this is a situation where inheritance would help.  We really have a bunch if different kinds of grocery list items, one kind of item for neutron data, one for diffcal tables, one for grouping workspaces, etc.

tests/resources/inputs/testInstrument/fakeSNAPFocGroup_Localized.xml
rboston628, line 1: I still think this should be more properly called `fakeSNAPFocGroup_Bank.xml`, since it corresponds to what a "bank" grouping is.  See for instance [CreateGroupingWorkspace](https://docs.mantidproject.org/nightly/algorithms/CreateGroupingWorkspace-v1.html)

ekapadi: I changed this to 'Bank'.  Interestingly, this definition does not seem to be "in use" anywhere at present.

src/snapred/backend/dao/ingredients/GroceryListItem.py
ekapadi: These are good comments.  I'd recommend that all of this wait until the next time this class gets refactored.  **My own "two cents" here is that it seems overly difficult to add new workspace types (unless someone is _very_ familiar with this class).  I'd prefer to see all of this automatically implemented from a YML file somewhere.**

src/snapred/backend/data/GroceryService.py
ekapadi: Ironically, I had that same thought, but was trying to be consistent with the existing implementation.  :(

ekapadi: I ended up changing these to "match / case" statements.

walshmm:
codecove failed upload, gonna rerun it

walshmm (approved):

PR #196, opened 2024-01-09
fix the issue with lite mode output

rboston628:
## Description of work

This fixes a defect spotted with running the lite data algorithm through the `new_diffcal_script`

It also addresses some issues noted by Michael within normalization calibration.

It is unknown and probably unknowable how on earth we got this far into the process without noticing this.

## Explanation of work

For reasons known mystical and arcane, it is required to update the property with a pointer to the workspace in the ADS (the same pointer) at the end of the algorithm.  So we must set the thing back to itself, otherwise the mantid algorithm lifecycle will undo all of our work at the end of the algorithm.

Basically, any algorithm which uses

``` python
outputWS = self.mantidSnapper.mtd[self.getPropertyValue("OutputWorkspace")]
```
then edits `outputWS` using API function calls, must at some point perform
``` python
self.setProperty("OutputWorkspace", outputWS)
```

### CIS testing
``` python
from mantid.simpleapi import mtd

from snapred.backend.dao.ingredients.GroceryListItem import GroceryListItem
from snapred.backend.data.GroceryService import GroceryService

#User input ###########################
badRunNumbers = ["57514", "58812"]  # known to have failed before
isLite = True
Config._config["cis_mode"] = False
LITE_DATA_RESOLUTION = 18432
#######################################

clerk = GroceryListItem.builder()
grocer = GroceryService()

for x in badRunNumbers:
    badList = clerk.neutron(x).lite().buildList()
    badWorkspace = grocer.fetchGroceryList(badList)[0]
    print(badWorkspace)
    assert mtd[badWorkspace].getNumberHistograms() == LITE_DATA_RESOLUTION
```

## Link to EWM item

[EWM#3553](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3553)

### DIFF
--- a/src/snapred/backend/recipe/algorithm/DiffractionSpectrumWeightCalculator.py
+++ b/src/snapred/backend/recipe/algorithm/DiffractionSpectrumWeightCalculator.py
@@ -121,6 +121,8 @@ class DiffractionSpectrumWeightCalculator(PythonAlgorithm):
                 mask_indices = np.where(np.logical_and(x > peak.position.minimum, x < peak.position.maximum))
                 weights[mask_indices] = 0.0
             weight_ws.setY(index, weights)
+        
+        self.setProperty("WeightWorkspace", weight_ws)
 
         if isEventWorkspace:
             self.mantidSnapper.ConvertToEventWorkspace(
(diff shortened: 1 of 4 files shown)


mguthriem (approved):
This was tested with the previously problematic run 57514 and the changes have fixed the issue observed previously.

PR #197, opened 2024-01-10
Remove dependency on calib log files

rboston628:
### CIS testing

``` python
```

## Link to EWM item

[EWM#2912](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=2912)

### Work split off
 - PR #202
 - PR #203
 - PR #206 
 - PR #209 
 - PR #212


rboston628:
After merger of PR #212 , there is no longer any dependency on the calib logs.  This PR retains some stray and incomplete refactors of algorithms, which we have decided do not need to be addressed in this phase.  So this PR can be closed for now.

PR #198, opened 2024-01-10
allows workspace list to stretch

walshmm:
## Description of work
Lets the workspace list stretch left and all over, however the buttons move a little and dont looks as stylish when stretched.  
Function over Form I suppose.

## To test

Open the gui and stretch the window


rboston628:
Does what it says on the tin
image

rboston628 (approved):

PR #201, opened 2024-01-15
Grouping-schema map implementation.

ekapadi:
## Description of work
The grouping-schema map provides a centralized resource, allowing the CIS to specify which grouping files are applicable to instrument states in each of the `lite` or `native` instrument modes.  Using the `FocusGroup` class, this map also allows the association of a simple _name_ to each of these grouping files.  A default version of this grouping-schema map is placed at the `<instrument.calibration.powder.grouping.home>` location.  At initialization of a new `InstrumentState` version tree, a copy of this file is written to the `<state root>` directory.

## Explanation of work
Grouping-schema map, DAO implementation:
  The changes of this commit include:
  * Implementation of class `GroupingMap`, providing access to maps of `FocusGroup` applicable to a given `InstrumentState`;
  * Separate `FocusGroup` maps are provided for each of `native` and `lite` modes;
  * The associated JSON representation of the `GroupingMap` uses separate _lists_ to provide an easy to edit format for by-hand editing (by the CIS);
  * During its intialization process, only _valid_ `FocusGroup` are added to a `GroupingMap` instance;
  * Valid `FocusGroup` paths must exist, and must have a valid file format;
  * Invalid `FocusGroup` found during initialization are _logged_ as `warning`, but initialization is allowed to continue;
  * A relative path in a `FocusGroup` is referenced with respect to <instrument.calibration.powder.grouping.home> from `Config`;
  * An absolute path in a `FocusGroup` is allowed, and will accepted without modification, provided that the path is a valid `FocusGroup` path;

Grouping-schema map, connection to `InstrumentState` and `LocalDataService`:

  * `GroupingMap` has a `stateId` attribute, allowing association to only one <instrument.state.root> (which actually, may be the parent of several _versions_ of the state);
  * The <default grouping map> at <instrument.calibration.powder.grouping.home> has a special <default grouping-map state id>;
  * A new `ObjectSHA` class enforces required behavior of the <state id>, including: -- a standard hex-digest format: 16-character, lowercase hex string; -- initialization, or read and write of the state-id from a simple string; -- optional (though non-persistent) retention of the sometimes-provided `decodedKey` field (used nowhere at present);
  * `GroupingMap` is added as an attribute to `InstrumentState`: -- implementation of an _explicit_ method `attachGroupingMap` with optional coercion of the `GroupingMap.stateId` to match that of the `InstrumentState`; -- use of a _root_validator_ to enforce that `InstrumentState.groupingMap.stateId` == `InstrumentState`.id;
  * A state's `GroupingMap` has a separate JSON file to that of the `InstrumentState`: -- appropriate methods were added to `LocalDataService` to implement this file-separation behavior: note that this includes calling `InstrumentState.attachGroupingMap` in many existing _read_ methods; -- required modifications to existing tests in `LocalDataService`; -- the names of several tests were updated to include the behavior of what the tests are actually testing; hopefully this will help _future_ maintainers;
  * During the initialization of an `InstrumentState`: -- read of the <default grouping map> from the correct location at <instrument.calibration.powder.grouping.home>; -- write of the grouping map to the correct location at <state root> (i.e. at the _root_ of the state's version tree); -- `GroupingMap` write only occurs only _once_ at initializion of a _new_ state version tree (and _overwrite_ is not allowed);

### Dev testing
Unit tests were added at `test_GroupingMap.py` and at `test_LocalDataService.py`.  The former test the basic `GroupingMap` implementation, and in combination with the latter, all of the key _behavior_ requirements of the story are verified.

### CIS testing
  * A copy of `defaultGroupingMap.json` should be placed at <instrument.calibration.powder.grouping.home> on the SNS filesystem (the filename must not be modified);
  * Any applicable grouping-file paths should then be added to the JSON lists contained in this file; for grouping files residing in <instrument.calibration.powder.grouping.home> these files can be added as simple filenames; note that the `stateId` field should _not_ be modified; this field has a special value reserved for this _default_ version of the grouping-schema map;
  * Using the SNAPRed GUI, any operation that _initializes_ an `InstrumentState` should be performed;

  Following a new state initialization: a new `groupingMap.json` should have been placed at the _root_ of the state version tree.  The contents of this JSON should match that of the `defaultGroupingMap.json`, except for the `stateId` field, which should have been updated to match that of the new state.

## Link to EWM item

[EWM#3484](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=3484)


src/snapred/backend/dao/StateId.py
```diff
@@ -1,4 +1,9 @@
+import hashlib
```
walshmm, line 1: should this file be removed?  Seems like ObjectSha is replacing it.

src/snapred/backend/dao/state/InstrumentState.py
```diff
@@ -1,29 +1,95 @@
...
+    id: ObjectSHA
+
     instrumentConfig: InstrumentConfig
     detectorState: DetectorState
     gsasParameters: GSASParameters
     particleBounds: ParticleBounds
     pixelGroup: Optional[PixelGroup]
+
+    # 'groupingMap' (not really Optional), has its own separate JSON file
+    groupingMap: Optional[GroupingMap] = Field(default=None, exclude=True)
```
walshmm, line 36: Shouldnt this live on something like StateConfig/InstrumentConfig?  It applies to all calibrations for a given state, but InstrumentState is a member of a particular instance of Calibration.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -490,12 +486,19 @@ def readNormalizationRecord(self, runId: str, version: str = None):
             latestFile = self._getLatestFile(recordPath)
         record: NormalizationRecord = None  # noqa: F821
         if latestFile:
+            print(f"reading NormalizationRecord from {latestFile}")
```
walshmm, line 489: log?

src/snapred/backend/data/LocalDataService.py
```diff
@@ -490,12 +486,19 @@ def readNormalizationRecord(self, runId: str, version: str = None):
...
+            # Read the state's grouping-schema map from its separate JSON file.
+            groupingMap = self._readGroupingMap(stateId)
+            # Attach the grouping-schema map to the normalization record's state:
+            record.normalization.instrumentState.attachGroupingMap(groupingMap)
         return record
 
     def writeNormalizationRecord(self, record: NormalizationRecord, version: int = None):  # noqa: F821
         """
         Persists a `NormalizationRecord` to either a new version folder, or overwrite a specific version.
+        -- side effect: updates version numbers of incoming `NormalizationRecord` and its nested `Normalization`.
```
walshmm, line 501: Is this really a side effect?  Regardless I do think its a great idea to explicitly call them out if they are unavoidable.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -514,9 +523,11 @@ def writeNormalizationRecord(self, record: NormalizationRecord, version: int = N
         self.writeNormalizationState(runNumber, record.normalization, version)
         for workspace in record.workspaceNames:
             self.groceryService.writeWorkspace(normalizationPath, workspace)
+        print(f"wrote NormalizationRecord: version: {version}")
```
walshmm, line 526: log? x2

src/snapred/backend/data/LocalDataService.py
```diff
@@ -527,12 +538,20 @@ def readCalibrationRecord(self, runId: str, version: str = None):
         # read the file
         record: CalibrationRecord = None
         if latestFile:
+            print(f"reading CalibrationRecord from {latestFile}")
```
walshmm, line 541: log? (x3)

src/snapred/backend/data/LocalDataService.py
```diff
@@ -541,16 +560,26 @@ def writeCalibrationRecord(self, record: CalibrationRecord, version: int = None)
...
+        # However, if it is going to be _nested_, this marks it with the correct version.
+        # (For example, use pydantic Field(exclude=True) to _stop_ nesting it.)
+        record.calibrationFittingIngredients.version = version
+
         calibrationPath = self._constructCalibrationDataPath(runNumber, version)
         # check if directory exists for runId
         if not os.path.exists(calibrationPath):
             os.makedirs(calibrationPath)
-        # append to record and write to file
+        # append to record and write to file # *** DEBUG *** .version==2, .<calibration>.version==1
```
walshmm, line 574: debug?

src/snapred/backend/data/LocalDataService.py
```diff
@@ -541,16 +560,26 @@ def writeCalibrationRecord(self, record: CalibrationRecord, version: int = None)
...
         write_model_pretty(record, recordPath)
 
-        self.writeCalibrationState(runNumber, record.calibrationFittingIngredients, version)
+        self.writeCalibrationState(
+            runNumber, record.calibrationFittingIngredients, version
+        )  # *** DEBUG *** <calibration>.version==2
```
walshmm, line 579: debug? (x2)

src/snapred/backend/data/LocalDataService.py
```diff
@@ -541,16 +560,26 @@ def writeCalibrationRecord(self, record: CalibrationRecord, version: int = None)
...
         for workspace in record.workspaceNames:
             self.groceryService.writeWorkspace(calibrationPath, workspace)
+        print(f"Wrote CalibrationRecord: version: {version}")
```
walshmm, line 582: log? (x4)

src/snapred/backend/dao/state/InstrumentState.py
ekapadi: Yes.  I had been concerned about that as well.  Let me think about this a bit...

src/snapred/backend/data/LocalDataService.py
ekapadi: Actually, that wasn't one of ours.  :)   I'll take a look and make sure there aren't any more of them in `LocalDataService`.

src/snapred/backend/data/LocalDataService.py
ekapadi: Well, it's a side effect, in the sense that the primary purpose of the method is to _write_ the record.  However, I'd be glad to change the wording.  The only reason I added this comment is that it took me way too long to figure out why the comparison of records at the end of the test was failing (**the "nested" attribute ended up with the incorrect version**).

src/snapred/backend/data/LocalDataService.py
ekapadi: Missed in "cleanup".  To be removed...

src/snapred/backend/data/LocalDataService.py
walshmm: I see, well then experience dictates this is a necessary callout then.  Wording is fine as is.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -210,9 +210,10 @@ def _readPVFile(self, runId: str):
         return f
 
     @ExceptionHandler(StateValidationException)
-    def _generateStateId(self, runId: str) -> Tuple[Any, Any]:
+    def _generateStateId(self, runId: str) -> Tuple[str, str]:
         if runId in self.stateIdCache:
-            return self.stateIdCache[runId]
+            SHA = self.stateIdCache[runId]
+            return SHA.hex, SHA.decodedKey
```
rboston628, line 216: This return value isn't necessary if you use an `if/else`

src/snapred/backend/dao/state/InstrumentState.py
```diff
@@ -1,29 +1,95 @@
...
+        # 'groupingMap' has its own separate JSON file
+        return None
+    """
+
+    """ 'field_serializer' is not present in pydantic 1.10: see  = Field(..., exclude=True) above
+    @field_serializer('groupingMap')
+    def do_not_write_groupingMap(self, v: Any) -> None:
+        # 'groupingMap' has its own separate JSON file
+        return None
+    """
```
rboston628, line 95: You can overwrite the `json()` method to properly serialize it

src/snapred/backend/dao/ObjectSHA.py
```diff
@@ -0,0 +1,43 @@
...
+    def __str__(self):
+        return self.hex
+
+    @staticmethod
+    def fromObject(obj: Any):
+        # In combination with the Field validation above: this enforces the format of the SHA as a string.
+        hasher = hashlib.shake_256()
+        decodedKey = json.dumps(obj.__dict__).encode("utf-8")
+        hasher.update(decodedKey)
+        return ObjectSHA(hex=hasher.digest(8).hex(), decodedKey=decodedKey)
```
rboston628, line 43: Is this different from the `parse_obj`usual method?  Could this be written to overwrite `parse_obj`?

src/snapred/backend/dao/StateConfig.py
```diff
@@ -15,7 +16,14 @@ class StateConfig(BaseModel):
     rawVanadiumCorrectionFileName: str  # Needs to be removed when Normalization changes go in
     vanadiumFilePath: str  # Needs to be removed when Normalization changes go in
     focusGroups: List[FocusGroup]  # PixelGroupingParameters
-    stateId: str  # generated.
+    stateId: ObjectSHA  # generated.
```
rboston628, line 19: Is this guaranteed to be the same as the `_generateStateId` method in `LocalDataService`?

src/snapred/backend/dao/StateId.py
```diff
@@ -1,4 +1,9 @@
...
+import json
 from dataclasses import dataclass
+from typing import ClassVar
+
+from pydantic import Field
```
rboston628, line 6: It doesn't look like these new imports are being used

rboston628:
I can resolve the merge conflicts, if you want me to

src/snapred/backend/data/LocalDataService.py
rboston628, line 1: In `_readFocusGroups` at L141, is it possible to extract the focus groups from your GroupingMap object?

tests/unit/backend/dao/test_GroupingMap.py
```diff
@@ -0,0 +1,87 @@
...
+            GroupingMap.GroupingMap, "calibrationGroupingHome", lambda: Path(Resource.getPath("inputs/pixel_grouping"))
+        )
+        with caplog.at_level(logging.WARNING):
+            GroupingMap.GroupingMap.parse_obj(
+                {
+                    "stateId": "deadbeef00000004",
+                    "liteFocusGroups": [{"name": "RelativePath", "definition": "SNAPFocGroup_Column.xml"}],
+                    "nativeFocusGroups": [{"name": "RelativePath", "definition": "SNAPFocGroup_Column.xml"}],
+                }
+            )
```
rboston628, line 72: Should there be an `assert` here checking the path is relative to the default location?

src/snapred/backend/dao/state/GroupingMap.py
```diff
@@ -0,0 +1,98 @@
...
+    def native(self) -> Dict[str, FocusGroup]:
+        return self._nativeMap
+
+    @property
+    def isDefault(self):
+        return self.stateId == GroupingMap.defaultStateId
+
+    @property
+    def isDirty(self) -> bool:
+        return self._isDirty
```
rboston628, line 48: I don't think that "dirty" applies to groupings, as they should never be edited within SNAPRed for any reason.

src/snapred/backend/dao/StateConfig.py
```diff
@@ -14,5 +16,51 @@ class StateConfig(BaseModel):
...
+
+    def _validate(self: BaseModel):
+        # Manually trigger _root_validator_.
+        *_, validation_error = validate_model(self.__class__, self.__dict__)
+        if validation_error:
+            raise validation_error
+
+    def attachGroupingMap(self, groupingMap: GroupingMap, coerceStateId=False):
+        # Attach a grouping map to the InstrumentState:
```
dlcaballero16, line 34: Comment should be updated to say "StateConfig" instead of "InstrumentState"

src/snapred/backend/dao/state/InstrumentState.py
ekapadi: Moved to `StateConfig`.

src/snapred/backend/data/LocalDataService.py
ekapadi: I ended up moving most of these to `logger.info(...)`.  I still am confused about why they were their in the first place.  :(

src/snapred/backend/data/LocalDataService.py
ekapadi: That's correct.  However, I'm following the "coding standard" of using _early_ return rather than deep if / else chaining.  It's interesting, as when I learned software development, we were _strongly_ encouraged to have only one return statement per function.

ekapadi: **And, possibly more importantly, I'm trying to modify the original code as little as possible.**

src/snapred/backend/dao/ObjectSHA.py
ekapadi: Interesting question.  **The intent here was to modify the original code as little as possible.**

src/snapred/backend/dao/StateConfig.py
ekapadi: There's only _one_ `_generateStateId` method, and it's in `LocalDataService`.  That should be the only way a `stateId` is ever filled in.  **Note that `_generateStateId` still generates exactly the same hex digest as it did before.**  The only thing our changes in this PR do is _standardize_ the hex digest to a 16-charater lowercase hex string.

src/snapred/backend/dao/StateId.py
ekapadi: Removed.

src/snapred/backend/data/LocalDataService.py
walshmm: For me it depends, early escape cases like caching and validation I think are acceptable.

src/snapred/backend/data/LocalDataService.py
ekapadi: The only call to `_readFocusGroups` is in `DataFactoryService.getFocusGroups`.  Existing classes which call `getFocusGroups` (and there are only a few), should instead directly access the `StateConfig.groupingMap` attribute.  **Implementing all of these interconnections was beyond the scope of this PR's story.**

tests/unit/backend/dao/test_GroupingMap.py
ekapadi: That test is done in `GroupingMap`'s `root_validator`.

src/snapred/backend/dao/state/GroupingMap.py
ekapadi: `GroupingMap` is not a <grouping>.  The `GroupingMap` itself is "dirty" if its `stateId` has been modified from the default.  In this case, it must be written out to <state root>.  This only happens at initialization of a new state version tree, and the `GroupingMap` is only written once, and never allowed to be overwritten.

walshmm:
There still seems to be a merge conflict

src/snapred/backend/dao/request/FocusSpectraRequest.py
```diff
@@ -0,0 +1,11 @@
+from matplotlib import use
+from pydantic import BaseModel
+
+
+class FocusSpectraRequest(BaseModel):
```
walshmm, line 5: is this a merge hiccup?  This file already exists: https://github.com/neutrons/SNAPRed/blob/next/src/snapred/backend/dao/request/FocusSpectraRequest.py

walshmm (approved):

PR #205, opened 2024-01-18
Revert "diffraction calibration: xMin, xMax parameters"

rboston628:
Reverts neutrons/SNAPRed#191

This seems to be the cause of an error  when running the GUI, resulting in bounds of 0 on `GetDetectorOffsets`.

I'm still not entirely sure what the issue is.


rboston628:
I was getting the following error related to Xmin, Xmax in this algorithm
Screenshot from 2024-01-18 17-22-04

However, after inspection, it was not related to this, and was an issue with the peaks lists.

PR #208, opened 2024-01-19
Builder decorator

walshmm:
## Description of work

Decorate your favorite pydantic model and give it a go.
Helps if you need to keep a base model in an intermediate state!

## To test

tests pass


rboston628 (approved):
It looks good to me

PR #203, opened 2024-01-16
one true peak ingredients

rboston628:
There can be only one.

## Description of work

This work is split off from PR #197 

The algorithm `DetectorPeakPredictor` requires as inputs:
 - an `InstrumentState`
 - a `PixelGroup` (currently being pulled off of the `InstrumentState`)
 - a `CrystallographicInfo`
 - a float specifying peak threshold

There are four algorithms which need these same ingredients, only because they directly or indirectly call `DetectorPeakPredictor`:
 - `DiffractionSpectrumWeightCalculator` calls directly, doesn't use the ingredients for any other reason
 - `PurgeOverlappingPeaksAlgorithm` calls directly, doesn't use the ingredients for any other reason
 - `FitMultiplePeaksAlgorithm` calls indirectly through `PurgeOverlappingPeaks`, only needs `peakType`
 - `SmoothDataExcludingPeaksAlgo` calls indirectly through `DiffractionSpectrumWeightCalculator`, only needs `smoothingParameter`

Not all of these are required to call `DetectorPeakPredictor`, because in some cases it is possible to pass a list of detector peaks instead of populating them from the algorithm.

At the moment, the ingredients needed to `DetectorPeakPredictor` are split up, so that it is necessary to separately pass around an `InstrumentState` and  a `CrystallographicInfo`, just in case they might be needed.  In addition, the float for peak threshold needs to be populated, and in some places resorts to a default value.  

Because it is not necessary to remove `PixelGroup` from `InstrumentState`, there are two ways to do this.
1. add another DAO being passed through all of these algorithms and their associated recipes
2. make a single ingredients DAO for `DetectorPeakPredictor` holding the needed information

The cleaner and simpler way to do this is 2.  This allows for removing several other ingredients types, which only existed to eventually pass the correct objects to `DetectorPeakPredictor`:

 - `FitCalibrationWorkspaceIngredients` carried around 
   - an `InstrumentState`, 
   - a `CrystallographicInfo`, 
   - a list of `PixelGroupingParameters` (apparently this was not refactored for `PixelGroup`), 
   - a workspace name (does not need to be in ingredients), 
   - an optional float called `smoothingParameter` in case the smoothing algorithm ever needed to be called.
 - `FitMultiplePeaksIngredients` carried around
   - an `InstrumentState`
   - a `CrystallographicInfo`
   - a string for peak type, validated against a list
   - an input workspace name (does not need to be in ingredients)
   - a bool called `showResults`
[...]

PR #209, opened 2024-01-19
Sous Chef :man_cook: 

rboston628:
## Description of work

Just look at the changed methods in the calibration and normalization service, and the changes in the CIS tests, and I think it speaks for itself.

This is needed for EWM 2912, which requires untangling the handling of ingredients within the data service, in particular removing `pixelGroup` from off of `instrumentState`.  The logistics of doing so impacted multiple service endpoints that needed a `pixelGroup`, which almost always relied on a `calibration.instrumentState.pixelGroup` object.

## Explanation of work

The SousChef takes over a lot of functionality that was being handled by the data service, or haphazard methods inside the calibration service.  The data service should be focused on reading/writing to disk.  Now the SousChef takes over orchestrating all the pieces needed to make most ingredient types.

The `FarmFreshIngredients` comprise the minimal list of parameters needed to make basically everything.  These are comparable to the user inputs found in the prototype scripts or in the CIS test scripts.  From these, plus some smaller adjustments, all ingredients can be easily created.

The two new ingredients, `NormalizationIngredients` and `PeakIngredients`, sill show up in PR #203.

Many places that called for only a focus group name (aka grouping schema) or a focus group path were changed to required a `FocusGroup` object instead, which has both.  The required a slight adjustment inside the workflows.

Separating ingredients in this way makes apparent a defect in the Reduction workflow, which was setup to depend only in run number, but in fact depends on lite mode and on the focus group chosen.

### Dev testing

Run in calibration in the GUI with run ~~46680~~ 58882.

Run in normalization in the GUI with runs 58810, 58813.

Also try running the CIS tests.

### CIS testing

This is an internal issue and does nto require CIS testing.

## Link to EWM item

[EWM#2912](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=2912)


src/snapred/backend/data/LocalDataService.py
```diff
@@ -749,7 +749,16 @@ def readGroupingFiles(self):
...
         if len(groupingFiles) < 1:
             raise RuntimeError(f"No grouping files found in {groupingFolder} for extensions {extensions}")
-        return groupingFiles
+        focusGroups = {}
+        for file in groupingFiles:
+            focusGroups[file] = FocusGroup(
+                name=self.groupingSchemaFromPath(file),
+                definition=file,
+            )
+        return focusGroups
```
walshmm, line 761: this section should be its own method that consumes readGroupingFiles rather than hijacking its output

src/snapred/backend/service/ServiceFactory.py
```diff
@@ -35,6 +36,7 @@ def __init__(self):
         self.serviceDirectory.registerService(ApiService)
         self.serviceDirectory.registerService(LiteDataService)
         self.serviceDirectory.registerService(NormalizationService)
+        self.serviceDirectory.registerService(SousChef)
```
walshmm, line 39: its only necessary to register ui facing services

src/snapred/backend/data/LocalDataService.py
rboston628: Can do. This was just the minimal change I needed to touch the frontend as little as possible.

rboston628: Should be better now

walshmm (approved):

PR #220, opened 2024-01-29
patch problems with xtal rx

rboston628:
## Description of work

There was an apparent error inside `SousChef` when creating `NormalizationIngredients` from the `NormalizationService.vanaciumCorrection` method, where the `FarmFreshIngredients` did not have a CIF path.  This adds a safety check just before the xtal service is called, to ensure CIF path is present and to populate it from the calibrant sample path if it is now.

### CIS testing

``` python
```

## Link to EWM item


darshdinger (approved):
Looks good. Thanks for your help!

PR #223, opened 2024-01-31
Properly save 'vanadium correction'

dlcaballero16:
## Description of work

Appended workspace names to NormalizationRecord in save step of the Normalization Calibration Workflow
## Explanation of work

The error was being caused by workspaceNames not being set in NormalizationRecord, so I appended the required workspaces in the save step in the UI workflow. I also updated the field to initialize by default to an empty list to prevent NoneType errors from happening.

### Dev testing

To test this, make sure to fix your environment on SNS first. For us non CIS people, copy /SNS/SNAP/shared/Calibration to your home drive, and update the application.yml file to point to your copy, specifically the instrument.home field. This gets around the permissions issue. On the SNAPRed gui, open the test panel and click on the Normalization Calibration tab. For the fields:

Run Number: 58813
Background Run Number: 58813
Smoothing Parameter: 0.5
Sample: Silicon_NIST_640D_001
Grouping File: Any should work

Click continue and wait for calibration to finish. There might be a warning, just click OK on the box. On the Tweak Parameters tab, just click continue. On the saving tab, click continue and you should see no errors. The saved files should now appear under <your home drive>/Calibration/Powder/<shaID>/v_<#>

### CIS testing

Do the dev test, but you shouldn't have to fix your env beforehand, assuming you have write permissions.

## Link to EWM item

[EWM#<3813>](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=<3813>)

### DIFF
--- a/src/snapred/backend/service/NormalizationService.py
+++ b/src/snapred/backend/service/NormalizationService.py
@@ -150,6 +150,7 @@ class NormalizationService(Service):
             backgroundRunNumber=request.backgroundRunNumber,
             smoothingParameter=request.smoothingParameter,
             normalization=normalization,
+            workspaceNames=[request.workspace],
         )
         return record
 
--- a/src/snapred/backend/service/NormalizationService.py
+++ b/src/snapred/backend/service/NormalizationService.py
@@ -150,7 +150,6 @@ class NormalizationService(Service):
             backgroundRunNumber=request.backgroundRunNumber,
             smoothingParameter=request.smoothingParameter,
             normalization=normalization,
-            workspaceNames=[request.workspace],
         )
         return record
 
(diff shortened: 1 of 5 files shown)


src/snapred/backend/service/NormalizationService.py
```diff
@@ -150,6 +150,7 @@ def normalizationAssessment(self, request: NormalizationCalibrationRequest):
             backgroundRunNumber=request.backgroundRunNumber,
             smoothingParameter=request.smoothingParameter,
             normalization=normalization,
+            workspaceNames=[request.workspace],
```
walshmm, line 153: I dont think request.workspace gets used right?

walshmm (approved):

PR #224, opened 2024-01-31
Matplotlib extra features proof of concept

rboston628:
# DO NOT MERGE THIS

This is only here as a starting point for future work in [EWM 3730](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=3730)

Made for [EWM 3815](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=3815)


rboston628:
Was able to get the zoom feature.  Most other features imported from the workbench toolbar are not operation and may require further imports.

The graph is correctly showing the correct units, by default.  I was unable to generate error (1) described in EWM 3730.

The `normalize_by_bin_width` flag to the `plot` function is doing its job.

The below graph has set `normalize_by_bin_width = False`.  Note the y-axis is in "counts".

Screenshot from 2024-01-31 15-20-14

The below graph has set `normalize_by_bin_width = True`.  Note the y-axis is in "counts ($\mu$ s ${}^{-1}$)", and that the heights and relative heights are different. 

Screenshot from 2024-01-31 15-18-45

Notice both graphs show the x-axis units, and that the toolbar appears at the top.

rboston628:
The relevant features have been added to the view, so this PR is not longer needed as a reference.

PR #225, opened 2024-01-31
Quick fix to resource CIF file

rboston628:
## Description of work

The SNAPRed resources folder (in src/snapred, not the one in tests) holds a number of default files used for setting up the GUI and pre-populating necessary fields inside of JSON form views.

One of these files, `not_real.cif` was needed as the CIF file for the `test_id123.json` file, which contained information for a calibrant sample.  This file was not fully implemented, and the work done here was to complete the default CIF file.

## Explanation of work

Added some needed information to the `not_real.cif` file inside the application resources directory.

### Dev testing

This should not require testing, but ensure all unit tests still work.

### CIS testing

This is an internal issue and does not require CIS testing.

## Link to EWM item

There is no related EWM, this is something quick I noticed while working on another ticket.


ekapadi:
Not a blocking comment, but I'm wondering if we could name the file in a manner that is more descriptive about what it's for:  for example `defaultFormat.cif` or `genericForm.cif`?

ekapadi (approved):

PR #226, opened 2024-02-02
Fix plot views in normalization calibration

rboston628:
## Description of work

There were some visual issues present in the normalization graphs.

1. There was no navigation toolbar, which prevented zooming
2. Only the array of y-values was being plotted, so that the x-axis showed index instead of d-spacing
3. The graphs were not being normalized by binwidth, which is the mantid default
4. Extraneous errors were being logged into the workbench console.

Errors 1-3 were easily fixed, following work in PR #224.

Error 4 proved has related to
1. an older version of mantid
2. an `Axis` object being garbage-collected, while mantid workbench retained the object's event listener, triggering the errors.

## Explanation of work

Edited the normalization graph display view.

It will display a grid of graphs, and try to optimize the layout of the graphs based on the number of groups, trying to make things as square as possible.

The navigation toolbar works, but some of the workbench-specific features do not seem to work.  The navigation toolbar should probably be replaced with the standard matplot toolbar.

I needed to mock out the workbench.plotting module in the docs config in order for the docs to build.

### Dev testing

On analysis, pull this branch and launch workbench.  Open the SNAPRed interface.  Try running normalization with
- run number = 58813
- background run number = 58813 (just use the same one)
- smoothing = 0.5 (or literally anything)
- sample = any of the vanadiums
- grouping = Column

The graphs will be flat, but that isn't important.

Make sure the graphs are visible, labels don't overlap, the the size is reasonable.

On the navigation toolbar, click the zoom icon.  Try dragging a rectangle over one of the graphs: it should zoom in on the region selected.

Switch grouping from Column to All.  After a while, it should display a single graph.

Make sure no errors were logged in the console during this process.

### CIS testing

Do the same as above, but look at realistic data runs.

## Link to EWM item

[EWM#3730](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3730)

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -63,6 +63,7 @@ with mock.patch("builtins.__import__", side_effect=import_mock):
         "mantid.plots.plotfunctions",
         "mantid.plots.datafunctions",
         "mantid.plots.utility",
+        "workbench.plotting",
         "workbench.plugins.workspacewidget",
     ]
 
--- a/src/snapred/ui/view/SpecifyNormalizationCalibrationView.py
+++ b/src/snapred/ui/view/SpecifyNormalizationCalibrationView.py
@@ -41,7 +41,7 @@ class SpecifyNormalizationCalibrationView(QWidget):
         self.signalBackgroundRunNumberUpdate.connect(self._updateBackgroundRunNumber)
 
         # create the graph elements
-        self.figure = plt.figure(figsize=(50, 50))
+        self.figure = plt.figure(constrained_layout=True)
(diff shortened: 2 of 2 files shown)


mguthriem:
I just tested this with @rboston628  and all functionality was working as expected.

walshmm:
image
Can confirm default size is not scrunched, zooming and panning works. No logging errors.

walshmm (approved):

PR #229, opened 2024-02-08
Refactor normalization output

dlcaballero16:
## Description of work

This updates the output of the save step in the Normalization GUI.  dMin and SmoothingParameter now exist inside of NormalizationRecord, and NormalizationParameters is no longer created as it was redundant. The output is now also saved in it's own normalization folder. 
 ## Explanation of work

LocalDataService was updated so that Normalization uses it's own functions, as opposed to using the same functions that Diffraction Calibration does. Changes in here also make it so that the saved output are now under .../<stateId>/normalization/v_* . SousChef was updated to handle specific Normalization case.
### Dev testing

To test this, make sure to fix your environment on SNS first. For us non CIS people, copy /SNS/SNAP/shared/Calibration to your home drive, and update the application.yml file to point to your copy, specifically the instrument.home field. This gets around the permissions issue. On the SNAPRed gui, open the test panel and click on the Normalization Calibration tab. Click the Check Calibration Initialization button first. This can generate the required file to start the Normalization Process. For the fields:

Run Number: 58813
Background Run Number: 58813
Smoothing Parameter: 0.5
Sample: Silicon_NIST_640D_001
Grouping File: Any should work

Click continue and wait for calibration to finish. There might be a warning, just click OK on the box. On the Tweak Parameters tab, just click continue. On the saving tab, click continue and you should see no errors. The saved files should now appear under /Calibration/Powder/<stateId>/normalization/v_<#>. The files should now be 3 different nxs files and a NormalizationRecord.json file. In the record file, you should now see the dMin and the SmoothingParameter fields. 

### CIS testing

Do the dev test, but you shouldn't have to fix your env beforehand, assuming you have write permissions.

## Link to EWM item

[EWM#3814](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3814)

### DIFF
--- a/src/snapred/backend/dao/normalization/NormalizationRecord.py
+++ b/src/snapred/backend/dao/normalization/NormalizationRecord.py
@@ -12,3 +12,4 @@ class NormalizationRecord(BaseModel):
     normalization: Normalization
     workspaceNames: Optional[List[str]]
     version: Optional[int]
+    dMin: float
--- a/src/snapred/backend/dao/normalization/NormalizationRecord.py
+++ b/src/snapred/backend/dao/normalization/NormalizationRecord.py
@@ -2,14 +2,14 @@ from typing import List, Optional
 
 from pydantic import BaseModel
 
-from snapred.backend.dao.normalization.Normalization import Normalization
+from snapred.backend.dao.calibration.Calibration import Calibration
 
 
 class NormalizationRecord(BaseModel):
     runNumber: str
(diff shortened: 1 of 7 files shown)


walshmm:
image
image
looks good 👍

walshmm (approved):

PR #231, opened 2024-02-13
Patch ever-growing window heights

rboston628:
## Description of work

In PR #226 I foolishly coded the window heights to keep growing.  This led to unexpected view issues when Malcolm tried testing other functionality.

This fix will prevent the window from growing increasingly when the figures are populated.

## Explanation of work

I had previously used the window height, and added to it the figure height.  This changed the window height, so that the next time this was called the window got even larger, and larger, and larger.

This saves the initial window height and adds the figure height to that, so that the new window height remains pretty stable.

### Dev testing

In GUI run normalization calibration
- 58813
- 58813
- vanadium 001
- 0.5
- Column

Then try changing dMin, smoothing parameter, grouping, and verify that the window stays a reasonable height.

### CIS testing

``` python
```

## Link to EWM item

### DIFF
--- a/src/snapred/ui/view/SpecifyNormalizationCalibrationView.py
+++ b/src/snapred/ui/view/SpecifyNormalizationCalibrationView.py
@@ -106,6 +106,9 @@ class SpecifyNormalizationCalibrationView(QWidget):
 
         self.layout.setRowStretch(1, 3)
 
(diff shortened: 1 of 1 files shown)


mguthriem (approved):
have tested changes and it fixed the resizing problem.

PR #222, opened 2024-01-31
Ewm3742 fix d min empty array

darshdinger:
## Description of work
To address this issue, modifications were made to both the frontend and backend components. The objective was to correct a problem where the `dMin` parameter, when assigned extremely high values, did not yield plausible outcomes for the smoothing operation. The expected result should have been a plot displayed to the user without any cubic smoothing applied. This defect pertained to the `Normalization` Workflow.

## Explanation of work
For the purpose of explaining the solution in a clear and cohesive way, I will split this into frontend changes and backend changes.

### Backend Changes:
In the backend, substantial updates were implemented in the `CrystallographicInfo` structure and in the classes related to the `SmoothDataExcludingPeaks` algorithm. The changes in the `request` sub-directory largely entailed renaming variables to clearly differentiate between the sets of `dMin / dMax` values. This was done to avoid confusion between `dMin / dMax`, which pertain to pixel grouping parameters, and `crystalDMin / crystalDMax`, which are relevant to crystallographic data.

Further alterations were carried out in the `recipe/algorithm` sub-directory, particularly to the `SmoothDataExcludingPeaksAlgo` and its dependent algorithms, to fully allow for a functional fix to this defect. In the `CrystallographicInfoAlgorithm`, at line 40, the function `CreateWorkspace` was updated to `CreateSingleValueWorkspace` to streamline the process, as this workspace is solely utilized for storing crystallographic information. In the `DetectorPeakPredictor`, a line was inserted at line 65 to append a value, ensuring the array remains non-empty for subsequent threshold calculations. Furthermore, within `DiffractionSpectrumWeightCalculator`, the implementation of an `if` statement now accurately addresses cases where the list of peaks returned is empty, cloning the cached inputWorkspace and creating a weightWorkspace with all values being set to zero. This new weightWorkspace is recieved by `SmoothDataExcludingPeaksAlgo` where a boolean (`allZeros`) is added to check for the weight _ws consisting of zero values for all indexes within X. `If` this is true then the outputted smoothedWorkspace is this weightWorkspace containing all zero X and Y values leading to the frontend plot showing a flat line y = 0 for all x.

### Frontend Changes:
*NOTE:* The frontend changes having to do with the `PlotTestPresenter` should be disregarded. These changes were transferred over by one of Reece's push to this PR and do not affect the functionality of SNAPRed.

[...]

PR #234, opened 2024-02-14
dMin correct smoothing output

darshdinger:
## Description of work

Previous PR #222 had set spectra to flat lines at zero if there were no peaks in them.

@mguthriem said this was not the desired behavior.

This should fix the smoothing as required.

## Explanation of work

Removed the `if-elif` inside `SmoothDataExcludingPeaks` that skipped smoothing if no peaks were in the peak list.  Doing this skipped the important step of smoothing, using the smoothing-spline (which is not the same as a cubic spline and includes some not-node-fitting behavior).

## Link to EWM item
Created during Hackathon 2024/02/19

### DIFF
--- a/src/snapred/backend/recipe/algorithm/SmoothDataExcludingPeaksAlgo.py
+++ b/src/snapred/backend/recipe/algorithm/SmoothDataExcludingPeaksAlgo.py
@@ -144,10 +144,10 @@ class SmoothDataExcludingPeaksAlgo(PythonAlgorithm):
                 smoothing_results = tck(xMidpoints, extrapolate=False)
                 outputWorkspace.setY(index, smoothing_results)
 
-                self.mantidSnapper.WashDishes(
-                    "Cleaning up weight workspace...",
-                    Workspace=self.weightWorkspaceName,
-                )
+            self.mantidSnapper.WashDishes(
+                "Cleaning up weight workspace...",
+                Workspace=self.weightWorkspaceName,
+            )
             self.mantidSnapper.executeQueue()
             self.setProperty("OutputWorkspace", outputWorkspace)
 
--- a/src/snapred/backend/recipe/algorithm/SmoothDataExcludingPeaksAlgo.py
+++ b/src/snapred/backend/recipe/algorithm/SmoothDataExcludingPeaksAlgo.py
@@ -121,7 +121,11 @@ class SmoothDataExcludingPeaksAlgo(PythonAlgorithm):
 
         for index in range(numSpec):
             if np.all(weightWorkspace.readY(index) == 1):
-                smoothing_results = np.zeros_like(weightWorkspace.readY(index))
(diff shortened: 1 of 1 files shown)


src/snapred/backend/recipe/algorithm/SmoothDataExcludingPeaksAlgo.py
```diff
@@ -121,7 +121,12 @@ def PyExec(self):
 
         for index in range(numSpec):
             if np.all(weightWorkspace.readY(index) == 1):
-                smoothing_results = np.zeros_like(weightWorkspace.readY(index))
+                self.mantidSnapper.CloneWorkspace(
+                    "Cloning workspace for no peaks...",
+                    InputWorkspace=self.inputWorkspaceName,
+                    outputWorkspace=self.outputWorkspaceName,
+                )
+                self.mantidSnapper.executeQueue()
```
rboston628, line 129: Did Malcolm say to leave the workspace smoothed and not show a line at zero?

``` python
for index in range(numSpec):
    if not np.all(weightWorkspace.readY(index) == 1):
        x = inputWorkspace.readX(index)
        etc.
```

What you're doing here will re-create the output workspace from the input workspace.  That will end up overwriting results from previous `index` values.

rboston628: Or 
``` python
for index in range(numSpec):
    if np.any(weightWorkspace.readY(index) != 1):
        x = inputWorkspace.readX(index)
        etc.
```
or
``` python
for index in range(numSpec):
    if np.all(weightWorkspace.readY(index) == 1):
        continue
    else:
        x = inputWorkspace.readX(index)
        etc.
```

rboston628: Actually, removing the `if` entirely will do the same thing.

The values `outputWorkspace.readY(index)` are already equal to `inputWorkspace.readY(index)` because the output workspace was cloned from the input in L100.  If every value of `weightWorkspace.readY(index)` is equal to `1`, then at L136 `y = y[weightY != 0]` is  setting `y` equal to itself.  You then make a spline with all of the y-values, then get fitted values at the nodes of the spline, which means it return the exact same values.

So there isn't a need to branch if all of the weights are `1`.

However, it will save unecessary overhead calculations, as if all weights are `1` then you don't need to make a spline just to get back the node values.

rboston628:
Ran it as is, it is no longer zeroing out spectra without peaks.  You can see in the graph below, only the final two groups have peaks and the smoothing operation is being applied.  In the others, there are no peaks, so no smoothing is being applied.

image

rboston628 (approved):
As discussed in Hackathon 2024/02/19, this change fixed the issue with the normalization output.

PR #237, opened 2024-02-16
Centralized I/O for special workspaces

ekapadi:
## Description of work
Diffraction-calibration table and mask workspaces are simultaneously saved in a special hdf5 format, using Mantid's `SaveDiffCal` algorithm.  Unfortunately, this algorithm does not offload any of the associated instrument information.  When reloading these workspaces using `LoadDiffCal`, generally an instrument-donor workspace must be provided.  This is especially true considering the mutable detector _location_ parameters used by the SNAP instrument -- if these parameters are not restored correctly, the detector positions will be invalid

Since the primary responsibility of `GroceryService` is workspace loading and caching, it makes sense that it could also conveniently keep track of instrument-donor workspaces.  Previously this was accomplished using a `fromPrev` `GroceryListBuilder` method, however, in certain cases requiring the loading of a `MaskWorkspace`, no suitable `fromPrev` workspace would be available in the grocery list.  Since in many cases, a suitable instrument-donor workspace will already have been loaded by the service, caching the instrument donor is a straightforward solution.

An additional issue associated with loading and saving of these special workspaces is due to the fact that up to three workspaces are combined into one file on the disk.  This is also quite easy to deal with using the `GroceryService` caching mechanism, in combination with _consistently_ applying the `WorkspaceNameGenerator` naming scheme for these workspaces.

The last major issue dealt with by this pull request is removing all workspace _write_ methods from the `GroceryService`, and thereby removing all `GroceryService` methods from the `LocalDataService` "helper" class.  This then allows several required methods from the `LocalDataService` to be utilized within `GroceryService`, without generating circular import issues.

## Explanation of work

  * All mask and table workspace I/O has now been centralized. Special workspace loading now includes automatic caching of suitable instrument-donor workspaces by the `GroceryService`.  These changes remove some of the complexity from the previous treatment of instrument donor in `GroceryService`.  The previous `GroceryListBuilder.source` method is _retained_, in order to allow the instrument-donor caching to be overridden if necessary.  The `GroceryListBuilder.fromPrev` method has been removed.

  * A standardized signature is now used for all workspace "write" methods (now in `LocalDataService`).  This signature is close to what we had already, but gets rid of the conflation of "path" (as base path) and "path" (as completely-specified file path) arguments.  These write methods are then used in the `DataExportService` `exportCalibrationWorkspaces` and `exportNormalizationWorkspaces` methods (new), called by their respective services.

  * All references to `GroceryService` have been moved out of `LocalDataService`:  this minor change allows `LocalDataService` to be _the_ "helper class", and allows `GroceryService` to call `LocalDataService` methods (where necessary) without any circular imports.  Additionally, this allows `GroceryService` to focus on workspace loading and caching, which is appropriate to its name.

  * `LocalDataService.writeCalibrationReductionResult` has been removed as _dead_ code -- mostly due to issues arising during fixups to its unit test.  Not only was this method unused, but its code and unit test were contradictory enough to the current data-file naming scheme that it was almost impossible to ascertain what the method had been intended to accomplish.  If necessary, this method can be added back in and reworked to be consistent with the scheme used in `LocalDataService.writeCalibrationWorkspaces` or `LocalDataService.writeNormalizationWorkspaces`.

  * Lots and lots of spelling corrections.

## To test
  * Extensive modifications to the existing unit tests have been implemented.

### CIS testing
  * These changes are largely SNAPRed internal.  No CIS-testing will be necessary.

## Link to EWM item

[EWM#3856](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=3856>)

### DIFF
--- a/src/snapred/backend/dao/ingredients/GroceryListItem.py
+++ b/src/snapred/backend/dao/ingredients/GroceryListItem.py
@@ -81,13 +81,15 @@ class GroceryListItem(BaseModel):
                     # the Lite grouping scheme reduces native resolution to Lite mode
                     if v.get("useLiteMode"):
                         logger.warning(
-                            "the lite-mode flag must be False for the 'lite' grouping scheme -- this cannot be overridden"
+                            "the lite-mode flag must be False for the 'lite' grouping scheme" +
+                            " -- this cannot be overridden"
                         )                    
                     v["useLiteMode"] = False  # the lite data map only works on native data
                     
                     if v.get("runNumber") is not None:
                         logger.warning(
-                            "the run number must not be specified for 'lite' grouping scheme -- this cannot be overridden"
+                            "the run number must not be specified for 'lite' grouping scheme" +
+                            " -- this cannot be overridden"
                         )                    
                     # the Lite grouping scheme uses the unmodified native instrument
                     v["runNumber"] = cls.RESERVED_NATIVE_RUNID
--- a/src/snapred/backend/dao/ingredients/GroceryListItem.py
+++ b/src/snapred/backend/dao/ingredients/GroceryListItem.py
@@ -80,11 +80,15 @@ class GroceryListItem(BaseModel):
                 if v["groupingScheme"] == "Lite":
                     # the Lite grouping scheme reduces native resolution to Lite mode
                     if v.get("useLiteMode"):
-                        logger.warning("the lite-mode flag must be False for the 'lite' grouping scheme -- this cannot be overridden")                    
+                        logger.warning(
+                            "the lite-mode flag must be False for the 'lite' grouping scheme -- this cannot be overridden"
+                        )                    
                     v["useLiteMode"] = False  # the lite data map only works on native data
                     
                     if v.get("runNumber") is not None:
-                        logger.warning("the run number must not be specified for 'lite' grouping scheme -- this cannot be overridden")                    
+                        logger.warning(
+                            "the run number must not be specified for 'lite' grouping scheme -- this cannot be overridden"
+                        )                    
                     # the Lite grouping scheme uses the unmodified native instrument
                     v["runNumber"] = cls.RESERVED_NATIVE_RUNID
                 if v.get("runNumber") is None:
--- a/src/snapred/backend/dao/ingredients/GroceryListItem.py
+++ b/src/snapred/backend/dao/ingredients/GroceryListItem.py
@@ -1,29 +1,31 @@
-from typing import Literal, Optional, ClassVar
+from typing import ClassVar, Literal, Optional
 
 from pydantic import BaseModel, root_validator
 
-from snapred.meta.Config import Config
 from snapred.backend.log.logger import snapredLogger
+from snapred.meta.Config import Config
 
 logger = snapredLogger.getLogger(__name__)
+
 
 class GroceryListItem(BaseModel):
     """
     Holds necessary information for a single item in grocery list
     """
+
     # Reserved instrument-cache run-number values:
-    RESERVED_NATIVE_RUNID: ClassVar[str] = "000000" # unmodified _native_ instrument:
-                                                    #   from 'SNAP_Definition.xml'
-    RESERVED_LITE_RUNID: ClassVar[str] = "000001"   # unmodified _lite_ instrument  :
-                                                    #   from 'SNAPLite.xml' 
+    RESERVED_NATIVE_RUNID: ClassVar[str] = "000000"  # unmodified _native_ instrument:
+    #   from 'SNAP_Definition.xml'
+    RESERVED_LITE_RUNID: ClassVar[str] = "000001"  # unmodified _lite_ instrument  :
+    #   from 'SNAPLite.xml'
 
     workspaceType: Literal["neutron", "grouping", "diffcal", "diffcal_output", "diffcal_table", "diffcal_mask"]
     useLiteMode: bool  # indicates if data should be reduced to lite mode
-    
-    # optional loader:    
+
+    # optional loader:
     # -- "" tells FetchGroceries to choose the loader
     loader: Literal["", "LoadGroupingDefinition", "LoadNexus", "LoadEventNexus", "LoadNexusProcessed"] = ""
-    
+
     # the correct combinaton of the below must be set -- neutron and grouping require a runNumber,
     #   grouping additionally requires a groupingScheme
     runNumber: Optional[str]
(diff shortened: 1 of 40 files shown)


src/snapred/backend/data/GroceryService.py
```diff
@@ -190,52 +237,110 @@ def getCloneOfWorkspace(self, name: WorkspaceName, copy: WorkspaceName):
...
+    The fetch methods orchestrate finding data files, loading them into workspaces,
+    and preserving a cache to prevent re-loading the same data files.
+    """
+
+    def _fetchInstrumentDonor(self, runId: str, useLiteMode: bool) -> WorkspaceName:
+        self._updateInstrumentCacheFromADS(runId, useLiteMode)
+
+        key = self._key(runId, useLiteMode)
+        if self._loadedInstruments.get(key) is not None:
+            return self._loadedInstruments.get(key)
```
rboston628, line 251: Could you rewrite this function to have only a single return?

src/snapred/backend/data/GroceryService.py
```diff
@@ -190,52 +237,110 @@ def getCloneOfWorkspace(self, name: WorkspaceName, copy: WorkspaceName):
...
+        for paramName in ("arc", "lin"):
+            for index in range(2):
+                addSampleLog = AlgorithmManager.create("AddSampleLog")
+                addSampleLog.setProperty("Workspace", wsName)
+                addSampleLog.setProperty("LogName", "det_" + paramName + str(index + 1))
+                addSampleLog.setProperty("LogText", str(getattr(detectorState, paramName)[index]))
+                addSampleLog.setProperty("LogType", "Number Series")
+                addSampleLog.setProperty("UpdateInstrumentParameters", (logsAdded >= 3))
+                addSampleLog.execute()
+                logsAdded += 1
```
rboston628, line 304: This would end up being called twice inside PixelGroupingParametersCalculationAlgorithm.  Is there any negative consequence to that, and could that block of code be removed from there in favor of this one?

src/snapred/backend/data/GroceryService.py
```diff
@@ -339,8 +444,8 @@ def fetchNeutronDataCached(self, runId: str, useLiteMode: bool, loader: str = ""
         self._loadedRuns[key] += 1
         return data
 
-    def fetchLiteDataMap(self) -> WorkspaceName:
-        item = GroceryListItem.builder().grouping("Lite").build()
+    def fetchLiteDataMap(self, runId: str) -> WorkspaceName:
```
rboston628, line 447: The Lite data map really shouldn't be dependent on run ID, nor on state.  Is there a way to remove this dependency?

src/snapred/backend/data/GroceryService.py
ekapadi: Done.  Note that there are at least two coding conventions right now, **the one used by my one-previous development group was "early return".**  I was actually trained to always only have one return in a method, so that kind-of bugged me.

src/snapred/meta/builder/GroceryListBuilder.py
```diff
@@ -60,19 +61,15 @@ def useLiteMode(self, useLiteMode: bool):
...
             self._hasInstrumentSource = True
-            propuhdy, source = list(kwarg.items())[0]
-            self._tokens["instrumentPropertySource"] = propuhdy
-            self._tokens["instrumentSource"] = source
-        return self
-
-    def fromPrev(self):
-        self._hasInstrumentSource = True
-        self._tokens["instrumentPropertySource"] = "InstrumentDonor"
-        self._tokens["instrumentSource"] = "prev"
```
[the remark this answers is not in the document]
ekapadi: Added a `fromRun` builder method.

src/snapred/backend/dao/ingredients/GroceryListItem.py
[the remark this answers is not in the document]
ekapadi: I'm deferring this to a later refactor.  There are some other issues as well, and I don't think we can get all of them now.

src/snapred/meta/builder/GroceryListBuilder.py
```diff
@@ -18,8 +18,9 @@ def neutron(self, runId: str):
         self._tokens["runNumber"] = runId
         return self
 
-    def grouping(self, groupingScheme: str):
+    def grouping(self, runId: str, groupingScheme: str):
```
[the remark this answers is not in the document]
ekapadi: I managed to "special case" the Lite data map, so it will no longer require a run number.

src/snapred/backend/data/GroceryService.py
```diff
@@ -190,52 +237,110 @@ def getCloneOfWorkspace(self, name: WorkspaceName, copy: WorkspaceName):
...
+
+        key = self._key(runId, useLiteMode)
+        if self._loadedInstruments.get(key) is not None:
+            return self._loadedInstruments.get(key)
+
+        wsName = None
+        self._updateNeutronCacheFromADS(runId, useLiteMode)
+        if self._loadedRuns.get(key) is not None:
+            # If possible, use a cached neutron-data workspace as an instrument donor
+            wsName = self._createRawNeutronWorkspaceName(runId, useLiteMode)
```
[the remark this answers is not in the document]
ekapadi: For "nexus-loaded" workspaces, the instrument location parameters are encoded in the nexus file (which is actually an hdf-5 format file, but with a special schema that it must adhere to).  If we were loading _all_ of our workspaces from nexus, there would be _no_ instrument donor issues at all (it would be completely transparent  :(  ).

src/snapred/backend/data/GroceryService.py
ekapadi: **The "update instrument parameters" code will be removed from `PixelGroupingParametersCalculationAlgorithm`, but that's going to happen in my other story.**  So far, it doesn't appear to throw off the calculation (i.e. all of the "regression" tests are passing).  The key thing with calling it twice is whether or not the location-parameter changes are applied as _relative_ changes or not.  (The fact that the tests continue to pass suggests that they're absolute.)

src/snapred/backend/data/GroceryService.py
ekapadi: The dependency has now been removed, by using some reserved locations in the instrument-donor cache system (and adding a "special case" to the loading of the lite data map).  Note that, as I mentioned on slack, the lite-data map workspace won't have the correct location parameters -- right now, I am assuming that this doesn't matter for its intended usage.  (When it does matter, it's still possible to specify a run number for it, and the corresponding location parameters will be updated in a cached copy.)

rboston628 (approved):
Code looks good.  I ran both diffcal and normalization on this branch and both processes ran without errors.

PR #244, opened 2024-02-21
sort groupings and samples for consistent ui

walshmm:
## Description of work
What it says on the tin

## To test

open and close the test panel a few times and see that the dropdowns are always in the same abc order.

## Link to EWM item

[EWM#4068](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4068)


darshdinger (approved):
Looks good.

PR #247, opened 2024-02-23
ewm4057 round to 5 decimals and remove dupes in peaks

walshmm:
## Description of work

When running the diffcal flow sometimes PDCalibration would throw up because of a column name collision key on peak position.
This PR addresses this issue.

## Explanation of work

In addition to removing the dupes this pr also has to round the peak position to 5 decimal points because of how PDCalibration seems to handle things.

Column with name @0.48568
in this case 0.48568302651230516 and 0.48568302651230505 would be the same.

## To test

Run the Diffraction Calibration Workflow with the following params

```
run=52102, calibrant=NAC.json and Peak Intensity Threshold = 0.001.
```

Observe that on next a name collision occurs, observe of this branch is no longer does.

## Link to EWM item

[EWM#4057](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4057)


src/snapred/backend/recipe/algorithm/DetectorPeakPredictor.py
```diff
@@ -91,10 +99,16 @@ def PyExec(self) -> None:
...
+                if self.purgeDuplicates:
+                    d = round(d, 5)
+
                 singleFocusGroupPeaks.append(
                     DetectorPeak(position=LimitedValue(value=d, minimum=d - widthLeft, maximum=d + widthRight))
                 )
 
+            if self.purgeDuplicates:
+                singleFocusGroupPeaks = list({peak.position.value: peak for peak in singleFocusGroupPeaks}.values())
+
```
rboston628, line 111: why not using a set, instead of a dictionary?

rboston628 (commented):
Code looks good.

src/snapred/backend/recipe/algorithm/DetectorPeakPredictor.py
walshmm: Pydantic objects give an unhashable type exception.  I could introduce a hash method or do a pydantic "freeze" but this felt least like introducing a new moving part.

rboston628 (approved):

src/snapred/backend/recipe/algorithm/DetectorPeakPredictor.py
rboston628: No, the dictionary is fine

src/snapred/backend/recipe/algorithm/PurgeOverlappingPeaksAlgorithm.py
rboston628, line 1: I forgot we had this algorithm.  Shouldn't this be used to remove duplicates instead?

walshmm: it could be but this may remove more peaks than the ones that are causing the collision, as it accounts for range as well.  The current change just removes ones that are in the exact same position.

PR #214, opened 2024-01-24
Ewm3702 saving calibration records

bolotovskyr:
## Description of work

A calibration record file has a list of workspaces that are saved as files in the same directory.  The names of the workspaces and their filenames must meet certain criteria:

- The name should end with a run number followed by a version number, for example
`diffract_consts_046680_v0015` for a workspace and
`diffract_consts_046680_v0015.h5` for the corresponding file
 - Run numbers and version numbers should be zero-padded as shown in the example above
 - Workspace names should be easy to interpret. For example, instead of the current `difc`, use `diffract_consts`

The above rules also apply to the calibration **metric** workspaces.  The metric workspaces are not saved as files, but are listed in the Workspace window of the SNAPRed UI.

The names of calibration version directories should also be zero-padded, e.g. `v_0015` instead of the current `v_15`.

Filename extensions must reflect the corresponding workspace Mantid types.

Calibration and Normalization records must be saved in separate directories:
```
../Calibration/Powder/04bd2c53f6bf6754/diffraction
../Calibration/Powder/04bd2c53f6bf6754/normalization

```
**ATTENTION:**

This PR also fixes a UI bug introduced by [PR#288](https://github.com/neutrons/SNAPRed/pull/228): on the Assessing tab the `CalibrationIndex` drop-down list  is always empty. To fix the bug, a missing call to `CalibrationAssesmentView` `updateRunNumber` was re-instated:
``` python
   def _triggerCalibrationReduction(self, workflowPresenter):
    ....
        self._calibrationAssessmentView.updateRunNumber(self.runNumber)
```
Correspondingly, the implementation of `updateRunNumber` was also re-instated.

**KNOWN ISSUES**

- In `GroceryService` `readDiffCalTable` the instrument information is entered temporarily as:
`instrumentFilename = "tests/resources/inputs/pixel_grouping/SNAPLite_Definition.xml"`
This needs to be replaced with real instrument information
- If the workspace name contains an iteration number, the number is positioned after the run number. It should be positioned before the run number.
- Before saving a workspace the iteration number needs to be removed from the workspace name.

## Explanation of work

[...]

PR #249, opened 2024-02-26
A file-by-file rebase onto the 'SNAPRed / next' branch from the 'ewm3702_saving_calibration_records' branch.

ekapadi:
**I've kept most of the changes from the original branch, except for the following:**

  * Workspace I/O methods are updated to be consistent with the current service layer.

  * `CalibrationRecord` <workspace list> functionality has been modified to no longer use the proposed `WorkspaceInfo` class.  The information contained in this class was _derivable_ from the workspaces themselves, and so using this class didn't really make much sense.

  * `CalibrationRecord` now uses a **workspace dict** keyed by `WorkspaceNameGenerator` workspace types.  At each key there can be a list containing several workspaces.  This is only "loosely" connected to the WNG-types at the moment, but will serve as a placeholder until this can be done more completely.

  * Arguably, since the workspaces in the `CalibrationRecord` are always the same, the only purpose of storing this information in the record would be if for some reason the end users need to see this information.  We need to check with Malcolm about this point.  If this information is unnecessary, the actual list should be moved to `application.yml`, and be by WNG-type.

  * All unit tests are presently passing.

  * All UI and Filesystem functionality of the original story is functioning as expected.

## Description of work

A calibration record file has a list of workspaces that are saved as files in the same directory.  The names of the workspaces and their filenames must meet certain criteria:

- The name should end with a run number followed by a version number, for example
`diffract_consts_046680_v0015` for a workspace and
`diffract_consts_046680_v0015.h5` for the corresponding file
 - Run numbers and version numbers should be zero-padded as shown in the example above
 - Workspace names should be easy to interpret. For example, instead of the current `difc`, use `diffract_consts`

The above rules also apply to the calibration **metric** workspaces.  The metric workspaces are not saved as files, but are listed in the Workspace window of the SNAPRed UI.

The names of calibration version directories should also be zero-padded, e.g. `v_0015` instead of the current `v_15`.

Filename extensions must reflect the corresponding workspace Mantid types.

Calibration and Normalization records must be saved in separate directories:
```
../Calibration/Powder/04bd2c53f6bf6754/diffraction
../Calibration/Powder/04bd2c53f6bf6754/normalization

```
**ATTENTION:**

This PR also fixes a UI bug introduced by [PR#288](https://github.com/neutrons/SNAPRed/pull/228): on the Assessing tab the `CalibrationIndex` drop-down list  is always empty. To fix the bug, a missing call to `CalibrationAssesmentView` `updateRunNumber` was re-instated:
``` python
   def _triggerCalibrationReduction(self, workflowPresenter):
    ....
        self._calibrationAssessmentView.updateRunNumber(self.runNumber)
```
Correspondingly, the implementation of `updateRunNumber` was also re-instated.

**KNOWN ISSUES**

- If the workspace name contains an iteration number, the number is positioned after the run number. It should be positioned before the run number.
- Before saving a workspace the iteration number needs to be removed from the workspace name.
- Because this does _not_ occur, although multiple iterations will execute without any issues,  they cannot be correctly reloaded by the "assessment" panel.
- Correct implementation of this feature probably requires a WNG-name parser.
- **THIS MAY STILL NEED SOME WORK**, in another story...

## Explanation of work

For the purposes of `WorkspaceNameGenerator` (WNG), all relevant name templates have been updated to make sure that the run number and version tokens are positioned as required. Also, token name changes like `difc` -> `diffract_consts` were made.

Zero-padding of run number and version number is accomplished with the help of a new class `ValueFormatter`.
Correspondingly, a one-line change to `NameBuilder` `__getattr__` method:

  ``` python
    def __getattr__(self, key):
        if key not in self.keys:
            raise RuntimeError(f"Key [{key}] not a valid property for given name.")

        def setValue(value):
            
             # FORMAT THE VALUE
             value = ValueFormatter.formatValueByKey(key, value)
            
            self.props[key] = value
            return self
  ```
allows for **automatic** formatting of the corresponding name tokens without any changes to the client code. For example, an existent line of code 
``` python
wng.diffCalInput().runNumber(runId).build()
```
will result in automatic formatting of `runId`.

Filenames, unlike workspace names, might not be built exclusively with WNG: a version suffix (as well as a filename extension) may be added "by hand" immediately before saving the workspace. To format the version number in such cases a call is placed directly to `ValueFormater`, e.g.
``` python
        if version:
            filePath += "_" + ValueFormatter.formatVersion(version)
            filePath += ".h5"
 
```
The same applies to formatting a calibration version directory name, e.g.
``` python
   def _constructCalibrationDataPath(self, runId: str, version: str):
        stateId, _ = self._generateStateId(runId)
        statePath = self._constructCalibrationStatePath(stateId)
        calibrationVersionPath: str = statePath + "v_{}/".format(
            ValueFormatter.formatVersion(version=version, use_v_prefix=False)
        )
        return calibrationVersionPath
```

A new name template was added to WNG to account for metric workspaces that are not assigned a version yet, but have a timestamp instead: `workspace:nameTemplate:timed_metric`.

## To test

Set up a directory structure:
../Calibration/Powder

Set up a directory structure:
../Calibration/Powder

For testing diffraction calibration use the following parameters:
Run Number: 46680
Convergence Threshold: 0.1
Peak Intensity Threshold: 0.0001
Bins Across Peak Width: 10
Sample: Diamond_001.json
Grouping File: Column lite

For testing normalization calibration use the following parameters:
Run Number: 58882
Background Number: 58813
Smoothing Parameter: 0.000001
Sample: Silicon_NIST_640D_001
Grouping File: Column lite

TEST 1

1. From the `Test Panel` `Calibrating` tab run Diffraction Calibration
2. When `Assessing` tab is shown, click Continue
3. On the `Saving` tab fill in the fields and click Continue
4. Check that the saved `CalibrationRecord.json` has the correct list of persistent workspaces.  Check that the corresponding filenames are also correct.

Close and reopen UI

TEST 2
1. From the `Test Panel` `Diffraction Calibration`/`Calibrating` tab run diffraction calibration
2. When `Assessing` tab is shown, verify that version 2 calibration is present in the drop-down list
3. Click Load. Check the Workspaces window to make sure that all expected workspaces have been created

Close and reopen UI

TEST 3
1. From the `Test Panel` `Normalization Calibration`/`Normalization Calibration` tab run normalization calibration.
2. When `Tweak Parameters` tab is shown, click Continue
3.  On the `Saving` tab fill in the fields and click Continue
4. Check that the saved `NormalizationRecord.json` has the correct list of persistent workspaces.  Check that the corresponding filenames are also correct.

## Link to EWM item

[EWM#3702](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3702)

### DIFF
--- a/src/snapred/backend/dao/calibration/CalibrationRecord.py
+++ b/src/snapred/backend/dao/calibration/CalibrationRecord.py
@@ -1,4 +1,4 @@
-from typing import List, Optional
+from typing import Dict, List, Optional
 
 from pydantic import BaseModel
 
@@ -6,7 +6,7 @@ from snapred.backend.dao.calibration.Calibration import Calibration
 from snapred.backend.dao.calibration.FocusGroupMetric import FocusGroupMetric
 from snapred.backend.dao.CrystallographicInfo import CrystallographicInfo
 from snapred.backend.dao.state.PixelGroup import PixelGroup
-from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceName
+from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceName, WorkspaceType
 
 
 class CalibrationRecord(BaseModel):
@@ -15,7 +15,7 @@ class CalibrationRecord(BaseModel):
     runNumber: str
     crystalInfo: CrystallographicInfo
     calibrationFittingIngredients: Calibration
-    pixelGroups: Optional[List[PixelGroup]]  # TODO: really shouldnt be optional, will be when sns data fixed
+    pixelGroups: Optional[List[PixelGroup]]  # TODO: really shouldn't be optional, will be when sns data fixed
     focusGroupCalibrationMetrics: FocusGroupMetric
-    workspaceNames: List[WorkspaceName]
+    workspaces: Dict[WorkspaceType, List[WorkspaceName]]
     version: Optional[int]
--- a/src/snapred/backend/dao/ingredients/GroceryListItem.py
+++ b/src/snapred/backend/dao/ingredients/GroceryListItem.py
@@ -4,6 +4,7 @@ from pydantic import BaseModel, root_validator
 
 from snapred.backend.log.logger import snapredLogger
 from snapred.meta.Config import Config
+from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceNameGenerator as wng
 
 logger = snapredLogger.getLogger(__name__)
 
@@ -26,11 +27,13 @@ class GroceryListItem(BaseModel):
     # -- "" tells FetchGroceries to choose the loader
     loader: Literal["", "LoadGroupingDefinition", "LoadNexus", "LoadEventNexus", "LoadNexusProcessed"] = ""
 
-    # the correct combinaton of the below must be set -- neutron and grouping require a runNumber,
+    # the correct combination of the below must be set -- neutron and grouping require a runNumber,
     #   grouping additionally requires a groupingScheme
     runNumber: Optional[str]
     version: Optional[str]
     groupingScheme: Optional[str]
+
+    unit: Optional[Literal[wng.Units.TOF, wng.Units.DSP]]
 
     # SpecialWorkspace2D-derived workspaces (e.g. grouping or mask workspaces)
     #   require an instrument definition, these next two properties indicate
@@ -67,13 +70,16 @@ class GroceryListItem(BaseModel):
         if v.get("instrumentSource") is not None:
             if v.get("instrumentPropertySource") is None:
                 raise ValueError("if 'instrumentSource' is specified then 'instrumentPropertySource' must be specified")
-
+        if v.get("unit") is not None:
+            unit_ = v.get("unit")
+            if unit_ not in (wng.Units.TOF, wng.Units.DSP):
+                raise ValueError(f"unknown unit '{unit_}' specified")
         match v["workspaceType"]:
             case "neutron":
                 if v.get("runNumber") is None:
                     raise ValueError("Loading neutron data requires a run number")
                 if v.get("groupingScheme") is not None:
-                    v["groupingScheme"] = None
+                    del v["groupingScheme"]
                 if v.get("instrumentPropertySource") is not None:
                     raise ValueError("Loading neutron data should not specify an instrument")
             case "grouping":
@@ -105,19 +111,15 @@ class GroceryListItem(BaseModel):
             case "diffcal_output":
                 if v.get("runNumber") is None:
                     raise ValueError(f"diffraction-calibration {v['workspaceType']} requires a run number")
-                if v.get("isOutput") is False:
-                    raise ValueError(
-                        f"diffraction-calibration {v['workspaceType']} output specification is special-order only"
-                    )
                 if v.get("instrumentPropertySource") is not None:
                     raise ValueError("Loading diffcal-output data should not specify an instrument")
+                if not v.get("useLiteMode"):
+                    v["useLiteMode"] = True  # don't care
             case "diffcal_table" | "diffcal_mask":
                 if v.get("runNumber") is None:
                     raise ValueError(f"diffraction-calibration {v['workspaceType']} requires a run number")
-                if v.get("isOutput") is False:
-                    raise ValueError(
-                        f"diffraction-calibration {v['workspaceType']} output specification is special-order only"
-                    )
+                if not v.get("useLiteMode"):
+                    v["useLiteMode"] = True  # don't care
             case _:
                 raise ValueError(f"unrecognized 'workspaceType': '{v['workspaceType']}'")
         return v
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -4,6 +4,7 @@ import json
 import os
 from errno import ENOENT as NOT_FOUND
 from pathlib import Path
+from copy import deepcopy
 from typing import Any, Dict, List, Tuple
 
 import h5py
@@ -41,8 +42,10 @@ from snapred.backend.recipe.algorithm.SaveGroupingDefinition import SaveGrouping
 from snapred.meta.Config import Config, Resource
 from snapred.meta.decorators.ExceptionHandler import ExceptionHandler
 from snapred.meta.decorators.Singleton import Singleton
-from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceName
-from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceNameGenerator as WNG
+from snapred.meta.mantid.WorkspaceNameGenerator import ValueFormatter as wnvf
+from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceName, WorkspaceType
+from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceNameGenerator as wng
+from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceType as wngt
 from snapred.meta.redantic import (
     write_model_list_pretty,
     write_model_pretty,
@@ -86,6 +89,8 @@ class LocalDataService:
 
         # look for the config file and verify it exists
         self.instrumentConfigPath = Config["instrument.config"]
+        if self.verifyPaths and not Path(self.instrumentConfigPath).exists():
+            raise _createFileNotFoundError(Config["instrument.config"], self.instrumentConfigFilePath)
 
     def readInstrumentConfig(self) -> InstrumentConfig:
         self._determineInstrConfigPaths()
@@ -211,6 +216,10 @@ class LocalDataService:
         return SHA.hex, SHA.decodedKey
 
     def _findMatchingFileList(self, pattern, throws=True) -> List[str]:
+        """
+        Find all files matching a glob pattern.
+        Optional: throws exception if nothing found.
+        """
         fileList: List[str] = []
         for fname in glob.glob(pattern, recursive=True):
             if os.path.isfile(fname):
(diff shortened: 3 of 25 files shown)


ekapadi:
It looks as if there's a glitch in `application.yml`.  I'll take a look at it in the morning...

src/snapred/backend/dao/calibration/CalibrationRecord.py
```diff
@@ -17,5 +17,5 @@ class CalibrationRecord(BaseModel):
     calibrationFittingIngredients: Calibration
     pixelGroups: Optional[List[PixelGroup]]  # TODO: really shouldnt be optional, will be when sns data fixed
     focusGroupCalibrationMetrics: FocusGroupMetric
-    workspaceNames: List[WorkspaceName]
+    workspaces: Dict[str, WorkspaceName]
```
walshmm, line 20: I'm tempted to say this should instead be its own Pydantic Model instead of a dict, the only downside to that I'm seeing is if at somepoint we abstracted the record as a whole to support N sets of metadata and M workspaces being saved.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -540,17 +545,20 @@ def writeCalibrationWorkspaces(self, record: CalibrationRecord):
...
-            self.writeWorkspace(calibrationDataPath, workspaceFilename, workspace)
+        calibrationTable = ""
+        maskWorkspace = ""
+        # Assumes all workspaces are of WNG-type:
+        for wsKey, wsName in record.workspaces.items():
+            match wsKey:
+                case wngt.DIFFCAL_TABLE:
+                    calibrationTable = wsName
+                case wngt.DIFFCAL_MASK:
+                    maskWorkspace = wsName
```
walshmm, line 556: would 
```suggestion
        calibrationTable = record.workspaces.pop(wngt.DIFFCAL_TABLE, "")
        maskWorkspace = record.workspaces.pop(wngt.DIFFCAL_MASK, "")
        outputWorkspace = record.workspaces.pop(wngt.DIFFCAL_OUTPUT, "")
        
        if len(record.workspaces) > 0:
           ValueError... 
        
        if outputWorkspace:
            filename = ...
        ...
```
work better?

src/snapred/backend/service/CalibrationService.py
```diff
@@ -216,39 +218,41 @@ def loadQualityAssessment(self, request: CalibrationLoadAssessmentRequest):
...
             CalibrationMetricsWorkspaceIngredients(calibrationRecord=calibrationRecord)
         )
 
-        # load persistent workspaces
-        for ws_name in calibrationRecord.workspaceNames:
-            self.dataFactoryService.getCalibrationDataWorkspace(
-                calibrationRecord.runNumber, calibrationRecord.version, ws_name
-            )
+        # load persistent data workspaces, assuming all workspaces are of WNG-type
+        for wsKey, wsName in calibrationRecord.workspaces.items():
```
walshmm, line 245: I understand why is has to be here but would like to look at refactoring it once we have a better way.  Better integration with `WorkspaceName` may help groceryClerk infer things.

(I think technically the `self.groceryClerk.name` can be dragged to the top of the loop as  `self.groceryClerk.name(wsKey)`)

src/snapred/backend/dao/calibration/CalibrationRecord.py
ekapadi: **I modified this treatment only slightly, with the argument that it's a temporary work around in any case and we shouldn't spend too much time on it right now.**  Malcolm just asked us to provide diffraction-calibration output workspaces, in TOF and DSP units simultaneously.  With this in mind, at each WNG.WorkspaceType as key, there is now a list of workspace names.

src/snapred/backend/data/LocalDataService.py
ekapadi: Same problem as above -- there's now a possible list of workspace names at each key value.  However, the current solution also _assumes_ that there's only one 'table' + 'mask' combination -- definitely not an ideal approach.  **We really need a WNG name parser.**

src/snapred/backend/service/CalibrationService.py
ekapadi: I moved the key to the top.  Because of Malcolm's recent request, I also had to break out a temporary fixup for the diffraction-calibration output-workspace "unit" treatment.

src/snapred/backend/data/LocalDataService.py
walshmm: Right but the loop on a map and case statements is effectively the same as just accessing each case, defaulting to empty.
Its removes a layer of nesting.
If the argument is that is is purely temporary and we shouldnt dwell on it then I concede.

ekapadi: **I actually ended up using your approach, although I made a copy, so the input record wouldn't be modified.**  :)  Doing stuff like assuming there was only _one_ mask + table combination really bugged me, so now I'm stepping through the lists (and there can now be multiple copies, although I expect that will never happen except for DIFFCAL_OUTPUT.

src/snapred/backend/dao/ingredients/GroceryListItem.py
```diff
@@ -26,12 +27,14 @@ class GroceryListItem(BaseModel):
...
     loader: Literal["", "LoadGroupingDefinition", "LoadNexus", "LoadEventNexus", "LoadNexusProcessed"] = ""
 
-    # the correct combinaton of the below must be set -- neutron and grouping require a runNumber,
+    # the correct combination of the below must be set -- neutron and grouping require a runNumber,
     #   grouping additionally requires a groupingScheme
     runNumber: Optional[str]
     version: Optional[str]
     groupingScheme: Optional[str]
 
+    unit: Optional[str]
```
walshmm, line 36: Should this be `Optional[Literal[wng.Units.TOF, wng.Units.DSP]]`?

walshmm:
Besides my additional comment and the test coverage, this should be fine to rebase and merge.

src/snapred/backend/dao/ingredients/GroceryListItem.py
ekapadi: Good idea, although I wish we were just using `Enum`.  I'll add this change.

walshmm: Regardless, I'm good with moving forward if you are.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -236,7 +240,7 @@ def _findMatchingDirList(self, pattern, throws=True) -> List[str]:
 
     def _constructCalibrationStateRootPath(self, stateId):
         # TODO: Propagate pathlib through codebase
-        return f"{self.instrumentConfig.calibrationDirectory}/Powder/{str(stateId)}/"
+        return f"{self.instrumentConfig.calibrationDirectory}/Powder/{str(stateId)}/diffraction/"
```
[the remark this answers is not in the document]
walshmm: Malcolm had run into this before, I know its slightly tangential

walshmm (approved):

src/snapred/backend/data/LocalDataService.py
```diff
@@ -236,11 +240,11 @@ def _findMatchingDirList(self, pattern, throws=True) -> List[str]:
 
     def _constructCalibrationStateRootPath(self, stateId):
         # TODO: Propagate pathlib through codebase
-        return f"{self.instrumentConfig.calibrationDirectory}/Powder/{str(stateId)}/"
+        return f"{self.instrumentConfig.calibrationDirectory}/{str(stateId)}/diffraction/"
```
walshmm, line 243: this points to `Config["instrument.calibration.home"]` (` /SNS/SNAP/shared/Calibration`)
so wouldnt this be incorrect now?

ekapadi: I just had to fix that so it would work on analysis.  The "Powder" seemed to be added twice.  Can you clear up the confusion?  Is the `InstrumentConfig` json coming from SNS incorrect?

walshmm: it gets set [here](https://github.com/neutrons/SNAPRed/blob/01c5f148e9ac77b86b716a423c11c9575ae003d3/src/snapred/backend/data/LocalDataService.py#L103)

ekapadi: Just let me know what the final path should look like and I'll make sure its working like that.  **I'm also wondering why `Config["instrument.calibration.home"]` isn't referenced directly here.**

walshmm: There is a conflict as to which params should go into the one stored on sns and the config that should be shipped with the software.  The final say on that was delayed till after phase 3 though calibration directory will probably stay in the `application.yml`

ekapadi: OR, is it `Config["instrument.calibration.powder.home"]`?

walshmm: > Just let me know what the final path should look like and I'll make sure its working like that. I'm also wondering why Config["instrument.calibration.home"] isn't referenced directly here?
The short answer is rust, it was something added early.

walshmm:
```suggestion
        return f"{Config['instrument.calibration.powder.home']}/{str(stateId)}/diffraction/"
```

ekapadi:
**I added the applicable additional comments and tests discussion from Robert's original [PR214](https://github.com/neutrons/SNAPRed/pull/214).**

I made the proposed changes to the calibration-file paths,  and also made another "cleanup" pass.  Essentially all of the cleanup work is in `LocalDataService` and `test_LocalDataService`.

**I'm very happy with how all of this is working now in the GUI on analysis, with the exception that I've not fully checked iteration behavior** -- my feeling at this point is that if there are defects associated with that, they should be part of another story.

ekapadi:
I managed to get loading and saving from "iterations" working correctly, up to a point (as mentioned in the previous comment).

walshmm (approved):

PR #250, opened 2024-02-26
Use the GroupingMap :world_map: 

rboston628:
## Description of work

In PR #201 we had implemented a GroupingMap object, which connects the instrument state with all valid grouping definitions for that state.  This will facilitate general grouping files, such as those made by mantid's [`GenerateGroupingPowder`](https://docs.mantidproject.org/nightly/algorithms/GenerateGroupingPowder-v1.html), which had been edited for this purpose in SNAPRed.

In order to take full advantage of the GroupingMap, it has to be propagated into the GroceryService, and into the workflows, so that they will lookup files from the map, rather than constructing them based on a known pattern.

Because the allowed groupings will now depend on the run number through the state ID, some additional functionality was added in the UI to disable the grouping dropdown until a valid run number is entered, and to re-populate the dropdown whenever the run number changes.

## Explanation of work

I left some comments on files to explain anything I think needed clarification.

### Dev testing

1. Open the GUI, run DiffCal
2. Try entering some nonsense run, like `123` and click away or press eneter -- you should see the default grouping map
3. Enter run number 46680 , and click away or press enter -- observe something happens in the background and the drop down populates
5. Select a grouping
6. Select diamond as the sample
7. Try clicking lite mode on and off and observe the grouping changing
8. Adjust peak threshold to 0.0 (otherwise will have insufficient peaks)
9. Run and make sure the ingredients and grouping can load

If the ingredients and grouping can load, then the workflows and grocery service can properly access the grouping definition files from inside the GroupingMap.

1. Open the GUI, run normalization
2. Try entering some nonsense run, like `123` and click away or press eneter -- you should see the default grouping map
3. Enter run number 46680 , and click away or press enter -- observe something happens in the background and the drop down populates
4. Use 46680 as the background
5. Select a grouping
6. Select diamond as the sample
7. Try clicking lite mode on and off and observe the grouping changing
9. Run and make sure the ingredients and grouping can load
10. On the graph screen, try changing dMin, intensity, and grouping and make sure everything loads as expected

### CIS testing

Try the above.

Experiment with different runs.

Make sure the overall UX is intuitive and it makes sense why things are grayed out or  inaccessible.  Make sure no unrecoverable errors occur.

## Link to EWM item

[EWM#3875](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3875)


src/snapred/backend/data/GroceryService.py
rboston628, line 1: Some of the changes marked changes involve simply moving some filename methods to be next to the other filename methods.

src/snapred/backend/recipe/algorithm/CustomGroupWorkspace.py
rboston628, line 1: This change is a minimal expected change for whenever this is refactored.

src/snapred/resources/application.yml
rboston628, line 1: there's no reason this should be limited to lite files, or only to hdfs

src/snapred/ui/presenter/TestPanelPresenter.py
rboston628, line 1: The title "NormalizationCalibration" was going off the bottom edge of the window and looked awful.  The shorter title helps it to fit.

tests/util/groupingMapUtil.py
rboston628, line 1: This path needs to be absolute for an internal check inside GroupingMap

tests/unit/backend/data/test_DataFactoryService.py
rboston628, line 1: The main change was to replace the pattern
``` python
dataExportService.lookupService.someMethodName = mock.Mock()
dataExportService.lookupService.someMethodName.return_value = X
```
with
``` python
dataExportService.lookupService.someMethodName = mock.Mock(return_value=X)
```

tests/unit/backend/dao/test_GroupingMap.py
rboston628, line 1: trying to match the return file name is pretty tough

tests/resources/application.yml
rboston628, line 1: this was superfluous

tests/unit/backend/service/test_ConfigLookupService.py
rboston628, line 1: This is not a new file.  It just used to be an empty file.

tests/unit/backend/data/test_LocalDataService.py
rboston628, line 1: This had a patch of `mantid.api` that caused several imports from mantid, including `mtd` and the workspace types needed for tests, to be mocked out, causing tests to fail.  The patch ended up not being necessary.

rboston628:
This needs to be merged after PR #254 due to some collisions ion behavior for invalid run numbers

rboston628:
By using the default grouping map when no valid state is available from the run number, this should be compatible with PR #254, and is ready for review again

src/snapred/backend/dao/state/GroupingMap.py
```diff
@@ -75,8 +82,9 @@ def validate_GroupingMapFile(cls, v):
             for group in groups[mode].copy():
                 fp = Path(groups[mode][group].definition)
                 # Check if path is relative
-                if str(fp) == fp.name:
+                if not os.path.isabs(fp):
```
walshmm, line 85: is the original check insufficient?  It seems be based on the idea that if it was not a path it should default to the root grouping folder.

src/snapred/backend/data/GroceryService.py
```diff
@@ -134,14 +136,27 @@ def _createNeutronFilename(self, runNumber: str, useLiteMode: bool) -> str:
...
+    def _createGroupingFilename(self, runNumber: str, groupingScheme: str, useLiteMode: bool) -> str:
         if groupingScheme == "Lite":
-            return str(Config["instrument.lite.map.file"])
-        instr = "lite" if useLiteMode else "native"
-        home = "instrument.calibration.powder.grouping.home"
-        pre = "grouping.filename.prefix"
-        ext = "grouping.filename." + instr + ".extension"
-        return f"{Config[home]}/{Config[pre]}{groupingScheme}{Config[ext]}"
+            path = str(Config["instrument.lite.map.file"])
+        elif self._groupingMapCache.get(runNumber):
```
walshmm, line 142: it may be unnecessary to cache this?  We may face similar issues to caching calibration jsons.

src/snapred/backend/dao/state/GroupingMap.py
rboston628: The original check was failing.  If the file is `/SNS/SNAP/path/to/some_grouping_definition.hdf`, then the check only passes if the `focusGroup.name` element is `some_grouping_definition`.  However, it's better to allow the file paths and focus group names be unrelated.  So  for instance having a `focusGroup` with `name="All"` and `definition="SNAPFocGroup_All.hdf`, as in the default list, which will here be expanded to point into the state folder

src/snapred/backend/data/GroceryService.py
rboston628: I'll remove the caching

src/snapred/backend/data/LocalDataService.py
```diff
@@ -816,6 +828,15 @@ def _readGroupingMap(self, stateId: str) -> GroupingMap:
             raise FileNotFoundError(f'required grouping-schema map for state "{stateId}" at "{path}" does not exist')
         return parse_file_as(GroupingMap, path)
 
+    def readGroupingMap(self, runNumber: str):
+        # if the state exists then lookup its grouping map
+        if self.checkCalibrationFileExists(runNumber):
```
walshmm, line 833: There probably shouldnt be a case in which we interact with a state that lacks a grouping map.  It was previously not hooked into initialization but @ekapadi had added it.

src/snapred/backend/recipe/algorithm/ReductionAlgorithm.py
```diff
@@ -56,7 +56,7 @@ def PyExec(self):
         # 8 CreateGroupWorkspace      TODO: Assess performance, use alternative Andrei came up with that is faster
         groupingworkspace = self.mantidSnapper.CustomGroupWorkspace(
             "Creating Group Workspace...",
-            StateConfig=reductionIngredients.reductionState.stateConfig.json(),
+            FocusGroups=reductionIngredients.reductionState.stateConfig.json(),
```
walshmm, line 59: This is missing an accessor to focusgroups I think.

src/snapred/backend/data/LocalDataService.py
rboston628: This check was added today to facilitate it working with PR #254.  It also handles the case of a user entering an errant run number.  In that case it loads the default grouping map.  If a user enters incomplete info, or looks for a state that isn't initialized, this allows them to select a default group and press "Continue", where they are directed to initialize state.

src/snapred/backend/recipe/algorithm/ReductionAlgorithm.py
rboston628: You are correct

src/snapred/backend/dao/state/GroupingMap.py
walshmm: I see, sounds reasonable.

src/snapred/backend/data/LocalDataService.py
walshmm: How is the default grouping map used in this case?

rboston628: We want the user to be able to enter a run number for a state that is not initialized and click continue, so Darsh's state initialization view comes up.  In order for them to click continue, they have to select a grouping, so the dropdown has to be populated.

When the user first opens the Test Panel window, so far nothing has been initialized in the backend.
image
The grouping map is tied to the state, so the only map we can put in at first is the default one.

The user then has to enter a run number. 
image

As soon as the user clicks away form the run number, but before clicking continue, the grouping map needs to change to reflect the available groupings in this state.

It does this by looking up the state ID for the run number, then looking up the grouping map for the state ID, and putting the appropriate list of groupings in the dropdown.

So far nothing has been initialized in the backend.

If the user at this point enters either an invalid run number, or a run number for which there is no state, then the only list of groupings we can use is the default one.

If the user enters a run number with no state, then selects a default grouping, then presses continue, only then will the user be guided to initialize the state.

The simplest thing here is to make this function to return a specific grouping map if it exists, and to otherwise return the default grouping map.

walshmm: Ah I see, populate defaults.

walshmm (dismissed):

walshmm (approved):

PR #241, opened 2024-02-20
Shade in detected peaks in normalization graphs

rboston628:
## Description of work

When performing normalization, we perform a smoothing operation with depends upon the vanadium peaks.  Not all peaks are included in the calculation, only those with d-values between to bounds, and only those with peak intensities above a threshold.  As these are adjusted, different peaks will appear in the calculation.

Which peaks are included is largely opaque to users.  This aims to make it more visibly obvious when peaks are excluded from the calculation.  It does so by displaying dmin on the graph, and by highlighting the peaks that are included and smoothed over.

## Explanation of work

Getting the peaks *into* the view required rearranging some of the normalization service, in particular the call to smoothing.

Because the graphs are normalized by bin width, the x,y values must be extracted from the workspace using mantid's `get_spectrum()` method, which can return the bin-normalized y value, which can then be used to plot the fill.  Maybe a future mantid update should include its own fill-between method.

The smoothing algorithm had the option to accept a list of peaks instead of peak ingredients, but this functionality had apparently gone untested and was not working.  Also for reasons unknown to mere mortals, the change to the graphing inside the view caused GenericRecipe to refuse to parse floats into strings to mantid algorithms.

Several methods in the normalization service were labeled "fromString" but did not have defined paths.  They probably should have defined paths in the future, but to avoid ambiguity I removed the "fromStrings" that I noticed were superfluous.

### Dev testing

Make sure all unit tests pass.

In workbench, run the script I will post in the comments below.  Verify the control over the peaks by adjusting dMin and peakIntensityThreshold.

Next run the GUI, and check everything as described in CIS testing below.

### CIS testing

For better visualization, first test using
- run = 58882
- background = 58813
- sample = silicon

Initial run should look something like this.  The red line indicates dMin, the blue-filled curves are those included in the peak list.
image

Try adjusting dMin, and watch the red line move up and fewer peaks be included.
image

Once you are satisfied, try with vanadium, such as run 58810 and vanadium sample.  It will look something like the following:
image

Try switching groups from Column --> All (will calculate) --> Column.  The change back to Column should be instantaneous, if no other parameters were touched.

**PLEASE VERIFY THAT SAVING WORKS**
The correct vanadium workspace should be saved with the focused/smoothed workspaces when the normalization is saved.

## Link to EWM item

[EWM#4071](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4071)

PR #251, opened 2024-02-27
Peak Peek :eye: :mountain:

rboston628:
## Description of work

In PR #226 , #231, #241  had added improvements to the graphs shown in normalization process.

The views were deemed very useful in diagnosing issues, including diagnosing issues within the separate diffraction calibration workflow.  @mguthriem  would actually first load data inside the normalization workflow in order to inspect the peaks then go back to the diffcal workflow to implement the values found.

This adds the graph view directly into the diffcal process, which makes it easier to visualize what is going on in peak selection process.

## Explanation of work

1. I renamed some views and workflows to be more appropriate.
2. Inside `CalibrationService`, I split up prepping ingredients and fetching groceries into their own processes that can be called through the interface controller
3. Inside `CalibrationService`, I added an additional method to focus spectra, which can be called through the interface controller
4. There is a new view which displays the graphs from the diffraction calibration process
5. I added needed methods to the diffcal workflow to handle the additional graph view
6. I rebased the normalization workflow on a WorkflowImplementer
7. I rebased several normalization views on the BackendView

#### DiffCal Workflow
1. Open, input run number 466, click away.  Only default groupings should be available.  Input 46680 and click away, the grouping menu will populate (may have to check terminal to see this).
2. Select diamond and column.  Gaussian should already be selected.  Press continue.
3. **The next screen should be a peak viewer.**
4. Ensure that the grouping list is already populated with non-default groupings and that Column is pre-selected
5. Adjust parameters and press recalculate until there are four peaks.  Make sure dMin, dMax move the red bars, make sure peak intensity threshold going lower selects more peaks.
6. Change the grouping from Column to Bank.
7. Press to continue, make sure diffcal process runs and ends on the assessment screen.  This is probably good enough.

#### Normalization Workflow
1. Open, input run number 588, click away.  Only default groupings should be available.  Input 58882, click away, grouping menu will populate (may have to check the terminal to see it's calling a function).
2. For background use 58813, select **silicon** and column, press continue.
3.  **The next screen should be a peak viewer.**
4. Ensure that "Column" is pre-selected and that the non-default groupings are populated.
5. Experiment with the slider, make sure its smoothing behavior makes sense and is as before.  As it goes to zero the smoothing should match all the tiny peaks but smooth over the selected (shaded) peaks.  As it goes to one, the tiny peaks should also be smoothed over.
6. Press continue and make sure the Save screen loads.

### CIS testing

Check the above, but using better data, and make sure final calculations complete.

Check that the layout is as you would like it, things in correct places.

I know the peak view is showing before pixel calibration.  Having the peak view after pixel calibration but before group calibration would require a lot more work, as it requires splitting and refactoring the diffcal recipe.

## Link to EWM item

[EWM#4161](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4161)
[EWM#4229](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4229)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # project specific files
 src/snapred/_version.py
 conda.recipe/
+src/snapred/resources/dev.yml
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/docs/source/api/snapred.ui.view.rst
+++ b/docs/source/api/snapred.ui.view.rst
@@ -12,10 +12,10 @@ snapred.ui.view.BackendRequestView module
    :undoc-members:
    :show-inheritance:
 
-snapred.ui.view.CalibrationReductionRequestView module
+snapred.ui.view.DiffCalRequestView module
 ------------------------------------------------------
 
-.. automodule:: snapred.ui.view.CalibrationReductionRequestView
+.. automodule:: snapred.ui.view.DiffCalRequestView
    :members:
    :undoc-members:
    :show-inheritance:
--- a/docs/source/api/snapred.ui.workflow.rst
+++ b/docs/source/api/snapred.ui.workflow.rst
@@ -4,10 +4,10 @@ snapred.ui.workflow package
 Submodules
 ----------
 
-snapred.ui.workflow.DiffractionCalibrationCreationWorkflow module
+snapred.ui.workflow.DiffCalWorkflow module
 -----------------------------------------------------------------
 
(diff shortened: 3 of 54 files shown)


darshdinger (approved):
Looks good! Man diffraction is so fast now.

PR #256, opened 2024-03-04
Masked pixel-grouping parameters calculation.

ekapadi:
## Description of work
`PixelGroupingParameters` provide the angular expectation of the d-spacing diffraction resolution values, accumulated separately over each pixel group.  Due to their dependence on the physical positions of the pixels, the calculated values of these parameters change when masking is taken into account.
## Explanation of work

  * Account for detector masking during the calculation of the `PixelGroupingParameters`.  As all of the Mantid algorithms in the calculation chain take the detector masking flags into account, it is sufficient to provide a grouping workspace at the start of the calculation, with both the correct instrument-location parameters and the desired detector masking flags.  Due to grouping workspace caching by `GroceryService`, a temporary copy of input grouping workspace is made prior to modifying its detector masking flags.  (Internally Mantid uses copy-on write [COW] pointers, and so the expected overhead associated with this workspace copy should be minimal.)

  * A new algorithm `MaskDetectorFlags` is implemented, allowing the application of mask-workspace values to the detector masking flags of a target workspace.  It will be noted that the analogous Mantid algorithm `MaskDetectors` additionally clears output workspace spectra, which is undesirable behavior in the present application.  (For example when applied to a grouping workspace, `MaskDetectors` will move all masked pixels to group "0".)

  * The generation and treatment of golden data by the `test_PixelGroupingParametersCalculationAlgorithm` was updated and made to be more consistent.  This series of tests had been implemented primarily as _regression_ tests, and this does seem to be a correct approach to test this type of calculation.  However, it did not make sense to allow the golden data to differ between local test runs, and test runs on the analysis nodes.  Further, golden-data files are now identified by a timestamp section in the naming scheme.  In addition, the ability to generate new golden data when required is now built into this set of tests.

### Dev testing
All changes are covered by both existing and new unit tests.  Additional changes have been made to `test_PixelGroupingParametersCalculationAlgorithm` so that the golden data is now treated consistently -- using different golden data between the local test runs and the analysis test runs is no longer allowed.

### CIS testing
A new CIS-test script is provided: `tests/cis_tests/pixel_grouping_parameters_script.py`.

## Link to EWM item

[EWM#3644](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=3644)


src/snapred/backend/dao/request/FarmFreshIngredients.py
```diff
@@ -13,10 +13,11 @@ class FarmFreshIngredients(BaseModel):
     from these, the Sous Chef can make everything
     """
 
-    # NOTE this class ia a REQUEST object, despite its name
+    # NOTE this class is a REQUEST object, despite its name
     # Do NOT use inside ingredients for algorithms
 
     runNumber: str
+    version: Optional[str]
```
darshdinger, line 20: Curious to why version is made into an optional attribute? Is this due to run numbers not having initialized state (and therefore no associated versions) when they are first reduced?

src/snapred/backend/recipe/algorithm/MaskDetectorFlags.py
```diff
@@ -0,0 +1,76 @@
...
+        if sourceInstrument.getNumberDetectors(True) != destInstrument.getNumberDetectors(True):
+            errors["MaskWorkspace"] = "Mask and output workspaces must have the same number of (non-monitor) pixels"
+
+        if self.maskWS.getNumberHistograms() != destInstrument.getNumberDetectors(True):
+            errors["MaskWorkspace"] = "Mask workspace must have one spectrum per (non-monitor) pixel"
+
+        return errors
+
+    def unbagGroceries(self):
+        pass
```
darshdinger, line 60: Remove?

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
```diff
@@ -58,11 +66,11 @@ def chopIngredients(self, ingredients: PixelGroupingIngredients) -> None:
         self.delTheta = ingredients.instrumentState.delTh
         return
 
-    def unbagGroceries(self, ingredients: PixelGroupingIngredients):
+    def unbagGroceries(self, ingredients: PixelGroupingIngredients):  # noqa: ARG002
```
darshdinger, line 69: Do we need to pass in `PixelGroupingIngredients` if they aren't being used?

src/snapred/backend/service/SousChef.py
```diff
@@ -30,6 +30,7 @@
 from snapred.meta.Config import Config
 from snapred.meta.decorators.FromString import FromString
 from snapred.meta.decorators.Singleton import Singleton
+from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceName
```
darshdinger, line 33: Remove?

src/snapred/backend/data/GroceryService.py
```diff
@@ -270,20 +274,24 @@ def _fetchInstrumentDonor(self, runId: str, useLiteMode: bool) -> WorkspaceName:
...
-                # (Reserved IDs will use the unmodified instrument.)
-                if runId != GroceryListItem.RESERVED_NATIVE_RUNID and runId != GroceryListItem.RESERVED_LITE_RUNID:
-                    self._updateInstrumentParameters(runId, wsName)
+                # (Reserved run-numbers will use the unmodified instrument.)
+                if (
+                    runNumber != GroceryListItem.RESERVED_NATIVE_RUNNUMBER
+                    and runNumber != GroceryListItem.RESERVED_LITE_RUNNUMBER
+                ):
+                    detectorState: DetectorState = self._getDetectorState(runNumber)
+                    self.updateInstrumentParameters(wsName, detectorState)
```
walshmm, line 283: Just checking 90% of this file is just a changing `runid` -> `runNumber`, plus the code here

src/snapred/backend/recipe/algorithm/CalibrationMetricExtractionAlgorithm.py
```diff
@@ -80,7 +80,9 @@ def PyExec(self):
 
             sigmaAverage = np.average(sigmas)
             strainAverage = np.average(strains)
-            twoThetaAverage = pixelGroupingParameters[index].twoTheta
+
+            # Convert to degrees from radians (?)
+            twoThetaAverage = pixelGroupingParameters[index].twoTheta * (180.0 / np.pi)
```
walshmm, line 85: Why is this required?

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
walshmm: ++

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
```diff
@@ -72,15 +80,32 @@ def PyExec(self):
         self.chopIngredients(ingredients)
         self.unbagGroceries(ingredients)
 
-        # create a dummy grouped-by-detector workspace from the grouping workspace
+        # create a copy of the grouping workspace with the correct detector mask flags
+        tmpGroupingWSName = mtd.unique_hidden_name()
+        self.mantidSnapper.CloneWorkspace(
```
walshmm, line 85: [MakeDirtyDish](https://github.com/neutrons/SNAPRed/blob/578236ba553de2b2fd6f2be1529c15b50bb51922/src/snapred/backend/recipe/algorithm/MakeDirtyDish.py)?

src/snapred/backend/recipe/algorithm/ReductionAlgorithm.py
```diff
@@ -182,8 +183,16 @@ def PyExec(self):
...
+        # Warning: <pixel grouping parameters>.dResolution will be None for fully-masked groups
+        #   "NaN" is interpreted by 'Mantid::RebinRagged' to indicate an unspecified value
+        dMin = {
+            pgp.groupID: pgp.dResolution.minimum if pgp.dResolution else np.nan
+            for pgp in reductionIngredients.pixelGroupingParameters
+        }
+        dMax = {
+            pgp.groupID: pgp.dResolution.maximum if pgp.dResolution else np.nan
+            for pgp in reductionIngredients.pixelGroupingParameters
+        }
```
walshmm, line 195: Is this related to these [similar lines](https://github.com/neutrons/SNAPRed/pull/256/files#diff-a0c5b6461a255c223c08e37521c6b7f014236def75719ef186c4e90fa46e5a21R52-R59)?

if possible should we have some dmin getter method that handles these cases and centralizes this domain knowledge?

src/snapred/backend/dao/request/FarmFreshIngredients.py
ekapadi: **Prior to _acceptance_ by a user, all of the calibration results have no version.**  Personally, I would have preferred giving them a version as the latest result, and then just dropping the results if they are rejected.  However, this "no version" treatment is the way that Robert wanted to do it.  (Note that actually giving them a version, would allow the WNG and filename generation to be a bit more straightforward.  However, at this point, I think this is a minor issue and should wait for refactoring.)

src/snapred/backend/recipe/algorithm/MaskDetectorFlags.py
ekapadi: We seem to be moving gradually towards an ABC implementation of `PythonAlgorithm`.  For this reason, I didn't remove anything, and would suggest we leave it in.

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
ekapadi: My response here would be the same as previous.  An ABC for `PythonAlgorithm` would have `unbagGroceries` with an `ingredients` argument (, although I suppose it could be optional, I personally wouldn't lean towards designing it that way (, as it requires "knowing too much" in advance about the design)).

src/snapred/backend/service/SousChef.py
ekapadi: removed

src/snapred/backend/recipe/algorithm/CalibrationMetricExtractionAlgorithm.py
ekapadi: **The initial test code for this section had its input in _degrees_, which indicated that `CalibrationMetric` is "expecting" to be initialized that way.**  (`PixelGroupingParameters` is calculated in radians.)  I personally would prefer if the conversion to degrees happened in the GUI panel code, but that probably should wait for refactoring.  In general, _radians_ would be considered to be the "mathematical unit".

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
ekapadi: Unfortunately, `MakeDirtyDish` only makes a clone if `CIS_mode == True`.  I don't consider cloning an "input parameter" that will be modified to be optional.  (Obviously this can be argued both ways, but my years of frustrating experience maintaining code that did _not_ follow this practice has lead me to this assertion.)

src/snapred/backend/recipe/algorithm/MaskDetectorFlags.py
darshdinger: Okay, gotcha.

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
walshmm: This is probably the fault of some inconsistent design, most `unbagGroceries` methods in algos actually only have the self argument, the main purpose of the method is to assign workspaces to class members.
`chopIngredients` would take the ingredients and load them into class members.

There is the case of `PixelGroupingParametersCalculationAlgorithm` breaking this pattern, and it should really be refactored to follow suit, pushing the `self.loadNeededLogs(self.groupingWorkspaceName, ingredients)` call out of `unbagGroceries` into `PyExec` after both `chopIngredients` and `unbagGroceries` have executed.

walshmm: I do lean in favor of ABC impl, and having an explicit pass is informative even if it doesnt execute anything.

src/snapred/backend/recipe/algorithm/CalibrationMetricExtractionAlgorithm.py
walshmm: Gotcha, I would like to create a meta type at some point to help handle unit dependent values.

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
walshmm: Oh, I guess that is right.  For some reason I was under the impression that it was just a book keeping wrapper.

src/snapred/backend/recipe/algorithm/ReductionAlgorithm.py
walshmm: If we could at least mark these with todo's for when we do figure it out we dont miss them.  Code is best when you dont have to think about these things!

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
rboston628: It is inconsistent, I take the blame for it.  This method  calls the to load the logs, which needs the ingredients.  If that's no longer happening, then ingredients can be removed from this method's arguments.

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
rboston628: MakeDirtyDish is probably mostly obsolete at this point.  It's goal was to help Malcolm diagnose individual algorithms by logging intermediate workspaces.  We probably don't need most of them anymore.

src/snapred/backend/recipe/algorithm/ReductionAlgorithm.py
rboston628: I think PR #263 treats both code blocks in the desired way.

walshmm: I think the main difference may be the nan handling

walshmm (approved):

src/snapred/backend/recipe/algorithm/ReductionAlgorithm.py
rboston628: I think both this and the other can be moved into the `dMin()`, `dMax()`, and `dBin()` methods in `PixelGroup`

PR #258, opened 2024-03-04
Change Some Names

rboston628:
## Description of work

Several of the views and workflows had names which (in my humble opinion) were sort of confusing.  This is a minimal update to change the names of the files.

Also, there was a slight error inside the normalization workflow not caught in PR #250 , which is fixed here.  At the tweaking stage, the grouping list was being populated with the default groupings.

## Explanation of work

All view names within the DiffCal workflow now beign with `DiffCal`, then have some description that follows:
- `DiffCalRequestView`
- `DiffCalAssessmentView`
- `DiffCalSaveView`

All view names within the Normalization workflow now begin with `Normalization`, then have some description that follows:
- `NormalizationRequestView`
- `NormalizationTweakPeaksView`
- `NormalizationSaveView`

I changed these names everywhere I could find them in the code.

`BackendRequestView` now has a default value in the `_labeledField` method, which will create a `QLineEdit` with a label.

### Dev testing

Run through the diffcal process using 46680, make sure you can save.

Run through the normalization process using 46680.  Double-check the available groupings on the tweaking stage.  Make sure you can save.

### CIS testing

This is an internal issue and does not need CIS testing.

## Link to EWM item
None.

### DIFF
--- a/src/snapred/resources/dev.yml
+++ b/src/snapred/resources/dev.yml
@@ -2,4 +2,4 @@ instrument:
   name: SNAP
   home: /SNS/SNAP/
   calibration:
-    home: ${instrument.home}shared/Calibration_dynamic
+    home: ${instrument.home}shared/Calibration
--- a/docs/source/api/snapred.ui.view.rst
+++ b/docs/source/api/snapred.ui.view.rst
@@ -12,10 +12,10 @@ snapred.ui.view.BackendRequestView module
    :undoc-members:
    :show-inheritance:
 
-snapred.ui.view.CalibrationReductionRequestView module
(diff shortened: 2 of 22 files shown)


src/snapred/resources/dev.yml
rboston628, line 1: This should be ignored by the gitignore.  Why is it here???

walshmm:
I'd probably reccomend waiting on other PR's before merging this one

rboston628:
My hope was this would make PR #251 easier to merge, if I merged this first

darshdinger (approved):
Looks good!

PR #261, opened 2024-03-07
this line was just defaulting to 0, replaced with line from neutrons …

walshmm:
…pyproject template repo
https://github.com/neutrons/python_project_template/blob/main/conda.recipe/meta.yaml

Check to see that builds pass and in the conda build step "Build conda libraray"  it no longer produces `snapred-0-py310.tar.bz2` but `snapred-{INSERT VERSION HERE}-py310.tar.bz2`


rboston628 (approved):
Verified version number appears in conda build step
image

PR #262, opened 2024-03-07
Fix diffcal output workspace name

dlcaballero16:
## Description of work

This is to fix the issue of the output workspace name in diffcal being wrong. The name was "tof_runnumber_diffoc" but the units were in dSpacing. The output will now have 2 outputs, one in dSpacing and one in TOF units, both with their names having the correct units.
## Explanation of work

I modified GroupDiffractionCalibration.py to make sure the names were being set correctly and added support for another output workspace.

### Dev testing

In the GUI, just run the first step of diffraction calibration. When that is done, look back at the main SNAPRed window and look for two workspaces with the names "tof_runnumber_diffoc" and "dsp_runnumber_diffoc". Click on the workspaces and verify the units match with the name.

### CIS testing

See Dev testing.

## Link to EWM item

[EWM#4282](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4282)


walshmm:
image
image
Can confirm these both exist

walshmm:
I can also confirm that after one iteration the units are still correct.

walshmm:
image
output workspaces in save
image
Recorded workspaces.

walshmm (approved):

PR #263, opened 2024-03-07
Rebase to ReductionAlgorithm and AlignAndFocusAlgorithm

rboston628:
## Description of work

The code has moved a lot since these were first made.

The goal here is to remove all load statements, have all workspaces be inputs, and to structure things as simply as possible using the rebased functionality in `PixelGroup`.

This is not a final state, just closer to what it will look like after a full rebase.

## Explanation of work

Removed all load statements inside these algorithms.

The `CustomGroupWorkspace`  did not seem to do anything but load a bunch of grouping workspaces then group them together.  Its output was a `GroupedWorkspace`, but `ReductionAlgorithm` was treating it as though it were a `GroupingWorkspace`.  After this rebase it is mostly a SNAPRed wrapper on the mantid algorithm to group workspaces.

`ReductionAlgorithm` no longer loads data, and more perspicuously performs identical operations on the data and the vanadium.  Also only has a single `executeQueue()` call.

`AlignAndFocusAlgorithm` just calls `AlignAndFocusPowder`, but after `chopIngredients()` extracts the correct dmin, dmax, dbin.

### Dev testing

Make sure the unit tests work.  I made new unit tests with the test utils that should run the algorithm with synthetic data and make sure it runs, then one that makes sure it fails for bad input data.

Make sure `ReductionAlgorithm` and `AlignAndFocusAlgorithm` have the same list of inputs.

### CIS testing

this is an internal issue and does not require CIS testing

## Link to EWM item
None


src/snapred/backend/recipe/algorithm/ReductionAlgorithm.py
```diff
@@ -9,214 +9,129 @@
...
-
-        # compress data
-        # data = self.compressEvents(InputWorkspace=data, OutputWorkspace='event_compressed_data')
-
-        # sum chunks if files are large
-        # TODO: Implement New Strip Peaks that allows for multiple FWHM, one per group,
-        # for now just grab the first one to get it to run
-        peakPositions = ",".join(
-            str(s) for s in reductionIngredients.reductionState.stateConfig.normalizationCalibrant.peaks
+        self.mantidSnapper.CloneWorkspace(
```
walshmm, line 55: Are these clones necessary if we are depending on GroceryService to keep our raw data safe?

src/snapred/backend/recipe/algorithm/ReductionAlgorithm.py
```diff
@@ -9,214 +9,129 @@
...
+        # divide the data by the vanadium
+        # NOTE this must be done before ragged rebinning, since Divide cannot handle unequal axes
+        self.mantidSnapper.Divide(
+            "Dividing data by vanadium...",
+            LHSWorkspace=self.outputWorkspace,
+            RHSWorkspace=self.vanadiumWorkspace,
+            OutputWorkspace=self.outputWorkspace,
+        )
+        # ragged-rebin the data
+        self.mantidSnapper.RebinRagged(
```
walshmm, line 121: So this has been converted to operate on only one group at a time?

rboston628: For the time being.  I couldn't really tell what it was using before, but it seemed to be using the `GroupedWorkspace` of `GroupingWorkspace`s return by the `CustomGroupWorkspace` algorithm inside of mantid algorithms that require a `GroupingWorkspace`, and that didn't seem valid.

Assuming that was valid, then returning it to its former condition this should be as simple as passing it a `GroupedWorkspace` or `GroupingWorkspace`s inside of its own `GroupWorkspace` property.  If not, then when we refactor we can change this to take a `GroupedWorkspace` and iterate over all the workspaces in the group.

src/snapred/backend/recipe/algorithm/ReductionAlgorithm.py
rboston628: @ekapadi had mentioned before that we should distinguish Inputs from InOuts.  It isn't necessary for the reasons you said, but for the sake of clarity this algorithm doesn't make any changes to its input workspace.  I could rewrite the input workspace as an InOut property

walshmm: I would prefer they be InOuts then

src/snapred/backend/recipe/algorithm/ReductionAlgorithm.py
walshmm: Its probably better this way since we have moved away from GroupedWorkspaces

src/snapred/backend/recipe/algorithm/ReductionAlgorithm.py
rboston628: Done

src/snapred/backend/service/ReductionService.py
```diff
@@ -15,13 +15,20 @@ class ReductionService(Service):
...
     @staticmethod
     def name():
         return "reduction"
 
+    @FromString
+    def fakeMethod(self):
+        # NOTE this is not a real method
+        # it's here to be used in the registered paths above, for the moment
+        # when possible this should be deleted
+        return {}
```
walshmm, line 30: `NotImplementedError`?

rboston628: Done

src/snapred/backend/recipe/algorithm/ReductionAlgorithm.py
```diff
@@ -1,231 +1,126 @@
...
 
 #######################################################
 # ATTENTION: Could be replaced by alignAndFocusPowder #
-# please confirm that attenuation correction before  #
+# please confirm that attenuation correction before   #
 # and after is equivalent                             #
 #######################################################
 class ReductionAlgorithm(PythonAlgorithm):
+    """
+    NOTE: this was originally written to work with a list of grouping workspaces,
```
walshmm, line 18: So one catch with this is we may need to split the algo if it only applied to one group.  The reason it was for a set of groups was there were "cached" steps reused for each group.

walshmm (approved):

PR #264, opened 2024-03-08
Adjust graph view row heights

rboston628:
## Description of work

When Malcolm was testng in workbench, the view looked like this:
image (7)

We would like the canvas row to stretch and everything else remain the same height.

## Explanation of work

It seems the wrong row was given a row stretch factor.  It is now on the correct row for the canvas, and its stretch factor is much larger.

### CIS testing

``` python
```

## Link to EWM item


mguthriem:
beautiful!

mguthriem (approved):
much better use of space in UI panel.

PR #268, opened 2024-03-12
forward a lambda to let workflows reset their implementors

walshmm:
## Description of work

Merge this after PR#267 otherwise you wont be able to test this.

This resets the full state of a workflow upon completion, thus preventing the errant missing metrics bug. 
This bug is typically cause by an old workspace name being stored in the expected outputs after a workflow has been reset in some form.

## To test

Run diffraction calibration to completion (`46680`, Diamond, Col)
Run it again but this time try to iterate.
It should iterate without error.

## Link to EWM item


dlcaballero16 (approved):
I was able to run and iterate with no errors.

PR #275, opened 2024-03-14
fix calib/normalization version forwarding

walshmm:
## Description of work

Version saving had not been fixed in the rework to saving calibs and normalization, this pr fixes it instead.

## To test

In your local Calibration Folder
Run a calibration/normalization (49525, diamond, col)
Go all the way to save
Set version to 999
Save
Observe new entry and folder for version 999
repeat
Set version to 1
Save
Observe that version 1 had been overwritten.
Receive warning stating that you have overwritten version 1.


darshdinger (commented):
So cod cov is complaining so that will need to be addressed but otherwise, I will attempt to review this PR following the steps outlined.

Run a calibration/normalization (49525, diamond, col):
image

Go all the way to save & set version to 999 & save:
image

Repeat and set version to 1 & Observe:
image

How these files are persisted to disk:
image


Looks good!

PR #276, opened 2024-03-18
Fix sous chef to call the grouping map when needed

rboston628:
## Description of work

Using a combination of SousChef and GroceryService allows SNAPRed functionality to also be easily run from scripts.

With some changes to how SousChef gets the focus group paths, in order for SousChef to always have valid focus group definitions, it is necessary for it to check the definitions it has, then load them from a grouping map otherwise.

This is not just about scripts.  It makes sure the focus group object passed to the SousChef is always a valid grouping file.

### Dev testing
Open the CIS test script new_diffcal_script inside workbench.

Check the new unit tests, and verify they pass.

You may have to change the name of the grouping, maybe from "Column (Lite)" to "Column" or "Column (Default Lite)" depending on your setup.

Run, and make sure it reaches the first `assert False` stop statement.  If it did, then the ingredients were successfully constructed.

### CIS testing
Make sure you can run the new_diffcal_script inside workbench up to the first `assert False` stop.

## Link to EWM item

### DIFF
--- a/src/snapred/backend/service/SousChef.py
+++ b/src/snapred/backend/service/SousChef.py
@@ -68,7 +68,7 @@ class SousChef(Service):
         return self.dataFactoryService.getCalibrantSample(calibrantSamplePath)
 
     def prepFocusGroup(self, ingredients: FarmFreshIngredients) -> FocusGroup:
-        if os.path.isfile(ingredients.focusGroup.definition):
+        if self.dataFactoryService.fileExists(ingredients.focusGroup.definition):
             return ingredients.focusGroup
         else:
             groupingMap = self.dataFactoryService.getGroupingMap(ingredients.runNumber)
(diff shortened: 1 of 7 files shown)


src/snapred/backend/service/SousChef.py
```diff
@@ -66,36 +67,44 @@ def prepRunConfig(self, runNumber: str) -> RunConfig:
     def prepCalibrantSample(self, calibrantSamplePath: str) -> CalibrantSamples:
         return self.dataFactoryService.getCalibrantSample(calibrantSamplePath)
 
-    def prepPixelGroup(self, ingredients: FarmFreshIngredients):
+    def prepFocusGroup(self, ingredients: FarmFreshIngredients) -> FocusGroup:
+        if os.path.isfile(ingredients.focusGroup.definition):
```
walshmm, line 71: This check should probs be abstracted to the data layer, otherwise this is a service that touches disk directly.

walshmm (approved):

PR #270, opened 2024-03-12
Fix run 57474 failing due to default value of Chi2

dlcaballero16:
## Description of work

Added MaxChiSq param/constant to PDCalibration in GroupDiffractionCalibration
## Explanation of work

We are now just giving the user a warning if there's not enough peaks with a Chi2 value under 100. From Malcolm:

> For the algorithm to complete, there needs to be at least 2 peaks available and usable. To be usable, they need to have a value Chi2 < 100. So, in the case that any pixel group does not meet the condition that at least 2 peaks have chi2<100, the error should be something like "Insufficient number of well fitted peaks (chi2 < 100). Try to adjust parameters in Tweak Peak Peek tab"

### Dev testing

On analysis, run diffcal with run number 57474, convergence thresh = 0.1, peak thresh = 0, bins = 7, sample = Diamond, grouping = column. CIS mode can be true or false in application.yml. On the Tweak Peaks tab, set dMin to 0.85, dMax to 100, and threshold to 0. Recalculate and hit continue. You should see a pop warning with 3 entries that have the new warning message. Hit OK and make sure you can continue to the save step.

### CIS testing

See Dev testing.

## Link to EWM item

[EWM#4351](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4351)

PR #280, opened 2024-03-21
Fit peaks before diffcal to verify peak lists

rboston628:
## Description of work

Recent work in PR #270 would throw a detailed warning if PDCalibration encountered too few peaks.

It is good for that error to be there.  However, it would be better to also inform users before getting to that point.

This will fit the peaks prior to starting the diffcal process, so the users can see:
- which peaks will be included in their calculation
- the fit that will be matched to their peaks
- which selected peaks are failing to fit (highlighted in red)

This allows users to immediately see the fits to the data, and also to know when bad peaks have been selected.

## Explanation of work

Firstly, added an endpoint for `FitMultiplePeaks` inside `CalibrationService`.  This runs instantly, so no loss of performance from performing this prior to calibration.

`FitMultiplePeaks` will be called anytime the user changes any parameters and presses the Recalculate button.

It is necessary to have at least 2 well-fit peaks for PDCalibration to function.  The user will be stopped before continuing if they have fewer than two well-fit peaks, as determined by $\chi^2$ of the fits.

It is recommended ([or at least was, at some point](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4072)) to have at least 4 well-fit peaks for calibration.  If the user tries to calibrate with fewer, they will be warned to use 4.  The user can either adjust peaks, or ignore the warning and continue anyway.

### Dev testing

This required using signals.  Please try changing and pressing everything inside the workflow to make sure no segfaults are generated.

- Open the diffcal workflow in GUI, input run 46880 (or your favorite).
- Select Diamond and **Bank**, then continue
- Tweak Peak Peek should open.  Examine the graphs.  The left side should have two blue peaks, and if you zoom in on them you should see black lines indicating a nice fit to the peaks.  The right side should have only red peaks.
- Press Continue.  You should see an error and not be able to continue.
- Adjust dMin = 0.4, dMax = 1.0, threshold = 0.0, recalculate.
- Now many more peaks should be being fit.  Examine the peaks on the left side to make sure the fit is good.  Make sure *nothing* is being fit on the right: all red, no blue, no black fit peaks.
- Press Continue.  You should see an error and not be able to continue.
- Adjust dMin = 0.4, dMax = 3, threshold = 0.05, **grouping = All**, recalculate.
- You should see a single graph with two good peaks.
- Press continue.  You should be warned that it is recommended to use 4 peaks.  Press No,
- Adjust dMin = 1.0, threshold = 0.0.  This is just to make sure you can make ajdustments in this stage.  There should still be fewer than four good peaks selected.
- Press continue.  Same warning.  Press yes.
- The diffcal process should now begin.  Make sure it finishes, then inspect the outputs to make sure they are as expected.

### CIS testing

Follow the above but with run 58882.  Make sure only good peaks are being selected when in blue and fits are working properly when all advice is followed.

## Link to EWM item

[EWM#4405](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4405)

### DIFF
--- a/src/snapred/backend/service/SousChef.py
+++ b/src/snapred/backend/service/SousChef.py
@@ -1,4 +1,5 @@
 import json
+import os.path
 import time
 from datetime import date
 from functools import lru_cache
@@ -23,6 +24,7 @@ from snapred.backend.data.DataFactoryService import DataFactoryService
 from snapred.backend.data.GroceryService import GroceryService
 from snapred.backend.recipe.GenericRecipe import (
     DetectorPeakPredictorRecipe,
+    PurgeOverlappingPeaksRecipe,
 )
 from snapred.backend.recipe.PixelGroupingParametersCalculationRecipe import PixelGroupingParametersCalculationRecipe
 from snapred.backend.service.CrystallographicInfoService import CrystallographicInfoService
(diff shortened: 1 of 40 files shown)


src/snapred/backend/service/SousChef.py
```diff
@@ -121,20 +122,27 @@ def prepPeakIngredients(self, ingredients: FarmFreshIngredients) -> PeakIngredie
             peakIntensityThreshold=ingredients.peakIntensityThreshold,
         )
 
-    def prepDetectorPeaks(self, ingredients: FarmFreshIngredients) -> List[GroupPeakList]:
+    def prepDetectorPeaks(self, ingredients: FarmFreshIngredients, purge_peaks=True) -> List[GroupPeakList]:
```
dlcaballero16, line 125: just a nitpick, but can you update purge_peaks to be camel case?

rboston628: Can do

dlcaballero16 (approved):
Tested with suggested steps and everything functioned as expected.

dlcaballero16 (approved):

PR #283, opened 2024-03-22
It's not the test panel anymore

rboston628:
## Description of work

@backmari  just asked me why the GUI window is called the Test Panel.

It is now called the Calibration Panel.

## Explanation of work

I changed the name.

I changed the error when loading the calibration panel.  I think this message is more helpful, as it says how an expert user could fix the issue, and who to contact if they need help.

### Dev testing

Open it, launch the calibration panel, make sure it is called Calibration Panel.

Try running something for good measure, but nothing should depend on the window title.

To verify the new error message occurs, edit your dev.yml to point somewhere that doesn't exist then try launching.

### CIS testing

Not needed.

**NOTE**: changing the name is just to avoid confusing users during the beta testing.  Changing the name does **NOT** signal this is the final form of the GUI.

## Link to EWM item

### DIFF
--- a/src/snapred/ui/main.py
+++ b/src/snapred/ui/main.py
@@ -79,11 +79,12 @@ class SNAPRedGUI(QMainWindow):
             traceback.print_exception(e)
             from qtpy.QtWidgets import QMessageBox
 
+            msg = "Sorry!  Error encountered while opening Calibration Panel.\n"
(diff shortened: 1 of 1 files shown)


dlcaballero16 (approved):
I verified the name is updated and no issues running diffcal.

PR #286, opened 2024-03-26
Allow Peak Function Selection in Tweak Peak Peek

rboston628:
## Description of work

While testing, Malcolm noticed a defect when running diffcal with the following:
- run number = 49525
- grouping = Column (but irrelevant)
- peak function = PseudoVoigt
- dMin = 0.4 (mostly irrelevant)
- dMax = 100 (mostly irrelevant)
- peak threshold = 0.05 (mostly irrelevant)

On the initial tweak peak peek, some peaks showed as poorly fitting, but also many showed up as blue:
Pasted Graphic 1

However, when Malcolm selected to continue, PDCalibration ran and failed to fit **any** peaks for **any** of the groups.

The reason for failing to fit **all** peaks is a failure of the PseudoVoigt fitting function, which has some problem inside mantid's `FitPeaks` algorithm.

However, the failure for SNAPRed is that it was not properly passing the fit function parameter to `FitPeaks` during the tweak peak peek.  This function was **only** being passed during calibration, which represented a difference in the two processes.   This is why the peak peek would show good fits but PDCalibration would find bad fits.

This PR will
1. properly propagate the user's peak function selection to `FitMultiplePeaks`
2. allow the user to change the peak function during tweak peak peek.

**ALSO** will fix the issue raised in testing meeting, so that the peak intensity threshold is propagated form first screen to second.

## Explanation of work

Added 

### Dev testing

Wait until Malcolm leaves an approving comment.

### CIS testing

Open in CIS mode.  Try doing exactly what you did to generate this error, including using PseudoVoigt.  It might be easier to use "**All**" grouping to have fewer tables to open.

**ALSO** enter whatever peak intensity threshold you like, but not 0.05.

Verify on the Tweak Peak tab, that **every single** peak is red.

**ALSO** verify that the peak intensity threshold is whatever you had put on the first screen,

Change peak function to Gaussian.

Verify at least **some** of the peaks are not red.

Run diffcal.

Open the _PDCal_diag_X workspace groups, which should contain the `_fitparam` tables.  Compare the peak centers of this, to the peak centers in the `fit_peaks_diag_49525_2_fitted_params_X` table.  Verify peak centers and chi-sq are approximately the same.

## Link to EWM item

[EWM#4538](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4538)

### DIFF
--- a/docs/source/api/snapred.ui.view.rst
+++ b/docs/source/api/snapred.ui.view.rst
@@ -52,7 +52,7 @@ snapred.ui.view.DiffCalTweakPeakView module
 .. automodule:: snapred.ui.view.DiffCalTweakPeakView
    :members:
    :undoc-members:
-   :exclude-members: signalRunNumberUpdate, signalValueChanged, signalUpdateRecalculationButton
+   :exclude-members: signalRunNumberUpdate, signalValueChanged, signalUpdateRecalculationButton, signalPeakThresholdUpdate
    :show-inheritance:
 
 snapred.ui.view.FitMultiplePeaksView module
--- a/src/snapred/backend/dao/request/FitMultiplePeaksRequest.py
+++ b/src/snapred/backend/dao/request/FitMultiplePeaksRequest.py
@@ -3,6 +3,8 @@ from typing import List
 from pydantic import BaseModel
 
 from snapred.backend.dao.GroupPeakList import GroupPeakList
+from snapred.meta.Config import Config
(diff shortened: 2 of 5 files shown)


mguthriem:
Works as expected

darshdinger (approved):

PR #287, opened 2024-03-26
forgot to add the new version to the calibration entry

walshmm:
## Description of work

Forgot to update the entry's value of version. 

## To test

Run Calibration and/or normalization and check their index, see that the correct version is populated.
Calibration: 49525
Normalization: 58813, 58810

## Link to EWM item


dlcaballero16 (approved):
Tested both calibration and normalization, I verified that both records now have the version.

PR #290, opened 2024-03-26
Allow user to change FWHM multipliers

rboston628:
### Dev testing

Run GUI with 46680 and **Bank**.

On the tweak peak view, change the FWHM Right multiplier to 8.  Recalculate.  Make sure the peak is now selected blue.

Because this is using signals, change all kinds of things and recalculate.  Change groupings multiple times, change dMin/dMax multiple times, change peak threshold, etc., to make sure nothing accidentally broke.

When done, click to continue.  You will get a **warning**, but just continue anyway.

When done, save.  

After saving, go to your local folder (because you're running this in your local dev copy of Calibration) and open `CalibrationRecord.json`, and scroll down to "calibrationFittingIngredients":"instrumentState":"fwhmMultiplierLimit", and confirm these values are the ones you set on the tweak peak peek (should be 2.0 and 8.0).

Open `CalibrationParameters.json` and scroll down to "instrumentState":"fwhmMultiplierLimit" and scroll down and verify same values.

### CIS testing

Find the runs which generated this problem.

Adjust the FWHM until you get good peak fits, and verify that the diffcal process executes with good, tightly-fitting peaks.

## Link to EWM item

[EWM#4556](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4556)

### DIFF
--- a/src/snapred/backend/dao/request/CalibrationAssessmentRequest.py
+++ b/src/snapred/backend/dao/request/CalibrationAssessmentRequest.py
@@ -2,7 +2,7 @@ from typing import Dict, List
 
 from pydantic import BaseModel
 
-from snapred.backend.dao.Limit import Limit
+from snapred.backend.dao.Limit import Pair
 from snapred.backend.dao.RunConfig import RunConfig
 from snapred.backend.dao.state.FocusGroup import FocusGroup
 from snapred.meta.Config import Config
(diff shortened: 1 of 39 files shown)


src/snapred/backend/dao/request/CalibrationAssessmentRequest.py
```diff
@@ -30,3 +31,7 @@ class CalibrationAssessmentRequest(BaseModel):
     nBinsAcrossPeakWidth: int = Config["calibration.diffraction.nBinsAcrossPeakWidth"]
     peakIntensityThreshold: float = Config["calibration.diffraction.peakIntensityThreshold"]
     peakType: ALLOWED_PEAK_TYPES = "Gaussian"
+    fwhmMultiplierLimit: Limit[float] = Limit(
```
walshmm, line 34: Something we may want to nip in the bud with this pr is treating the multipliers as a limit, could we change its type to:
```
class Pair(GenericModel, Generic[T]):
    value1: T
    value2: T
    
    def __init__(self, minimum: T = None, maximum: T = None, **kwargs):
        if minimum is not None and maximum is not None:
            super().__init__(value1=minimum, value2=maximum)
        else:
            super().__init__(**kwargs)
            
    @property
    def minimum(self) -> T:
        return self.value1
    
    @property
    def maximum(self) -> T:
        return self.value2
```

as an intermediate solution?

walshmm:
In the state it is currently, it works as intended.
image

walshmm:
However ❌

On Iterate it does not apply the FWHM supplied by the user initially
image

rboston628:
Pre-populate on iterate should be working

walshmm:
image
Can confirm all previously mentioned parts are working.
Its not backwards compatible with whats already on disk but that probably doesnt matter as we havent released yet.

walshmm (approved):

PR #291, opened 2024-03-26
toss an exception stating which group contains a nan metric

walshmm:
## Description of work

This will toss a stopping exception in the Calibration workflow if it detects nans in the metrics.

This is typically indicative of a fit failing, and thus the user should retweak params till fits succeed.

## Explanation of work

It just scans for nans in the end result of metric and catalogs in an error which groups are flagrant

## To test

Probably easiest to do out of your own Calibration folder.
Even easier if you wait till FWHM gets exposed on the front end.

(if having trouble reproducing use the calibration from the mainline Calibration_dynamic) 

(make sure in your calibration's CalibrationParameters.json that FWHM is 2 on both multipliers)
Try to calibrate with:
49525
Convergence: 0.5
Peak Intensity: 0
Bin: 7
Diamond
Column
Lorentzian

Observe it failing

start again with your calibration's CalibrationParameters.json FWHM set to 3 each

Observe it passing.

## Link to EWM item


rboston628:
# Test Log - Take 1

I have moved the indicated directory `/SNS/SNAP/shared/Calibration_dynamic/Powder/ffefaa93ccb23678/diffraction/v_0002` into my local Calibration directory in the corresponding area.
image

I launch SNAPRed to point at my local Calibration direction.
image

After approximately 4 days it finally opens and I go to run as indicated in instructions.
image

Pressing to continue, I am asked to initialize my state.  It asks me this every single time.
image

I select yes, and it creates a v_0003 folder.  Excellent!  Except not really.
image

I delete the v_0003 folder and continue.  I cannot continue, because my state isn't initialized, and it leads me through the whole thing again.

I'm not sure if this is a feature or a bug.  Going to try re-making my Calibration folder.

rboston628:
# Test Log - Take 2

I completely deleted that state folder and am launching the app.  After several days of waiting, it finally loads.  I enter all the right values as before.

I am asked to initialize state, I initialize and the state folder pops up with a v_0001.  I move the indicated v_0002 from `/SNS/SNAP/shared` into this folder.  I **also** edit the calibration parameter JSON so that it refers to `fwhmMultipliers` instead of `fwhmMultipliersLimit`.

I get some nice peaks.  Can confirm FWHM is 2.0 for left and right.  I continue.
image

There's the error I wanted!
image

I change left/right FWHM to 3.0 using the handy, and handsomely designed, feature to do this.  The dev who installed this feature probably slaved for hours to make it, and deserves a raise.  I press continue.

It passes
image

rboston628:
# Test Log -- Take 3

I am using run `58882`.  I load it, and I can see two prominent bad peaks.
image

I click to continue, and receive this much more specific message.
image

However, the error message is errantly saying all peaks are bad, as opposed to only those two bad peaks.

I adjut the dMin, dMax, threshold to exclude the bad peaks, using this extremely handy and functional system thast some very intelligent and handsome software engineer must have added in.
image

I press continue, I'm warned there are only three peaks, I continue anyway, and diffcal runs.

Checking the calibration workspaces, there are no NaNs.

Checking the logs, the string "NaN" was not logged.

# Remaining Issue

Fix the output to only print the bad peaks, not all peaks.

rboston628:
# Test Log - Take 4

I run with `49525` (which has returned from its break) and see the following.
image

There are some nearly hidden, tiny bad peaks that a user might miss.  I click to continue and receive this informative warning
image

Sure enough, those are the bad peaks
image

I adjust FWHM right to 3, and now they all fit
image

I press continue, and it actually points out to tiny bad peaks I didn't notice because it's so small.  This would have otherwise thrown an error only at the very end of the entire diffcal process, so it's nice I'm being told now.
image

I'm not sure if the `[]` should be there for group 0, but I guess it isn't really a stopping issue.

I reduce dMin to 0.5 to skip these bad peaks, and press continue.  I now receive this error message
image

This is a failure that occurred after PDCalibration.  It did not create calibration workspaces, but there are the fitPeaksWS fitparams tables.  For reasons I still do not understand, these have only two peaks in them, even though I passed four peaks.  In each group, both peaks are bad fits.
image

I go back to the peak tweak view and adjust FWHM left to 3.  This causes a large peak to fail so I move dMax down to 1.1, and press to continue.

The process finishes with warnings
image

But it finished, and I am able to save.
image

rboston628 (approved):

PR #298, opened 2024-04-03
 Peak filtering update:

walshmm:
## Description of work

This pr shuffles around how we filter peaks to provide a more intuitive user experience.
The noise to background ratio fitting was causing random non-obvious failures and mostly accounted for an edge case in a near failure state of a run where you needed to use it to prevent fitting of random peaks in noise.  It has been turned off till we can reevalute it.

The intensity threshold has been reworked a bit as below.

The main issue is as you changed the drange it would also "find" new peaks to try and fit.
This is because the relative intensity threshold changed as the drange changed.
With the intensity threshold being calculated based on the whole of the spectra, it will be uniform and a change in drange wont randomly cascade in new peaks.

## Explanation of work

* disable background to noise filter - To be reevaluated
* filter on peak intensity relative to whole spectra
* filter peak intensity post purge overlaps

As needed, use
  ``` python
    x
  ```
to paste code blocks.
-->

## To test

I was using the following in the tweak peaks to illustrate the fix:
(convergence thresh of .1, bins 7)
Before:
image
image

After:
image
image
image

## Link to EWM item


dlcaballero16 (approved):
Was able to test and see the same results.

PR #300, opened 2024-04-04
add better exception handling for when recalc fails

walshmm:
## Description of work

Exception handling for recalc and other signals do not take into account a failure state.
This is a temp solution to combat that.
Also, toss an exception if no peaks are left after purging, to prevent recalc from getting into a weird state where it trys to delete workspaces that dont exist?

## Explanation of 

I wrapped a bunch of code in try/excepts that really should have been spinning up threads with fail and cleanup conditions.

## To test

Convergence: .1
Bins: 7
image

These params on recalc should net an exception, but it should now fail gracefully instead of below:
image

and you should be able to recalc again.


dlcaballero16 (approved):
Ran and was able to recalculate when getting the error.

PR #301, opened 2024-04-05
messed something up in the merge such that this didnt trigger

walshmm:
## Description of work
 same as https://github.com/neutrons/SNAPRed/pull/300
I just messed up the merge, so the user friendly error didnt appear.

Note that all peaks have to be purged to trigger this, just set dmin and dmax to the same value.
Otherwise you get the expected case of only some groups appearing as only some groups have peaks.


darshdinger (approved):

mguthriem:
error message looks good now.

PR #307, opened 2024-04-22
add the isLite flag to calib/norm records

walshmm:
## Description of work

This adds the isLite flag to calibration and normalization records

## To test

if you would like to regression test the ui, try the following for calib:
```
Run Number: 46581
Convergence Threshold: 0.1
Peak Intensity Threshold: 0.0001
Bins Across Peak Width: 10
Sample: Any
Grouping File: Any
Peak Function: Any

dmin: 1.1
```

for normalization any will work really but heres one I copy pasted from another pr:
```
   run = 58813
background = 58813
sample = silicon
```

## Link to EWM item

[EWM#4770](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4770)

### DIFF
--- a/src/snapred/backend/dao/calibration/CalibrationRecord.py
+++ b/src/snapred/backend/dao/calibration/CalibrationRecord.py
@@ -25,6 +25,7 @@ class CalibrationRecord(BaseModel):
 
     runNumber: str
     crystalInfo: CrystallographicInfo
+    isLite: bool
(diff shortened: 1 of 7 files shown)


darshdinger (approved):
Confirmed with ui regression test on analysis!

PR #308, opened 2024-04-22
integration-tests from CIS-test scripts

ekapadi:
## Description of work

The overall purpose of this enabler is to explore the idea of running CIS-test scripts as integration tests.  The prototype solution calls a CIS-test script inside of a wrapper function, which then is itself run as a test by the "pytest" system.

As an example: "tests/cis_tests/new_diffcal_script.py" was used as a reference.  A new script with required additions was created as "tests/cis_tests/diffcal_integration_script.py", which still can be run as a cis-test script.  However, it can also be run by the test framework as called from "tests/integration/diffcal_test.py". In test mode, the results from this script will be compared to golden data.  To facilitate the possibility of moving any golden data to a repository, this type of integration test is intended to compare _metric_ values, and not actual workspace results.

## Explanation of work

The integration-test wrapper function, treated specifically as an integration test, generated the following set of requirements:

1) **The ability to mount a subset of the SNS contents at a local directory location, on a possibly-isolated processing node**:

  * Integration-test input data can include the same data normally used by end users.  The ability to run integration tests from any node (which may not have access to the SNS-filesystem), requires remounting a subset of the SNS contents at a local directory.  Associated changes were required in SNAPRed's Config system, including the definition of the "IPTS.root" and "IPTS.default" entries.

  * For GUI application, it is now sufficient to define "IPTS.root" as required -- everything else is taken care of within SNAPRed's __main__ function.

  * For running isolated scripts, or for running the integration tests themselves, an `IPTS_override` _context_ _manager_ is provided.  Using this manager in its default mode (e.g. `with IPTS_override() as IPTS_root:`) simply _applies_ the "IPTS.root" entry from the `Config`, in a safe manner.  (Under the hood: note that this in turn safely modifies the "datasearch.directories" property in Mantid's configuration system.)

  * An additional required change was allowing the paths specified in the <crystal data> JSON files to be either _absolute_ (as at present) _or_ to be relative to `Config[....samples.home]`.  Primarily this allows the subset of the SNS contents to be mounted _anywhere_ on the target node, without requiring any of the included metadata files to be modified.  (Assuming that the paths in the "samples.home" JSON have been adjusted to be relative paths.)

2) **The ability to easily initialize a state from scratch as part of a test sequence**:

  * An `state_root_override` context manager is provided.  Due to the complexity of SNAPRed's calibration filesystem, this context manager does _not_ modify the state-root location (except as per "IPTS.root" above).  However, it does create a new state-root directory, and initialize its contents in a safe manner -- this directory is _optionally_ deleted at the completion of the test.  Note that if this directory already exists, an exception is thrown.

3) **The ability to safely override any entry in SNAPRed's `Config` system.**

  * A `Config_override` context manager is provided.  This context manager is suitable for most purposes, but note that it does _not_ deal with "multi-level substitution" config entries (i.e. entries where the key lookup and replacement descends _into_ the entry itself, possibly with multiple required substitutions).  Since this type of entry generally refers to other (generally unsubstituted, or at most single-level substituted) config entries, it is felt that any required configuration override can be accomplished by correctly modifying a few of these referred-to entries.  (In case this is all a bit confusing: the recommendation is to _use_ this context manager -- it will almost certainly work for what most developers will need.)

  * `Config` now automatically populates a `module.root` key at initialization.  This allows the "test" version of the "application.yml" to be more easily used for integration tests -- bypassing the previous requirement for a "by hand" override in `conftest.py`.

  * During debugging of associated unit tests, it was found necessary to centralize the "clear_loggers" fixture, promote it to "scope='session'" from _module_ scope, and move its definition to `conftest.py`.  This should clear up a lot of logger difficulties within the test framework.  [This change modifies _many_ tests, however.  Sorry about that!]

4) **New test _markers_ for integration tests (in `pyproject.toml`)**:

  * An "integration" test marker.  By default, tests with this marker are _not_ run by pytest (also controlled in "pyproject.toml").  To run the integration tests use `env=integration_test pytest [-k <part of the test name>] -m integration`.

  * A "golden_data(...)" test marker.  This marker allows the specification of the golden-data base path, a short name to abbreviate the current test, and the _data_ to select from possibly multiple golden-data files.  (See `goldenData` fixture discussion below.)

5) **Decorators to allow test fixtures and associated support functions to be called transparently in non-test scripts**:

  * Applying the `test_only` or `script_only` decorators any function can be substituted with the empty function when a test _is_ / _is_ _not_ running, respectively. (Definition: "empty function": call it with any parameters -> it doesn't do anything but returns `None`.)

  * Using the `not_a_test` decorator, a function (or class) can be marked so that pytest will not mistakenly try to run it as a test case.

6) **A new `goldenData` test fixture, and an associated `goldenDataPath` test fixture.**  (These fixtures can be used with tests that are not part of `unittest.testcase`.)  Generally the `goldenData` fixture is intended to be used as part of a call to the `assertMatchToGoldenData` function.  (However, to see how these fixtures might be used separately, either look at `assertMatchToGoldenData`, or look at the associated unit tests for examples.)

7) **A new `assertMatchToGoldenData` function** -- this function does nothing at all when not run in a test environment.

  * In a test environment, this function uses the same arguments as `pytest.approx`, with an additional `goldenData` argument.

  * The specifics of the golden data file to be used will have been marked using the `golden_data` test marker.

  * If the golden data file exists, it will be loaded and compared to the generated test data.  In general, any nested composite object can be passed to `assertMatchToGoldenData` -- it will be "flattened" prior to comparison -- attribute names, and dictionary keys must match exactly; non-float primitives must match exactly; floating-point values must approximately match, as for `pytest.approx`.

  * If the golden data file doesn't exist, it will be created and then saved at the end of the test.

  * `assertMatchToGoldenData` can be called multiple times within a test -- the specific objects to be compared are keyed by the call number, within the golden-data dict employed -- it is sufficient to just pass the `goldenData` fixture, which does all of this automatically.

### Dev testing

  * **Test the "clear_loggers" fixture**: In `your_favorite_test.py` somewhere put an `assert False`, so that a single test will fail.  On running `pytest -k your_favorite_test` note that the loggers do not spam the output on shutdown.  Then comment out the `pytest.fixture`-line on "clear_loggers" in "conftest.py" -- run the test and now see that the loggers spam the output at shutdown.

  * **Run the "diffcal" integration test**: Create a copy of the "/SNS" filesystem containing at _least_ the supporting files for "tests/cis_tests/diffcal_integration_script.py" (It's only one runnumber, with a fairly small input-data file + associated sample files & configuration files.)  Make sure to include the crystal-sample files (i.e. CIF files) in "samples.home", and edit the JSON files there to use _relative_ paths within that same directory.  Place this filesystem copy whereever you want on your local system.  Edit "integration_test.yml" to include that location at the "IPTS.root" key (note that "IPTS.root" is "/SNS" when not modified).  Run `env=integration_test pytest -m integration`.  One fine point here: make sure that the state-root directory doesn't already exist: the test will create it, initialize the state, and remove it at exit.

  * **Run the SNAPRed GUI from a local mirror of the "/SNS" system**:  Edit your "application.yml" or "dev.yml" to change "IPTS.root" as required (leave "IPTS.default" alone, as it's required for the new capability.) Make sure to include the crystal-sample files (i.e. CIF files) in "samples.home", and edit the JSON files there to use _relative_ paths within that same directory. Run SNAPRed, either using `python -m snapred`, or launch it from `workbench` -- it should just work.

### CIS testing
The last test-item above (i.e. "Run the SNAPRed GUI from a local mirror..." )  should also be run by the CIS.

## Link to EWM item

[EWM#4399](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=4399)


ekapadi:
One glitch in the "GUI tests" still needs to be fixed.

pyproject.toml
```diff
@@ -36,8 +36,12 @@ pythonpath = [
   "src"
 ]
 markers = [
-  "mount_snap: mark a test as using /SNS/SNAP/ data mount"
+  "integration: mark a test as an integration test",
+  "mount_snap: mark a test as using /SNS/SNAP/ data mount",
+  "golden_data(*, path=None, short_name=None, date=None): mark golden data to use with a test"
 ]
+addopts = "-m 'not integration'"
```
walshmm, line 43: spelling error?

walshmm:
NOTE: This doesnt seem to be able to run in Headless mode because of `tests/unit/ui/presenter/test_InitializeStatePresenter.py`
I dont recall this being a thing with normal pytest, but also dont see why this pr would cause this.

walshmm:
Upon failure of the test with message: 
```
FAILED tests/integration/test_diffcal.py::test_diffcal - RuntimeError: Object <class 'dict'> structure doesn't match the original: golden data must be regenerated.

```

The test state folder is not deleted, is this intentional?  I know its optional but this makes it less clean to rerun integration tests again and again.

walshmm:
What calibration folder/files was this tested against?
I just created a new Calibration folder based on the files currently in `SNS/SNAP/shared/Calibration` and this time it returned:
```
FAILED tests/integration/test_diffcal.py::test_diffcal - AssertionError
```

pyproject.toml
ekapadi: I'm not sure what you mean.  The "addopts" line works for me.  Can you clarify?

ekapadi:
I'll check the "headless mode" issue.  I've been successfully running all three modes: 1) pytest, (2) `python -m snapred` and (3) starting SNAPRed from within Mantid workbench.

ekapadi:
Yes.  The state directory is supposed to be deleted (unless you ask for it _not_ to be).  I think the issue is that the test kicked out before it could clean up.

ekapadi:
Note: the test state directory should _not_ exist prior to the test (but it should _tell_ you that). Is that possibly your issue?

ekapadi:
FAILED tests/integration/test_diffcal.py::test_diffcal - AssertionError: is this the error that you included (i.e. the golden-data structure), or is there another one?

ekapadi:
**What are you using as an "integration_test.yml" file?  Can you show me that?**

ekapadi:
**Even stranger: `pytest -k InitializeStatePresenter` actually passes for me.**

rboston628:
Where is the CIS script being run inside the CI/CD checks?  Should it run as its own check, separate from CI/tests?

ekapadi:
To clear up some confusion here.  This PR is about an integration-test _prototype_.  As we don't yet have a shared "developer's" partition on SNS, this is not yet integrated into the CI/CD checks.  Mostly this PR is about _fixtures_, and test-helper _utilities_ (mostly _context_ _managers_).

walshmm (approved):

PR #309, opened 2024-04-23
Create reduction UI elements

darshdinger:
## Description of work
This work was the capture the integration of the agreed upon UI elements to facilitate user interaction for the future Reduction feature within SNAPRed.

## Explanation of work
There was a previous existing `ReductionView` which was modified to incorporate the list of required user interactive UI elements as specified:
1. The lite mode widget (this can be identical to the difCal/normCal element)
2. An input run number text field. [ENABLED but RESTRICTED (For the initial release, this shall be constrained to only accept a single run number at a time)]
3. A checkbox enabling retention of unfocussed data [DISABLED]
4. A drop down allowing the user to select from a list of pixel masks [DISABLED]

An attempt was made to create a signal endpoint for the checkbox along with following the current accepted UI code structure for other additions.

**NOTE:** Slight updates will be needed to this view later on once backend implementation is complete. There are no endpoints to handle form creation within the `TestPanelPresenter` and therefore ReductionView was left as an inheriting OWidget class as apposed to a `BackendRequestView` class. This would require a request call associated to a path done within L83 - L86 for DiffCal and L88 - L91 for Normalization within TestPanelPresenter.

### Dev testing
Launch SNAPRed and view the reduction tab within the Test Panel.

### CIS testing
None needed for now!

## Link to EWM item

[EWM # 4775](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4775)

### DIFF
--- a/src/snapred/ui/presenter/TestPanelPresenter.py
+++ b/src/snapred/ui/presenter/TestPanelPresenter.py
@@ -37,8 +37,7 @@ class TestPanelPresenter(object):
 
         self.view.tabWidget.addTab(self.diffractionCalibrationWidget, "Diffraction Calibration")
         self.view.tabWidget.addTab(self.calibrationNormalizationWidget, "Normalization")
-        # TODO reenable in Phase 3
-        # self.view.tabWidget.addTab(self.reductionWidget, "Reduction")
+        self.view.tabWidget.addTab(self.reductionWidget, "Reduction")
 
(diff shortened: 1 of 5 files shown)


dlcaballero16 (approved):
Was able to view the reduction tab as described.

PR #311, opened 2024-04-24
Create WorkspaceMetadata object

rboston628:
## Description of work

This is for EWM 4808, which will add metadata tags to workspaces to indicate how the calibration process was handled.  The goal is to add these as logs.

The first step in adding these is to create a DAO for holding the needed values.  Later work will implement algorithm and service methods for handling this data.

## Explanation of work

Creates a simple DAO.

The literals ensure the values may be set only to one of a set of values.

The list of literals can be accessed by other code by importing the list, as seen in the test.

### Dev testing

1. Be Michael.
2. Make sure the values here are the ones desired.

### CIS testing

1. Be Malcolm.
2. Make sure the values here are the ones desired, namely:

- diffraction calibration states:
  - unset, this is a default value for internal use
  - none, meaning the user elected to use internal geometey
  - exists, SNAPRed found it
  - alternative, user elected to use some other diffcal file
- normalization states:
  - unset, this is a default value for internal use
  - none, meaning the user elected to not normalize
  - exists, SNAPRed found it
  - fake, the user elected to create a fake vanadium file

## Link to EWM item

[EWM#4808](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4808)

### DIFF
--- a/src/snapred/backend/dao/WorkspaceMetadata.py
+++ b/src/snapred/backend/dao/WorkspaceMetadata.py
@@ -19,7 +19,7 @@ diffcal_metadata_state_list = list(get_args(DIFFCAL_METADATA_STATE))
 NORMCAL_METADATA_STATE = Literal[
     "unset",  # the default condition before any settings
     "exists",  # the state exists and the corresponding .h5 file located
-    "alternate",  # the user supplied an alternate .h5 that they believe in usable
+    "alternate",  # the user supplied an alternate .h5 that they believe is usable
     "none",  # proceed without applying any normalization
     "fake",  # proceed by creating a "fake vanadium"
 ]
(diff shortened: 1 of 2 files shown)


mguthriem:
I checked the values and they are correctly what I asked for. Having said that I'm now worrying/wondering we should include `alternate` as an allowed operation for `NORMCAL_METADATA_STATE`. This would not be used in the currently conceived reduction workflow, but it might be helpful in special cases, say where we wanted to experiment with two different `rawVanadiumCorrection` recipes.

rboston628:
I can add `alternate` as an allowed option at the code level, and we can choose to not implement it

walshmm (dismissed):

walshmm (approved):

PR #312, opened 2024-04-24
User Docs button implementation

darshdinger:
## Description of work
This task encapsulated two main steps:

1. Figure out if `QtWebView` can run JS. If not, then what else should we use?
2. Implement within SNAPRed UI a mechanism that gives users access to the docs that Malcolm has produced for them.

`QtWebView` does have support for Java Script, see this link [here](https://doc.qt.io/qt-5/qml-qtwebview-webview.html). There is a method listed within the docs that is labeled **runJavaScript()**. This does what is says.  

**Note:** This documentation is for Qt 5.15. Our current Qt version is 5.15.8 so we should be good. :)

## Explanation of work
The implementation of this was simple and straight forward. I created a widget to facilitate both the creation of the UI element (a button in this case) and then logic to allow the button to be mapped to a local path and launch a web view window once clicked. In this case the path is: `/SNS/SNAP/shared/Malcolm/SNAPUserDocs/_build/html/index.html`. This path was added to the application.yml. Then within `main.py` this widget is implemented.

### Dev testing
1. Use `conda env remove --name SNAPRed` to remove your current SNAPRed env from within analysis.
2. Use `conda env create --file environment.yml` from within the parent SNAPRed directory within analysis.
3. Use `python -m pip install -e .` to make sure you re-install in editable mode.
4. Launch the snapred UI and insure that the `User Documentation` button from within the main window of SNAPRed functions properly.

### CIS testing
None needed, but welcome to insure that the docs are accessible. (Follow the same instructions above.)

## Link to EWM item

[EWM # 4841](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4841)

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,7 +2,7 @@ name: SNAPRed
 channels:
 - conda-forge
 - default
-- mantid/label/nightly
+- mantid/label/main
 dependencies:
 - python=3.10
 - pip
--- a/environment.yml
+++ b/environment.yml
@@ -27,5 +27,6 @@ dependencies:
 - python-build
 - boa
 - pycifrw
+- pyqtwebengine
 - pip:
   - check-wheel-contents==0.4.0
--- a/environment.yml
+++ b/environment.yml
@@ -2,7 +2,7 @@ name: SNAPRed
 channels:
 - conda-forge
 - default
-- mantid/label/main
+- mantid/label/nightly
 dependencies:
 - python=3.10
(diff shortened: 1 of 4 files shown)


dlcaballero16 (approved):
Tested and was able to view the user docs by clicking the button.

PR #317, opened 2024-04-26
Create Recipe base class

rboston628:
## Description of work

This creates a base class `Recipe` to serve as a template for future new-style recipes.

It cuts down on a lot of boilerplate inside the new-style recipes.  

Also adds a default validation method, which checks the ingredients can be cast to the ingredients, and that all of the workspaces in the grocery list exist.  This is probably good enough for most cases.

## Explanation of work

Creates an abstract base class (`ABC`) called `Recipe` to act as a base for future new-style recipes.

This provides several functions, some of which **must** be overwritten (abstract) and others of which **may** be overwritten, such as for outputs.

### CIS testing

This is internal and does not require CIS testing.

## Link to EWM item


src/snapred/backend/recipe/Recipe.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def cater(self, shipment: List[Pallet]) -> bool:
+        """
+        A secondary interface method for the recipe.
+        It is a batched version of cook.
+        Given a shipment of ingredients and groceries, it prepares, executes and returns the final workspaces.
+        """
+        for ingredient, grocery in shipment:
+            self.prep(ingredient, grocery)
+        return self.execute()
```
walshmm, line 111: would this mean its up the the implementation of execute to accrew the batched results?

rboston628: If a recipe needs results, then it has to overwrite this.  See what I did in the `ApplyWhateverItWas` recipe

rboston628: It would be way better if it handled outputs automatically, I just wasn't sure how to do that

walshmm (approved):

PR #320, opened 2024-04-30
Reduction Record Specification

darshdinger:
## Description of work
In this simple and quick task, the main object was to produce a pydantic object to hold a record of the information used within the reduction process. This is done to retain important meta data for reproducibility and tracking.

## Explanation of work
I had engaged in brief conservation with Malcolm to receive the requirements for this task. Within the EWM story is a list of requirements he gave me along with possibly adding some placeholders for further additions that may be incorporated later on. (Noting that this is liable to change.) The basic parameters added for now within the current implementation of `ReductionRecord` includes:

1. runNumbers - This accepts a list of runNumbers but by extension a single run number as well.
2. isLite - This is a boolean to simply track if Lite mode is employed or if full resolution data is being processed.
3. calibrationRecord - Tracking of which (diffraction) calibration is applied.
4. normalizationRecord - Tracking of which normalization is applied.
5. pixelGroupingParameters - Tracking all pixel groups used during the reduction.

## To test

Please ensure that this makes sense and is inline with the requirements listed within the EWM story below. Please give suggestions where further modification may be needed. No testing required.

## Link to EWM item

[EWM # 4877](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4877)


ekapadi (changes requested):
Requests:

1. Why is version _optional_?  Can it be mandatory?  All of the optional parms seem to bite us in the long run.
2. Add `stateId: ObjectSHA`;
3. Add `root_validator`: validate that the `stateId` matches those from `calibration` and from `normalization`.  Validate that the <run numbers> generate the same SHA -- see `StateConfig` for an example of this.
4. Brevity:  the types are both "record", so I suggest using `calibration` and `normalization` as the names -- the "Record" part is redundant.
5. I'm wondering about a `workspaces` field (as in the `CalibrationRecord`).

**Note:  most of these are things I need right now at my "reduction output" story, so if you don't add them, I'll need to.  :)**

ekapadi (approved):

PR #324, opened 2024-05-02
Create default calibration file from instrument geometry

rboston628:
## Description of work

When the user first initializes state, a special version folder (`v_0001`) is created, storing default calibration parameters.

When the user calibrates, additional folders are created storing calculated diffcal tables.

It is desired to have a default diffcal table calculated from the instrument's geometry created in the `v_0001` folder.

## Explanation of work

The `GroceryService` has a method to create a diffcal table from the instrument geometry by pulling a donor instrument (possible loading an empty instrument) and using SNAPRed algorithm `CalculateDiffCalTable`.

On state initialization. the `LocalDataService` will save a copy of this file to the default version folder.

### Dev testing

In your local SNAPRed calibration folder, go to statefolder of your favorite run.  Delete the folder called `diffraction` and all of its contents.

Enter your favorite run information (mine is 46680, Diamond, with 5deg-slice). You will be prompted to initialize state.  Click Yes.  Close the spurious duplicate popup window.

Initialize the state, and in the background some gears will turn to calculate the DIFC value.

When it is done press Okay.

In your dev calibration folder, navigate to  `diffraction/v_0001`.  You should see a file `diffract_consts_default_v0001.h5`, together with `CalibrationParameters.json`.  Those should be the only files.

In the smaller, non-calibration window, click on this workspace and look at it.  Make sure it looks like a diffcal table.  Should have `detid`, `difc`, `difa`, and `tzero` columns.  It should have one row per pixel in SNAPLite.  `detid` should increment from 0, `difc` should be a fairly large float (around 5000-7000), and `difa`, `tzero` are always zero.

Finish calibrating to the point where you can save.  (If using 46680, set the right FWHM to 10 and dMin to 0.8, using Bank grouping)

Save.  Make sure it saves to folder `v_0002`, and that it saves all of the workspaces from this run.

Delete the `v_0002` folder, but keep the `v_0001` folder.

Run it again.  This time it should not ask you about initializing state.

### CIS testing

Do the same, but run from workbench so that you can inspect the data more thoroughly.

## Link to EWM item

[EWM#4806](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4806)

### DIFF
--- a/tests/unit/backend/recipe/test_PixelGroupingParametersCalculationRecipe.py
+++ b/tests/unit/backend/recipe/test_PixelGroupingParametersCalculationRecipe.py
@@ -2,67 +2,60 @@ import json
 import unittest.mock as mock
 
 import pytest
+from snapred.backend.dao.ingredients.PixelGroupingIngredients import PixelGroupingIngredients
+from snapred.backend.dao.Limit import Limit
+from snapred.backend.dao.state.PixelGroupingParameters import PixelGroupingParameters
+from snapred.backend.recipe.PixelGroupingParametersCalculationRecipe import PixelGroupingParametersCalculationRecipe
+from snapred.meta.redantic import list_to_raw
 
-with mock.patch("mantid.api.AlgorithmManager") as MockAlgorithmManager:
-    from snapred.backend.dao.ingredients.PixelGroupingIngredients import PixelGroupingIngredients
-    from snapred.backend.dao.Limit import Limit
-    from snapred.backend.dao.state.PixelGroupingParameters import PixelGroupingParameters
-    from snapred.backend.recipe.PixelGroupingParametersCalculationRecipe import PixelGroupingParametersCalculationRecipe
-    from snapred.meta.redantic import list_to_raw
 
-    mockAlgo = mock.Mock()
-    MockAlgorithmManager.create.return_value = mockAlgo
+@mock.patch("snapred.backend.recipe.PixelGroupingParametersCalculationRecipe.BinnedValue")
+def test_execute_successful(mockBinnedValue):
+    # mock algorithm execution result and output
+    params = PixelGroupingParameters(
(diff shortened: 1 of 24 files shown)


rboston628:
Two issues slowing this up:
1. Callbacks.  I have a result from an algorithm executed with `MantidSnapper`.  If I don't call `.get()` on it, then the GUI will not replace it with the string value.  However, inside the unit tests it *is* being replaced, and calling `.get()` results in the error that strings have no `.get()` attribute.  So either the GUI fails, or the tests fail.
2. Versioning.  The logic for which version number gets saved and where the data service looks is rather hard to decipher.

tests/unit/backend/recipe/test_PixelGroupingParametersCalculationRecipe.py
rboston628, line 1: This additional twenty lines is to test the additional two lines I added to the recipe to handle the fact that in the unit tests the Callbacks do not automatically populate, while in the GUI they do.

darshdinger:
### Testing:

1. Starting with deleting all state directories:
image

2. Starting Calibration with specified parameters:
image

3. Initialize State (still two instances being produced) & success: 
image
image

4. Inspect the initialized state directories for `CalibrationParameters.json` & `diffract_consts_default_v0001.h5`:
image

5. Workspace does look like a diffcal table:
image

6. Set the right FWHM to 10 and dMin to 0.8, using Bank grouping & Save:
image
image

7. Inspect state directory again & confirm that DiffCal is saved under v_0002 directory:
image

8. Delete v_0002:
image

9. Run again & insure that SNAPRed does not initialize state:
image

darshdinger (approved):
Works as intended, looks good!

PR #325, opened 2024-05-02
Cache IPTS result to avoid calls to `GetIPTS` and OnCat

rboston628:
## Description of work

During our review, it was pointed out SNAPRed is pinging the OnCat server hundreds of times.

This is an unintended result of calling GetIPTS to construct filenames for neutron data.

This can be reduced by caching the IPTS results inside the LocalDataService.  Since this is the single point of contact with GetIPTS for the entire system (or should be), this should apply globally.

Caching should reduce the number of times calls to GetIPTS and therefore OnCat get made.

## Explanation of work

Very simple, added a dictionary cache of IPTS results to LocalDataService.  Whenever the `getIPTS` method is called, it first checks if the (run, instrument) pair is in the cache, and if not then and only then it calls `GetIPTS`.  Otherwise it just returns the cached value.

### Dev testing

The added unit test should be convincing.  Note this is calling `GetIPTS` directly, not any mock.

### CIS testing

Try the below script, turning off your internet between the first two calls.  That should demonstrate that it does not call OnCat after the value is cached.

``` python
import os
from mantid.kernel import amend_config

lds = LocalDataService()
lds.iptsCache = {}
correctPath = os.path.expanduser("~/SNAPRed/tests/resources/inputs/testInstrument/IPTS-456/")

new_config = {"data_dir": correctPath}
with amend_config(**new_config):
    assert correctPath == lds.getIPTS("123", "SNAP")

## PAUSE HERE AND TURN OFF YOUR INTERNET
## Then run the below while it can't send a GET to OnCat
assert False

with amend_config(**new_config):
    assert correctPath == lds.getIPTS("123", "SNAP")
assert True
```

The test would be more convincing if I didn't have to point directly at  the IPTS file to make it work.

## Link to EWM item

### DIFF
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -154,7 +154,6 @@ class LocalDataService:
 
     def getIPTS(self, runNumber: str, instrumentName: str = Config["instrument.name"]) -> str:
         key = (runNumber, instrumentName)
-        print(key)
         if key not in self.iptsCache:
             self.iptsCache[key] = GetIPTS(RunNumber=int(runNumber), Instrument=instrumentName)
         return str(self.iptsCache[key])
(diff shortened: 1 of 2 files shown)


dlcaballero16 (approved):
Ran the CIS script and was able to verify oncat was not called after turning off my internet

PR #326, opened 2024-05-02
Algo No Mo'

rboston628:
## Description of work

Following a conversation during SNAPRed review.

Within `GroceryService` and `LocalDataService`, remove construction of algorithms directly through `AlgorithmManager`, and instead rely on `MantidSnapper`.

`AlgorithmManager` should only appear in two places `MantidSnapper` and its test.

## Explanation of work

`AlgorithmManager`  was only meaningfully used in the following places:
- `LocalDataService` inside some write methods
- `GroceryService` in various places
- `CrystallographicInfoRecipe` which just used it to call the related algorithm; but some minor refactoring was needed to grab both of the outputs
- `PixelGroupingParametersCalculationRecipe`, which also just calls the related algorithm

The other occurrences were stray imports or similar.

Where changed, it was changed from

``` python
algo = AlgorithmManager(BakeCake.__name__)
algo.initialize()
algo.setProperty("CakeType", cakeType)
algo.execute()
```

to call from `MantidSnapper` like so

``` python
self.mantidSnapper.BakeCake(
    f"Bake a {cakeType} cake",
    CakeType=cakeType,
) 
```

within the `LocalDataService` and the `GroceryService`.

Inside the services, `MantidSnapper is initialized as
``` python
self.mantidSnapper = MantidSnapper(None, "Utensils")
```

### Dev testing

Do a grep test for `AlgorithmManager` occurring in the repo, and ensure it is only in `MantidSnapper`

### CIS testing

This is internal and doesn't need CIS testing

## Link to EWM item

### DIFF
--- a/src/snapred/backend/recipe/algorithm/CrystallographicInfoAlgorithm.py
+++ b/src/snapred/backend/recipe/algorithm/CrystallographicInfoAlgorithm.py
@@ -36,7 +36,7 @@ class CrystallographicInfoAlgorithm(PythonAlgorithm):
 
         # Load the CIF file into an empty workspace
         # use this to set the crystallography output
-        # also creat a mantid CrystalStructure object
+        # also create a mantid CrystalStructure object
         if not self.getProperty("CifPath").isDefault:
             ws = "xtal_data"
             self.mantidSnapper.CreateSingleValuedWorkspace(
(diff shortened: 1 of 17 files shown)


dlcaballero16 (approved):

src/snapred/backend/recipe/algorithm/CrystallographicInfoAlgorithm.py
```diff
@@ -21,21 +21,23 @@ def category(self):
...
     def PyExec(self):
-        cifPath = self.getProperty("cifPath").value
+        cifPath = self.getProperty("CifPath").value
         # run the algo
         self.log().notice("ingest crystallogtaphic info")
 
         # Load the CIF file into an empty workspace
-        if not self.getProperty("cifPath").isDefault:
+        # use this to set the crystallography output
+        # also creat a mantid CrystalStructure object
```
dlcaballero16, line 39: should be create

dlcaballero16 (approved):

PR #327, opened 2024-05-03
compare workspaces doesn't work like that

rboston628:
## Description of work

Several tests were written using
``` python
assert CompareWorkspaces(ws1, ws2)
```
This doesn't work the way we want, and actually always evaluates to `True` if the algorithm executes.

Instead, we want
``` python
assert CompareWorkspaces(ws1, ws2).Result
```
which will pass/fail in the correct circumstances.  However, it will also return the meaningless error "assert False where False = ..."

This is preferred,
``` python
res, messageTable = CompreWorkspaces(ws1, ws2)
if not res:
    raise Error(messageTable)
assert res
```
because then it prints out whatever error message was stored inside the table.

This introduces a helper in `util.helpers` that will make this comparison and fail in the correct circumstances.

It might be worth looking into using [`assert_almost_equal`](https://docs.mantidproject.org/nightly/api/python/mantid/testing/assert_almost_equal.html) from the Mantid testing fixtures.  This would work like
``` python
mantid.testing.assert_almost_equal(ws1, ws2)
```

Note that currently there are three failing tests.  These are comparing two workspaces that are not equal.

Tests still failing:
- `test_DiffractionSpectrumWeightCalculator.test_with_predicted_peaks`
- `test_LoadGroupingDefinition.test_load_from_hdf_file_with_instrument_donor`
- `test_LoadGroupingDefinition.test_load_from_hdf_file_with_instrument_file`

## Explanation of work

You should not see `CompareWorkspaces` anywhere in the code, except in `util.helpers`.

The unit tests are now using either `assert workspacesEqual()`, or the mantid tool `assert_almost_equal`.

The CIS scripts are now using mantid tool `assert_almost_equal`.

Make sure there is no `assert assert_almost_equal`.

### Dev testing

Grep the repo for CompareWorkspaces.  Verify it does not occur inside any unit tests.

Grep the repo for `assert assert_almost_equal`, just because I'm an idiot and have written that many times.

Run the test scripts:
- `lite_data_creation`
- `save_grouping_definition_with_xml`

Make sure the assertions pass.

These can run form the command line, eg:
``` bash
python tests/cis_tests/lite_data_creation.py
```
Finishing indicates it worked.

### CIS testing

This is an internal issue and does not require CIS testing.

## Link to EWM item

### DIFF
--- a/tests/unit/backend/recipe/algorithm/test_CalculateDiffCalTable.py
+++ b/tests/unit/backend/recipe/algorithm/test_CalculateDiffCalTable.py
@@ -10,7 +10,6 @@ from mantid.simpleapi import (
     mtd,
 )
 from mantid.testing import assert_almost_equal as assert_wksp_almost_equal
-
 from snapred.meta.Config import Resource
 
 
--- a/tests/unit/backend/recipe/algorithm/test_CalculateDiffCalTable.py
+++ b/tests/unit/backend/recipe/algorithm/test_CalculateDiffCalTable.py
@@ -4,6 +4,7 @@ from mantid.simpleapi import (
     DeleteWorkspace,
     mtd,
 )
+from mantid.testing import assert_almost_equal as assert_wksp_almost_equal
 
 # the algorithm to test
 from snapred.backend.recipe.algorithm.CalculateDiffCalTable import (
--- a/tests/unit/backend/recipe/algorithm/test_CalculateDiffCalTable.py
+++ b/tests/unit/backend/recipe/algorithm/test_CalculateDiffCalTable.py
@@ -9,7 +9,6 @@ from mantid.simpleapi import (
     DeleteWorkspace,
     mtd,
 )
-from mantid.testing import assert_almost_equal as assert_wksp_almost_equal
 from snapred.meta.Config import Resource
 
 
--- a/tests/unit/backend/recipe/algorithm/test_DiffractionSpectrumWeightCalculator.py
+++ b/tests/unit/backend/recipe/algorithm/test_DiffractionSpectrumWeightCalculator.py
@@ -6,6 +6,7 @@ import pytest
 from snapred.backend.dao import GroupPeakList
 from snapred.meta.redantic import list_to_raw
 from util.diffraction_calibration_synthetic_data import SyntheticData
+from util.helpers import workspacesEqual
 
 with mock.patch.dict(
     "sys.modules",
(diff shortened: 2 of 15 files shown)


tests/util/helpers.py
```diff
@@ -180,6 +182,30 @@ def deleteWorkspaceNoThrow(wsName: str):
         pass
 
 
+def workspacesEqual(Workspace1: str, Workspace2: str, **other_options):
+    if other_options == {}:
+        assert_wksp_almost_equal(Workspace1, Workspace2)
```
peterfpeterson, line 187: This should return `True` or the other branch shouldn't return anything. At the moment, this code will run both functions if the first one doesn't throw an `AssertionError`.

peterfpeterson (changes requested):
Please add docstrings to `workspacesEqual` and `workspacesNotEqual`. Normally these type of functions return nothing if their conditions are met and throw an `AssertionError` if the condition is not met. That is how pytest assumes python's [assert](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) works. Be like the language when possible.

tests/unit/backend/recipe/algorithm/test_CalculateDiffCalTable.py
```diff
@@ -9,6 +9,7 @@
     DeleteWorkspace,
     mtd,
 )
+from mantid.testing import assert_almost_equal as assert_wksp_almost_equal
```
peterfpeterson, line 12: I don't think this line is needed

tests/unit/backend/recipe/algorithm/test_DiffractionSpectrumWeightCalculator.py
```diff
@@ -49,6 +49,7 @@ def _setup_teardown():
         yield
         teardown()
 
+    @pytest.mark.xfail(strict=True)
```
peterfpeterson, line 52: Please add of `reason="To be fixed in EWM5043"` to the call to make it more obvious

tests/unit/backend/recipe/algorithm/test_LoadGroupingDefinition.py
```diff
@@ -311,9 +312,13 @@ def test_load_from_xml_file_with_instrument_file(self):
     #     self.do_test_load_with_instrument_name("xml")
 
     # test hdf
+    # TODO THIS IS BAD -- EWM 5043
+    @pytest.mark.xfail(strict=True)
```
peterfpeterson, line 316: Please add of `reason="To be fixed in EWM5043"` to the call to make it more obvious

peterfpeterson (changes requested):
Please add the EWM item to the `pytest.mark.xfail()` calls

peterfpeterson (approved):
Looks good now. That ended up being a bigger job than I expected.

PR #329, opened 2024-05-03
Change method for checking that workspaces are equal

peterfpeterson:
## Description of work

This is a change to how workspaces are compared to use a mantid facility. The facility throws an assertion error. Previous code did not check the result of `CompareWorkspaces` so it was giving false confidence in code.

related: #327

## Explanation of work

```python
assert CompareWorkspaces(ws1, ws2)
```
is verifying that the object returns anything. The proper check is actually to look at the `Result` attribute. The testing fixture from mantid does this and includes the difference in the assertion error that is thrown.

## To test

This is a refactor. If the tests pass, it is ready.

## Link to EWM item


peterfpeterson:
#327 is doing the work in the agreed upon way

PR #330, opened 2024-05-06
Reduction output: effective instrument geometry.

ekapadi:
At the last step of the reduction process, the real SNAP instrument is replaced with an effective instrument consisting of one pixel per output spectrum.  The physical location of each of these effective pixels is computed as the unmasked mean of the physical locations of original pixels in the group contributing to the spectrum concerned.

## Description of work

Note that since SNAPRed's `PixelGroupingParameters` were already calculated as a mean of several physical values over pixels contributing to each group, the required changes were implemented by adding a few additional parameters to this structure.  Both masked and unmasked calculations are available using the existing `GroceryService` and `SousChef` infrastructure.

## Explanation of work

This work consists of the following changes:

  * Add 'L2' and 'azimuth' to 'PixelGroupingParameters';

  * Clean up test code so that golden data can more easily be generated and compared.

  * Add an exception clause to deal with instabilities in `DetectorInfo.azimuthal()` which do not allow it to successfully complete the azimuth calculation for all pixels in the SNAP instrument. This will be spun off as a separate defect EWM#5073.

### Dev testing

Covered by existing unit tests for `PixelGroupingParametersCalculationAlgorithm`.  Golden data has been updated as required by the new attributes.

### CIS testing
The original CIS-test script used to test the masked `PixelGroupingParameters` calculation can be used to test these changes.

## Link to EWM item

[EWM#4798](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=4798)

See also:
[defect EWM#5073](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=5073)


src/snapred/backend/dao/state/PixelGroupingParameters.py
```diff
@@ -10,9 +10,15 @@ class PixelGroupingParameters(BaseModel):
...
 
-    # the average two-theta (radians) for this pixel grouping
+    # the mean sample to pixel distance (metres) for this pixel grouping
+    L2: float
+
+    # the mean two-theta (two times the polar angle in radians) for this pixel grouping
     twoTheta: float = Field(ge=0.0, le=2.0 * np.pi)
 
+    # the mean azimuth (radians) for this pixel grouping
+    azimuth: float = Field(ge=-np.pi, le=np.pi)
```
rboston628, line 20: Why is the range being selected this way, instead of $[0, 2\pi)$?  What would happen to a value like $3\pi$?  Can a modulo operator be slipped in here?

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
```diff
@@ -134,21 +137,43 @@ def PyExec(self):
...
+
+                groupMeanL2 += detectorInfo.l2(int(detIndex))
+
+                twoTheta = detectorInfo.twoTheta(int(detIndex))
+                groupMin2Theta = min(groupMin2Theta, twoTheta)
+                groupMax2Theta = max(groupMax2Theta, twoTheta)
+                groupMean2Theta += twoTheta
+
+                try:
+                    groupMeanPhi += detectorInfo.azimuthal(int(detIndex))
```
rboston628, line 162: I'm wondering if we shouldn't use
``` python
avgPos = [0,0,0]
for detID in detIDs:
    avgPos += detectorInfo.position()
groupMean2Theta = 2 * asin( avgPos .dotprod. z-axis)
groupMeanPhi = <logic to get phi from the vector>
```

src/snapred/backend/dao/state/PixelGroupingParameters.py
ekapadi: The problem is that `DetectorInfo::azimuthal` ends up using `std::atan2` without modification.  I would have preferred `[0, 2π)` myself for the azimuth, as I believe that it's more standard.  Note that this also means that there should be no need for a _modulo_ operator here.

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
```diff
@@ -134,21 +137,43 @@ def PyExec(self):
...
+                except RuntimeError as e:
+                    # Also entered as defect EWM#5073:
+                    # `DetectorInfo.azimuthal()` has issues in calculating ambiguous azimuth values:
+                    #   * by convention, these values can be set to zero, without overly affecting the mean value.
+                    if "Failed to create up axis orthogonal to the beam direction" not in str(e):
+                        raise
```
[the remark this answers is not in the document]
ekapadi: By definition, `raise` (by itself) rethrows the current exception, so `raise from e` would be redundant.

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
ekapadi: I considered something like this.  However, after looking at what `DetectorInfo` was doing to get the angles in the first place, I went with the idea of going with the implementation that modified the existing code as little as possible.
**Eventually, we want to replace `PixelGroupingParametersCalculation` with the equivalent _Mantid_ algorithm anyway.**  (There _is_ something that's called from the GUI that produces almost the same values, but we don't know the specifics yet.  Or, how much difficulty it would be to substitute it in.)

rboston628 (approved):
Code looks good and correctly accounts for solid angle factor in effective location.

PR #331, opened 2024-05-06
Default version is 0000, user versions start at 0001

rboston628:
## Description of work

When SNAPRed initializes, it creates an initial default directory, labeled `v_0001`.  Then when the user begins saving calibrations, these are saved as `v_0002` and up.

However, for normalization, the first user directory is `v_0001`, and increments from there.

It would make sense if the user's calibration versions began at 1 and incremented, while the default were treated specially.

This PR will make it so that the default is written as `v_0000`, and the user's calibrations written to `v_0001` and up.

But also, if the user prefers the original behavior, it can be fixed inside the `application.yml`.

## Explanation of work

The `Calibration`, `CalibrationRecord`, `Normalization`, and `NormalizationRecord` objects were given a default version number, which defaults to the initial version (say, 0).  They will thus always have a value.

The primary change was replacing
``` python
if not version:
    version = record.version
if not version:
    version = previousVersion + 1
```
with
``` python
if version is None:
    version = max(record.version, previousVersion + 1)
```
and then having `initializeState()` specify the version to write to as the starting version.

Since `version=0` would evaluate to false, several places in the data service using `if not version:` were changed to explicitly check against `None`. 

The logic for getting latest or specific versions was simplified to use python built-ins like `max()` and `index()`.

There is some additional work trying to standardize how versioning is printed in the service and the tests.

### Dev testing

In dev calibration folder, delete `diffraction` and `normalization` directories, if they exist.

Launch SNAPRed, go to diffcal workflow.  Use:
- 46680
- diamond
- bank

It asks you to initialize state.  Do so.  Check the directory `v_0000` is created with a calibration parameters file and a diffcal constants file.

On peak tweak, set FWHM = 10, recalculate, continue.

On assessment, verify that `v_0000` is in the drop-down as an option.  Continue.

Save.  Verify the directory `v_0001` exists and has the four (4) needed files.

Continue.  Do the exact same thing again.  On assessment verify `v_0000` and `v_0001` are both available, and after save verify that `v_0002` is created.

Close.  Open again.

In normalization workflow, enter
- 46680
- 46680
- diamond
- bank

Continue until you can save.

Verify the dir `normalization/v_0001` is created and it has four (4) files in it.  There is no `v_0000`.

### CIS testing

Verify the above, and also that you'd like it to work this way.

## Link to EWM item


rboston628:
This is what my directory tree looks like after running
image

rboston628:
Moved into draft until PR #333 is merged, which will have implications for this.

walshmm:
image

walshmm:
```
analysis-node15 - 2024-05-14 13:59:10,910 - ERROR    - InterfaceController - Failed to call service
Traceback (most recent call last):
  File "/SNS/users/wqp/git/SNAPRed/src/snapred/backend/api/InterfaceController.py", line 40, in executeRequest
    result = service.orchestrateRecipe(request)
  File "/SNS/users/wqp/git/SNAPRed/src/snapred/backend/service/Service.py", line 37, in orchestrateRecipe
    retValue = route(request.payload) if request.payload is not None else route()
  File "/SNS/users/wqp/git/SNAPRed/src/snapred/meta/decorators/FromString.py", line 59, in inner
    return func(*args, **kwargs)
  File "/SNS/users/wqp/git/SNAPRed/src/snapred/backend/service/CalibrationService.py", line 271, in loadQualityAssessment
    calibrationRecord = self.dataFactoryService.getCalibrationRecord(runId, useLiteMode, version)
  File "pydantic/decorator.py", line 40, in pydantic.decorator.validate_arguments.validate.wrapper_function
  File "pydantic/decorator.py", line 134, in pydantic.decorator.ValidatedFunction.call
  File "pydantic/decorator.py", line 206, in pydantic.decorator.ValidatedFunction.execute
  File "/SNS/users/wqp/git/SNAPRed/src/snapred/backend/data/DataFactoryService.py", line 100, in getCalibrationRecord
    return self.lookupService.readCalibrationRecord(runId, useLiteMode, version)
  File "pydantic/decorator.py", line 40, in pydantic.decorator.validate_arguments.validate.wrapper_function
  File "pydantic/decorator.py", line 134, in pydantic.decorator.ValidatedFunction.call
  File "pydantic/decorator.py", line 206, in pydantic.decorator.ValidatedFunction.execute
  File "/SNS/users/wqp/git/SNAPRed/src/snapred/backend/data/LocalDataService.py", line 530, in readCalibrationRecord
    recordFile = self._getFileOfVersion(recordPath, version)
  File "/SNS/users/wqp/git/SNAPRed/src/snapred/backend/data/LocalDataService.py", line 431, in _getFileOfVersion
    where = fileVersions.index(version)
ValueError: 0 is not in list
GETTING LATEST VERSIONS [0]
GETTING LATEST VERSIONS [1, 0]

```

walshmm:
empty diffcal removed from comparison, should be gtg

walshmm (approved):

PR #335, opened 2024-05-08
Multi run number validation frontend

darshdinger:
## Description of work
This task required the addition of a frontend validation step for input run numbers within the Reduction Workflow.

## Explanation of work
Within this work, logic was added within the view to validate some of the UI elements to ensure that both a pixel mask is selected and a valid digit run number is entered into the text box. This should be functional for both a single run number and a list of run numbers. In addition, within the workflow, logic is added to check that all entered run numbers has a IPTS path that exists and a calibration state that exists. (This may need to be changed to support uncalibrated reduction runs.)

## To test
Review code please.

## Link to EWM item

[EWM # 5007](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5007)


src/snapred/ui/workflow/ReductionWorkflow.py
```diff
@@ -1,40 +1,55 @@
...
+        useLiteMode = self._reductionView.litemodeToggle.field.getState()  # noqa: F841
+
+        self._reductionView.litemodeToggle.setEnabled(False)
+        self._reductionView.pixelMaskDropdown.setEnabled(False)
+
+        for runNumber in runNumbers:
+            try:
+                self.request(path="reduction/hasState", payload=runNumber).data
+            except Exception as e:  # noqa: BLE001
+                print(e)
```
rboston628, line 45: Why is this being handled inside this function for pixel mask drop down, and not another function?

src/snapred/ui/view/ReductionView.py
```diff
@@ -20,27 +20,31 @@ def __init__(self, pixelMasks=[], parent=None):
...
-        if self.dropDown.currentIndex() < 0:
-            raise ValueError("Please select a pixel mask")
+        runNumbers = self.runNumberField.text().split(",")
+        for runNumber in runNumbers:
+            if not runNumber.strip().isdigit():
+                raise ValueError(
+                    "Please enter a valid run number or list of run numbers. (e.g. 46680, 46685, 46686, etc...)"
+                )
+        if self.pixelMaskDropdown.currentIndex() < 0:
+            raise ValueError("Please select a pixel mask.")
```
rboston628, line 47: I think it would be better to change the UI to support a scrollable list of run numbers that the user can add to by filling in a field and pressing a "+" button

darshdinger: That is a change I can make.

src/snapred/ui/workflow/ReductionWorkflow.py
darshdinger: I was just following the template laid out within Diffcal workflow. Figured it would be good for consistency.

rboston628 (approved):
Ran the GUI, works smoothly.  It will parse individual numbers, or comma-separated strings of numbers.  IF a user enters something invalid in a string, it will not be added.

PR #336, opened 2024-05-08
Import and use algorithms like mantid simpleapi

rboston628:
## Description of work

This updates the `__init__.py` in the `algorithm` module so that algorithms are automatically put into the mantid simpleapi.

This means all algorithms can now be called as functions.

This also means MantidSnapper can easily import all of the algorithms, so that it is no longer needed to import an algorithm before using it with MantidSnapper.  This cleans up a lot of imports in files.

## Explanation of work

Created the `__init__.py` file to pull in all modules, initialize them, then put them into the simpleapi.

They can then be imported as follows:
``` python
import snapred.backend.recipe.algorithm
from mantid.simpleapi import WhateverAlgorithm
```

### Dev testing

Run the following (can run as command line script)
``` python
import pytest

with pytest.raises(ImportError):
    from mantid.simpleapi import Utensils

with pytest.raises(NameError):
    Utensils()

import snapred.backend.recipe.algorithm
from mantid.simpleapi import Utensils
Utensils()

```

### CIS testing

``` python
```

## Link to EWM item


tests/unit/backend/recipe/algorithm/test_CalibrationMetricExtractionAlgorithm.py
rboston628, line 1: This had a patch that created an error.  But also the patch was not necessary nor used.

src/snapred/backend/recipe/algorithm/__init__.py
rboston628, line 1: This is the main edit.

src/snapred/backend/recipe/algorithm/CalculateDiffCalTable.py
rboston628, line 1: This was previously setup to be called like an algorithm.  Now that is handled by the `__init__.py`

rboston628:
I ran this with both diffcal and normcal workflows and did not observe any errors

peterfpeterson (changes requested):
I'd prefer that the algorithms themselves kept the `AlgorithmFactory.subscribe()` calls so the algorithm looks like anybody else's. The code in `.../algorithm/__init__.py` is pretty nice and I like the comments in it.

PR #338, opened 2024-05-09
Move remaining references of pyqt5 to use qtpy

peterfpeterson:
## Description of work

Most of the repository has already been moved to qtpy, this changes the last two files and moves the references in the docs to something version agnostic.

## To test

This is a minor refactor. If the GUI comes up then it was done correctly.

*There isn't an EWM item*


rboston628:
There are still references in
- initialize_state_presenter.py
- normalization_tweak_peak_view.rst
- normalization_workflow.rst

rboston628 (approved):
no results found
image

PR #341, opened 2024-05-10
Create Reduction Service :tipping_hand_man: 

rboston628:
## Description of work

This (re)creates the Reduction service, which will handle all work needed for reduction.

With this PR, the Reduction service manages coordinating data and workspaces for the run of reduction.  Changes in other PRs will add in functionality for saving, loading, and indexing records.

For unit testing, this introduces two more utils,`InstaEats` and `WhateversInTheFridge`, which mock out the functionality of `GroceryService` and `LocalDataService`, respectively, allowing for more natural unit tests without a lot of mock constructions, and without the ambiguities of mocks. 

## Explanation of work

Added methods to the reduction service to orchestrate creating ingredients, fetching needed workspace data, and putting it all into reduction.

Had to make changes to `GroceryService` to ensure the normalization workspace was being loaded.

I added test utils in an effort to reduce our dependence on mocks.  Instead of mocking out LocalDataService functions, replace the LocalDataService object with WhateversInTheFridge, which can be more easily redirected for testing, without introducing the test vulnerabilities of mocks.  A lot more work is needed on these utils, which is planend for a future enabler.

### Dev testing

Make sure the Calibration Panel will launch.

Check my unit tests.  Run them locally and make sure no stray temp files are created.

Run the script up until the `assert False`, then run the last half, and make sure it completes with no issues.

### CIS testing

Run the reduction test script.  Run the top portion, which duplicates Michael's original efforts.  This will also verify that the normalization is being properly loaded.  It will stop when it reaches an `assert False`.

Run the bottom portion.  This calls everything from the service.  The asserts should show the service is reproducing what was done by the script prototype.

## Link to EWM item

[EWM#4711](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4711)


src/snapred/backend/dao/request/ReductionRequest.py
```diff
@@ -0,0 +1,16 @@
...
+
+from pydantic import BaseModel
+
+from snapred.backend.dao.state.FocusGroup import FocusGroup
+
+
+class ReductionRequest(BaseModel):
+    runNumber: str
+    useLitemode: bool
+    focusGroup: List[FocusGroup]
```
walshmm, line 11: The user doesnt actually choose the groups for reduction, it just reduces for each group applicable to a state.

src/snapred/backend/dao/request/ReductionRequest.py
```diff
@@ -0,0 +1,16 @@
...
+    calibrantSamplePath: str
+
+    smoothingParameter: float = Config["calibration.parameters.default.smoothing"]
```
walshmm, line 14: This should get pulled from a normalization record

src/snapred/backend/dao/request/ReductionRequest.py
```diff
@@ -0,0 +1,16 @@
...
+
+from snapred.backend.dao.state.FocusGroup import FocusGroup
+
+
+class ReductionRequest(BaseModel):
+    runNumber: str
+    useLitemode: bool
+    focusGroup: List[FocusGroup]
+    calibrantSamplePath: str
+
```
walshmm, line 13: We would probably also need a field to allow the user to provide a mask

src/snapred/backend/dao/request/ReductionRequest.py
rboston628: Agreed, I will edit SousChef to pull it directly

src/snapred/backend/dao/request/ReductionRequest.py
rboston628: This is now 
``` python
focusGroup:  Union[Optional[FocusGroup], List[FocusGroup]
```
with the intention that this property will be set inside the reduction service after it pulls all the groupings.  The alternative is creating a bunch of single-use requests.

src/snapred/backend/dao/request/ReductionRequest.py
rboston628: Would they provide it via a path?

src/snapred/backend/dao/request/FarmFreshIngredients.py
```diff
@@ -24,9 +24,6 @@ class FarmFreshIngredients(BaseModel):
     ## needs to be mandatory for diffcal
     cifPath: Optional[str]
 
-    # needs to be mandatory for normalizaiton/reduction
-    smoothingParameter: Optional[float]
-
```
rboston628, line 29: This should not be in FarmFreshIngredients, as it is internal data that should be loaded from a record.  The loading from record has been added tot he SousChef method.

src/snapred/backend/service/SousChef.py
```diff
@@ -168,10 +168,11 @@ def prepManyDetectorPeaks(self, ingredients: FarmFreshIngredients) -> List[List[
         return detectorPeaks
 
     def prepReductionIngredients(self, ingredients: FarmFreshIngredients) -> ReductionIngredients:
+        record = self.dataFactoryService.getNormalizationRecord(ingredients.runNumber, ingredients.useLiteMode)
```
rboston628, line 171: This will look up the smoothing parameter from the normalization record.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -264,7 +263,7 @@ def _constructNormalizationStatePath(self, stateId, useLiteMode):
             mode = "lite"
         else:
             mode = "native"
-        return f"{self._constructCalibrationStateRoot(stateId)}/{str(mode)}/normalization/"
+        return f"{self._constructCalibrationStateRoot(stateId)}{str(mode)}/normalization/"
```
walshmm, line 266: should `l258` also be changed to follow suit?

src/snapred/backend/data/LocalDataService.py
```diff
@@ -349,22 +348,26 @@ def _getVersionFromNormalizationIndex(self, runId: str, useLiteMode: bool) -> in
...
 
     @validate_arguments
-    def _constructCalibrationDataPath(self, runId: str, useLiteMode: bool, version: Version):
+    def _constructCalibrationDataPath(self, runId: str, useLiteMode: bool, version: Optional[Version]):
         """
         Generates the path for an instrument state's versioned calibration files.
         """
         stateId, _ = self._generateStateId(runId)
         statePath = self._constructCalibrationStatePath(stateId, useLiteMode)
+        if version == "*" or version is None:
```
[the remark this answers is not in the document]
rboston628: I had done it this way to accommodate your use of `version = "*"` in the script

tests/unit/backend/data/test_GroceryService.py
```diff
@@ -1217,39 +1214,48 @@ def test_fetch_grocery_list_diffcal_mask_cached(self):
...
+            tmpRoot.addFileAs(self.sampleDiffCalFilePath, diffCalTableFilename)
+            assert Path(diffCalTableFilename).exists()
 
             assert not mtd.doesExist(diffCalMaskName)
             assert not mtd.doesExist(diffCalTableName)
             items = self.instance.fetchGroceryList(groceryList)
             assert items[0] == diffCalMaskName
             assert mtd.doesExist(diffCalMaskName)
             assert mtd.doesExist(diffCalTableName)
+        # NOTE make sure the state root method is returned to its original pristine form
```
ekapadi, line 1239: We now have both a `state_root_override` context manager, and a `state_root_override_fixture` -- please use whichever is appropriate here.  For a test, the fixture is more reliable in terms of what happens when the test fails.

tests/unit/backend/data/test_LocalDataService.py
ekapadi, line 1: Just double checking here: `WhateversInTheFridge` is a mock for `LocalDataService`, and as such probably should _not_ be used in the `LocalDataService` tests themselves.

tests/util/WhateversInTheFridge.py
ekapadi, line 1: Be careful here:  if possible, please use our existing context manager `state_root_override` and the existing fixture `state_root_override_fixture`.  Also, please use the standard library `contextlib.contextmanager` decorator -- there's no need to implement a context manager "by hand".

ekapadi (changes requested):
I found a few minor issues.  The key one is probably that `_constructStateRoot` won't really work as a name -- it probably still should be `_constructCalibrationStateRoot` -- you'll see why in the associated comments.  **Also:  I really wasn't sure why you weren't using the existing context manager `state_root_override` and it's analogous fixture `state_root_override_fixture` -- these work well, and have associated unit tests.**

src/snapred/backend/data/GroceryService.py
```diff
@@ -879,10 +886,8 @@ def fetchGroceryList(self, groceryList: List[GroceryListItem]) -> List[Workspace
                         item.version = self.dataService._getVersionFromCalibrationIndex(
                             item.runNumber, item.useLiteMode
                         )
-                    record = self.dataService.readCalibrationRecord(item.runNumber, item.useLiteMode, item.version)
-                    item.runNumber = record.runNumber
```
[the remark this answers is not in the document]
rboston628: I was on the verge of editing all of them to just take the item.  I think it was the diffcal output workspace name method that has around 5 or 6 arguments to it and I just got tired of writing it all out.  I had originally made them to use arguments to avoid forcing users to create grocery list items just to get a workspace or file name; but I don't think these methods are ever called outside of the grocery service anyway.

src/snapred/backend/data/LocalDataService.py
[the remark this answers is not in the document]
rboston628: 1. The state root (`/SNS/SNAP/etc/etc/Powder/<stateid>`) is used for both calibration and normalization, and it's only defining property is it has the state id in it.  Could the reduction path be just `_constructReductionRoot`?  I can change it back to `_constructCalibrationStateRoot`, it just suggests there should be a `_constructNormalizationStateRoot` method, which there isn't, since they use the same state.
3. It's still a mess, but the way git renders diffs when methods are moved makes me not want to fix it too much.  I agree, the difference in a path to a containing directory and the path to a file could use further clarification, and maybe we should have `_constructStateRootPath` instead?

src/snapred/backend/service/SousChef.py
```diff
@@ -167,11 +169,28 @@ def prepManyDetectorPeaks(self, ingredients: FarmFreshIngredients) -> List[List[
...
+        # some of the reduction ingredients MUST match those used in the calibration/normalization processes
+        calibrationRecord = self.dataFactoryService.getCalibrationRecord(
+            ingredients.runNumber, ingredients.useLiteMode, version
+        )
+        normalizationRecord = self.dataFactoryService.getNormalizationRecord(
+            ingredients.runNumber, ingredients.useLiteMode, version
+        )
+        # grab nformation from records
+        ingredients.calibrantSamplePath = calibrationRecord.calibrationFittingIngredients.calibrantSamplePath
+        ingredients.cifPath = self.dataFactoryService.getCifFilePath(
```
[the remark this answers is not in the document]
rboston628: That's a better idea.  I've been trying to get rid of these multi-split methods of parsing filenames, as they are fragile and fail with meaningless arguments.   To answer what it's doing: it is taking a full path to a file, and trying to extract out just the name of the file, which is what the data factoy method expects.

tests/unit/backend/data/test_LocalDataService.py
rboston628: I agree, it definitely should not be used in these tests.  It's intended for tests of the other services.

tests/util/WhateversInTheFridge.py
rboston628: I'll see if I can use the context manager.

src/snapred/backend/service/ReductionService.py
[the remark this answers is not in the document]
rboston628: I just copied whatever NormalizationService had and removed the word "Normalization" :P

src/snapred/backend/data/GroceryService.py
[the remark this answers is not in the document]
rboston628: I had written a much easier method of dealing with versioning.  The current approach is kind of a labyrinth.  I was going to just have an `Indexor` data service whose only job is to keep track of version numbers in the index and the file tree and handle returning correct version numbers.  That would remove the need for the `"*"` special case.

tests/util/WhateversInTheFridge.py
rboston628: I moved my context manager to the `state_helpers` file and named it correspondingly, since the two are definitely related.  I left it as a class, as I also want it to have the other two methods defined on it.

src/snapred/backend/service/SousChef.py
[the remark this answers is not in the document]
rboston628: I was able to replace a lot of these with `Path().stem`

tests/unit/backend/data/test_GroceryService.py
rboston628: As discussed, my context manager does something a bit different, but because they're related I moved mine to the same helper file and renamed it in a corresponding way.

src/snapred/backend/data/LocalDataService.py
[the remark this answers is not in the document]
ekapadi: I still am using "Root" to mean: a _directory_ path, but with additional special meaning -- in this sense, "root path" would be redundant..  Unfortunately, the root isn't even necessarily the _stateID_ root in all cases.  If I'm remembering correctly, for reduction, one is the parent of the other (it's the special "reduction data root" directory with the IPTS name in it, before we even get to the state-id part).  
   Anyway, what _I_ would do with this is actually have a separate `_constructNormalizationStateRoot` method, which then just calls the other method -- this actually solves a problem that occurs during mocking, that the nested paths are all interdependent.  **Also I'd note that if we're not sure about something, or not completely happy with a proposed change, we should probably just leave it the way it was.**

ekapadi (approved):
I ran all of the tests from _scratch_ again.  Everything went fine!

PR #343, opened 2024-05-10
Added calls to RemovePromptPulse after neutron data gets loaded

dlcaballero16:
## Description of work

RemovePromptPulse is required to be called after neutron data is loaded in to fine tune the data.
## Explanation of work

Calls to RemovePromptPulse were added in neutron loading functions inside the GroceryService. The params are to be retrieved from SNAPInstPrm.json, defined by intsrument.config in the application.yml.

### Dev testing

 Make sure unit tests pass.

### CIS testing

Make sure running DiffCal and Normalization still works. 

I recommend running a cis_test script. I ran lazy_raw_vanadium.py. Make sure SNAPInstPrm.json is updated, or update your instrument.config in application.yml to point to the updated one.

## Link to EWM item

[EWM#5013](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5013)


src/snapred/backend/data/GroceryService.py
```diff
@@ -550,6 +550,18 @@ def fetchNeutronDataSingleUse(self, runNumber: str, useLiteMode: bool, loader: s
...
+        width = config["width"]
+        frequency = config["frequency"]
+        self.mantidSnapper.RemovePromptPulse(
+            "Removing prompt pulse",
+            InputWorkspace=workspaceName,
+            OutputWorkspace=workspaceName,
+            Width=width,
+            Frequency=frequency,
+        )
+        self.mantidSnapper.executeQueue()
```
rboston628, line 563: Won't this end up performing the operation on the workspace twice?  Because it's performed on the original data on load, then performed a second time on the clone.

src/snapred/backend/data/GroceryService.py
```diff
@@ -621,6 +633,19 @@ def fetchNeutronDataCached(self, runNumber: str, useLiteMode: bool, loader: str
...
```
rboston628, line 647: I think this belongs better to the `FetchGroceriesRecipe` (or the algorithm)

tests/unit/backend/data/test_GroceryService.py
```diff
@@ -65,10 +67,31 @@ def setUpClass(cls):
         # create some sample data
         cls.sampleWS = "_grocery_to_fetch"
 
-        LoadEmptyInstrument(
-            Filename=cls.instrumentFilePath,
+        CreateSampleWorkspace(
+            OutputWorkspace=cls.sampleWS,
+        )
```
rboston628, line 72: Duplicate function call

tests/unit/backend/data/test_GroceryService.py
```diff
@@ -65,10 +67,31 @@ def setUpClass(cls):
...
+        CreateSampleWorkspace(
```
rboston628, line 73: Can you time how much this slows down unit tests (if at all)?  My concern is this workspace is used in an awful lot of tests.

tests/resources/inputs/SNAPInstPrm.json
```diff
@@ -24,5 +24,7 @@
     "delThWithGuide": 6.40e-3,
     "alpha": 0.1,
     "beta_0": 0.02,
-    "beta_1": 0.05
+    "beta_1": 0.05,
```
rboston628, line 27: make sure Malcolm didn't delete this line in the One True SNAPInstPrm.json

src/snapred/backend/recipe/FetchGroceriesRecipe.py
```diff
@@ -1,14 +1,22 @@
...
 logger = snapredLogger.getLogger(__name__)
 
 
 @Singleton
 class FetchGroceriesRecipe:
+    def __init__(self):
+        # NOTE: workaround, we just add an empty host algorithm.
+        utensils = Utensils()
+        utensils.PyInit()
+        self.mantidSnapper = utensils.mantidSnapper
```
rboston628, line 18: since you're import mantid snapper now, remove the import of the FetchAlgo and then replace the use of `algo = FetchAlgo()` etc with something like
``` python
res = self.mantidSnapper.FetchGroceriesAlgorithm(
    Filename=filename,
    OutputWorkspace=workspace,
    LoaderType=loader,
    LoaderArgs=loaderArgs,
    {instrumentPropertySource: instrumentSource},
)

data["result"] = True
data["loader"] = res.LoaderType
data["workspace"] = res.OutputWorkspace
```

tests/unit/backend/recipe/test_FetchGroceriesRecipe.py
```diff
@@ -115,6 +116,21 @@ def test_fetch(self):
             Workspace2=res["workspace"],
         )
 
+    @mock.patch("snapred.backend.recipe.FetchGroceriesRecipe.FetchAlgo")
```
rboston628, line 119: If you do the above suggestion, you can remove this patch and use
``` python
def test_fetch_with_load_event_nexus(self):
    mockReturn = NamedTuple(LoaderType="LoadEventNexus", OutputWorkspace=self.fetchedWSname)
    self.instance.mantidSnapper.FetchGroceriesAlgo = mock.Mock(return_value=rmockReturn)
```

tests/resources/inputs/SNAPInstPrm.json
dlcaballero16: the one on SNS has been updated correctly

src/snapred/backend/recipe/FetchGroceriesRecipe.py
rboston628: I tried doing this.  Unfortunately it changes thrown exceptions so that several negative tests fail.  Fixing it requires fixing things like MantidSnapper and GenericRecipe and is not worth it for this PR

rboston628 (approved):
After changing my SNAPInstPrm.json file with the new parameters, then clearing the dev folder, I ran both diffcal and normalization.  I was able to initialize state, load past calibrations at the assess stage, normalize, and save in both workflows.

PR #333, opened 2024-05-07
Add lite/native mode folders

dlcaballero16:
## Description of work

This work is to add the mode(lite or native) to the paths where records are stored.
## Explanation of work

I started out with modifying the base function call that creates the path `_constructCalibrationStateRoot` in LocalDataService.py. If LiteMode is enabled, modify the path so that "lite" is added to the path, otherwise use "native". This function now takes in `useLiteMode` as a boolean parameter. Because of this, I had to update every related function call, and even more related calls to the already related calls to include the new boolean. 

### Dev testing

Make sure unit tests pass and do CIS testing below.

### CIS testing

Run both calibration and normalization tabs and complete the save step. Make sure the records are stored in the correct "lite" path.

## Link to EWM item

[EWM#4719](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4719)

PR #345, opened 2024-05-11
Validate function signatures inside data services

rboston628:
## Description of work

There are several functions requiring some combination of:
- runId
- useLiteMode
- version
- name
- etc...

Most functions were written in the form
``` python
def someFunction(runId: str, useLiteMode: bool, other_arguments)
```
However, prior to PR #333 , the lite mode flag was still considered optional and hardwired to `True`.  Several functions needed a run number, a version number, and an optional lite mode flag.  This means their signature was
``` python
def someFunction(runId: str, version: str, useLiteMode: bool = True)
```
This led to two contradictory conventions of argument order (useLiteMode always after runId, or version after runId), which led to contradictory calls, which in some places created erroneous values.

The goal here is to ensure functions are always being called with arguments in the correct order by adding validation decorators, then standardizing the argument signature.

## Explanation of work

The **only** changes should be
1. adding the decorator
2. switching argument order to `runId, useLiteMode, version` in a definition
3. switching argument order inside a function call
4. inside tests, replacing mocks with objects that pass validation
5. miscellaneous changes to make validation work

### Dev testing

Make sure unit tests pass.  Try opening one at random (say `test_DataFactoryService`) and change order of function parameters, and see that they will fail of the order is incorrect.

Clear dev calibration folder.

Ensure state initializes correct.

On tweak peak, ensure all changes complete (try changing grouping, dMin/dMax, threshold, smoothing).

On assessment, check drop down.  Should show `v_0001`.  It won't load (no calibration record) but make sure it is there.

On save, ensure it can save and check your directory that it exists.

Go back through. on assessment ensure `v_0002` is in dropdown and will load.

Do similar for normalization.

### CIS testing

``` python
```

## Link to EWM item


ekapadi:
This is a really cool way to fix this!  However, I'm wondering if a more straightforward approach would have been just to combine these arguments into one small structure (e.g. `RunSpec` or `RunInstance`)?  What I worry about adding the decorator, is that it's now something that needs to travel along with the codebase _forever_ ...

PR #339, opened 2024-05-09
Ewm4788 reduction recipe 3

walshmm:
## Description of work

This is the implementation of the reduction recipe.
It composes the previous enablers into the calculation unit that reduces a sample, given a diffcal, and a normalization.

### Dev testing

Ensure tests pass, run the cis test to completion

### CIS testing

Run the attached CIS test script and inspect the results [test script](https://github.com/neutrons/SNAPRed/pull/339/files#diff-38adf11fad631564d27039a2bae49ee84234161229417ebd282a3fe22a0982c1)

## Link to EWM item

[EWM#4788](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4788)

PR #348, opened 2024-05-14
Unpatch GroceryService tests

rboston628:
MERGE AFTER PR #339 

## Description of work

The grocery service unit tests had a bunch of unnecessary tests where simple mocks were sufficient.  To avoid more confusiong over patches in unit tests, I refactored all of these to use simple mocks.  There should be only three patches left.

## Explanation of work

In a unit test, it is not necessary to use patches to mock out methods of the class being tested, or to mock out methods of member objects belonging to the class being tested.

The tests of grocery service had dozens of tests mocking out
- methods within `GroceryService`
- methods within `LocalDataService`, which is a member object of `GroceryService`
when these could instead be handled with simple mocks.  That is,
``` python
@mock.patch.object(GroceryService, "nameOfFunction")
@mock.patch.object(LocalDataService, "nameOfAnotherFunction")
 def test_both_of_these_functions(self, mockFunction, mockAnotherFunction):
    pass
```
could instead be written
``` python
def test_both_of_these_functions(self):
    self.groceryService.nameOfFunction = mock.Mock()
    self.groceryService.dataService.nameOfAnotherFunction = mock.Mock()
```

### Dev testing

Nothing really to test, just wait for all of the tests to pass

### CIS testing

Not needed, purely internal matter

## Link to EWM item

### DIFF
--- a/tests/unit/backend/data/test_GroceryService.py
+++ b/tests/unit/backend/data/test_GroceryService.py
@@ -259,7 +259,7 @@ class TestGroceryService(unittest.TestCase):
     def test_diffcal_table_filename(self):
         # Test name generation for diffraction-calibration table filename
         self.instance.dataService._constructCalibrationDataPath = mock.Mock(return_value="nowhere/")
-        res = self.instance._createDiffcalTableFilename(self.runNumber, self.version)
+        res = self.instance._createDiffcalTableFilename(self.runNumber, self.useLiteMode, self.version)
         assert self.difc_name in res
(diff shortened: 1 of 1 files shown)


walshmm:
I feel as though this would conflict with the [reduction recipe pr](https://github.com/neutrons/SNAPRed/pull/339), so I ask that it is merged after

darshdinger (approved):
Looks good to me!

PR #349, opened 2024-05-14
added workbench arg to start snapred in workbench and preload necessa…

walshmm:
…ry imports

## Description of work

fixes snapred workbench startup issues by preimporting webui widget.

## Explanation of work

its necessary to start workbench through snapred because workbench will by default create a qapp, which messes with the import order of one of our dependencies.

## To test

launch snapred with 
```
snapred --workbench
```

see that workbench starts,
open snapred ui from workbench
see that the ui now comes up

### DIFF
--- a/src/snapred/__main__.py
+++ b/src/snapred/__main__.py
@@ -86,7 +86,7 @@ def _createArgparser():
         action="store_true",
         help="Stop the error reporter from opening if you suffer an exception or crash.",
     )
-    parser.add_argument("script")
+    parser.add_argument("script", default=None)
     return parser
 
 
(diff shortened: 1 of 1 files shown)


peterfpeterson (commented):
I think this is in the wrong place. mantid starts snapred in [this file[(https://github.com/mantidproject/mantid/blob/main/qt/python/mantidqtinterfaces/mantidqtinterfaces/SNAPRed.py)

dlcaballero16 (approved):
Tested and was able to open SNAPRed from workbench

dlcaballero16 (approved):
Retested and opening SNAPRed through workbench works still

PR #350, opened 2024-05-16
Setup algorithm to normalize by current the correct way

rboston628:
## Description of work

When trying to test PR #339, we encountered repeated issues that then `NormaliseByCurrent` is called on a workspace that has already been normalized by current, then the mantid algorithm just fails.  It just stops, fails, and crashes the app.

Instead, if it's already been normalized, then it just doesn't get normalized again.

We're already normalized!  We can't normalize any further!

## Explanation of work

Made a very simple algorithm that checks if the `NormalizationFactor` log exists on a workspace.

If it does, that means the workspace has already been normalized by current, and instead it just clones the input into the output.

If it doesn't, then it has not been normalized by current, so it calls the mantid algorithm.

## To test

The unit tests should cover the relevant scenarios.

## Link to EWM item

### DIFF
--- a/src/snapred/backend/recipe/ReductionGroupProcessingRecipe.py
+++ b/src/snapred/backend/recipe/ReductionGroupProcessingRecipe.py
@@ -62,8 +62,8 @@ class ReductionGroupProcessingRecipe(Recipe[Ingredients]):
             RebinOutput=False,
         )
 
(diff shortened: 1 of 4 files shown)


darshdinger (approved):
Looks good. :)

PR #353, opened 2024-05-28
minimal change to fix defect

rboston628:
### Dev testing

In your dev calibration folder, delete the state folder for the run you plan to test.

In `next`, open the GUI, try to run the diffcal workflow.  It should eventually fail with the ever-so-helpful "NoneType object has has no attribute ..." message.

Switch to this branch.  Delete the state folder from your dev calibration folder and try again.  You should be able to get to the peak tweak view.

## Link to EWM item

### DIFF
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -879,10 +879,8 @@ class GroceryService:
                         item.version = self.dataService._getVersionFromCalibrationIndex(
                             item.runNumber, item.useLiteMode
                         )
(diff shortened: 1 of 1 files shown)


dlcaballero16 (approved):
Tested as suggested and was able to verify the fix.

mguthriem:
I tested this too. I verified that the "NoneType object has has no attribute ..."error has gone away, but for me the subsequent diffcal workflow fails with the FitPeaks error and I cannot get to the tweak peek tab.

PR #354, opened 2024-05-29
Ewm4867 expose chi2

darshdinger:
## Description of work
This work was to enable exposure of the `maxChiSq` to the user within the Diffcal tweak peak view of the Diffcal workflow. 

## Explanation of work
This work largely consisted of updating pydantic models to include a parameter for max chi squared. Frontend changes included modification of views and the DiffCal workflow. Backend changes included changes to CalibrationService, DiffractionCalibrationRecipe, and PDCalibration.

## To test
Please open snapred UI and go through DiffCal workflow. Insure that changing the new maxChiSq parameter works as intended within the view.

[EWM # 4867](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4867)


src/snapred/ui/view/DiffCalTweakPeakView.py
```diff
@@ -56,8 +57,10 @@ def __init__(self, jsonForm, samples=[], groups=[], parent=None):
         # create the run number field and lite mode toggle
         self.runNumberField = self._labeledField("Run Number")
         self.litemodeToggle = self._labeledField("Lite Mode", Toggle(parent=self, state=True))
+        self.maxChiSqField = self._labeledField("Max Chi Sq", QLineEdit(str(self.MAX_CHI_SQ)))
```
rboston628, line 60: I believe this value is set to a constant equal to the default in the application.yml.  Is that okay here?

darshdinger: That's supposed to pre-populate the text field with the default value from within application.yml. (100.0) So it's just a place holder but can be modified by the user.

rboston628 (approved):
Ran both workflows to completion and saved,

The peak tweak view is not twitching, and adjusting maxChiSq allows me to select/deselect peaks.  Making it lower, only better-fit peaks appear in blue.

image

PR #355, opened 2024-05-30
fix bug in FocusSpectra that prevented workflows :space_invader: 

rboston628:
## Description of work

This should fix the bug that has been preventing us from running diffcal or normalization workflows.

## Explanation of work

The difference is 
``` python
self.getPropertyValue()
```
vs
``` python
self.getProperty().value
```
Most properties are string properties, and the two are equivalent for strings.

For a boolean property, the second is [the actual boolean](https://docs.mantidproject.org/nightly/api/python/mantid/kernel/BoolPropertyWithValue.html#mantid.kernel.BoolPropertyWithValue.value) `True` or `False`, while the first is a [string representation of the boolean](https://docs.mantidproject.org/nightly/api/python/mantid/api/Algorithm.html#mantid.api.Algorithm.getPropertyValue).

### Dev testing

In dev calibration, delete the state folder.

Run calibration, including initializing the state.  
At tweak peak, try entering nonsense valeus for FHWM, like 0, -1, or "matthew".  Confirm it gives an error and can be corrected.

You should be able to get all the way to the save step.

At save step, try entering nonsense as the version number, and confirm it requires an integer.

Run back through workflow, confirm it does not give an error about no applicable records.

Do the same in normalization/

### CIS testing

Run calibration.  Use a legit run, make sure it saves.  At tweak peak peek, try changing absolutely everything: grouping, dmin, threshold, etc.

Do the same in normalization.

## Link to EWM item

[EWM#5237](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5237)

### DIFF
--- a/src/snapred/backend/dao/calibration/CalibrationIndexEntry.py
+++ b/src/snapred/backend/dao/calibration/CalibrationIndexEntry.py
@@ -18,7 +18,7 @@ class CalibrationIndexEntry(BaseModel):
 
     runNumber: str
     useLiteMode: bool
-    version: Optional[str]
+    version: Optional[int]
     appliesTo: Optional[str]
     comments: str
     author: str
--- a/src/snapred/backend/dao/normalization/NormalizationIndexEntry.py
+++ b/src/snapred/backend/dao/normalization/NormalizationIndexEntry.py
@@ -19,7 +19,7 @@ class NormalizationIndexEntry(BaseModel):
     runNumber: str
     useLiteMode: bool
     backgroundRunNumber: str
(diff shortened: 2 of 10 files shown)


mguthriem:
I tested this with the difcal workflow: entire workflow is working again and, as a bonus, the error that allowed non-integer versions to be created is gone.

Norm cal workflow was also executed completely.

walshmm (approved):

PR #356, opened 2024-05-30
Ewm4880 reduction tab UI orchestration

walshmm:
**NOTE: This pr is against the branch** `ewm4711-reduction-service`
**This is to highlight my changes as my base branch(the above) is not merged into next**

## Description of work

I refactored out the jsonschema/jsonform widgets as they werent being actively used in our workflow forms, and only served to confuse and obfuscate.

I also implemented the most basic form of orchestration for the Reduction Workflow.
It is still kinda up in the air how we plan to handle saving as I dont think there is an explicit story out there specifying how the view for that should look.
Its still not certain saving will be part of this release.

**NOTE2:** CalibrantSample is hard coded here, we need to address where to get this and a couple other parameters that should be pulled from the calibration record/vanadium record.  

## To test

Load up the reduction ui and try to emulate the results of the [cis test script ](https://github.com/neutrons/SNAPRed/blob/next/tests/cis_tests/reduction_recipe_script.py)

## Link to EWM item

[EWM#4880](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4880)


walshmm:
superceded by pr #364

PR #357, opened 2024-05-30
Fix reduction cis test script

walshmm:
unrevert groceryservice such that diffcal_table is looked up via targetRunnumber -> calibrationRecord -> table in calib version named with calibRecord.runnumber

## To test

Setup your calibration directory such that it has a valid calibration and a valid normalization in lite mode.
Run the cis test script for reduction to completion.


darshdinger (approved):
This works!

rboston628:
Please stop doing this, it prevents the workflow from running if someone is in a newly initialized state.

PR #360, opened 2024-06-03
Indexer :robot::card_index:

rboston628:
## Description of work

On `next`, much of the `LocalDataService` is taken up with methods to track version numbers by comparing a directory tree and an `Index.json` file and picking a version.  Much of the code in the methods and in the DAOs is repeated.  The exact rules for what version is returned are kind of murky and not written in code anywhere; the version behavior emerges as different service methods interact with each other.

To simplify a large part of the `LocalDataService`, and to make version numbering easier to track, this introduces the `Indexer` data component.

`Indexer` is not a `Singleton`.  Each `Indexer` is created only as needed by the `LocalDataService`.  For each state, there could be up to four `Indexer`s corresponding to
- `/abc123/lite/diffraction/`
- `/abc123/lite/normalization/`
- `/abc123/native/diffraction/`
- `/abs123/native/normalization/`
Each `Indexer` is in charge of managing the versions inside its directory by examining its directory tree and comparing against the `Index.json` file.

`Indexer` manages the following:
- version numbering
- reading/writing `IndexEntry`s
- read/writing `Record`s
- reading/writing `CalculationParameters`s (i.e. the objects called `Calibration` and `Normalization` which are saved as `"CalibrationParameters.json"` and `"NormalizationParameters.json"`)

`Indexer`s can specialize as pertaining to Calibration, Normalization, or Reduction.  If specialized, they will handle the particular daughter classes of `IndexEntry`, `Record`, and `CalculationParameters` that correspond to the speciality.

## Explanation of work

[See Diagram Here](https://drive.google.com/file/d/1z9dimFbDdv8ILpYQbQ7vYqWb-AeO9RJ0/view?usp=sharing)

[See More Detailed Diagram Here](https://drive.google.com/file/d/12OrCI8K0DtKmrx0-YDisY7f3LV1U49Bx/view?usp=sharing)

The most important work is in the `LocalDataService`, which has had much of its former functionality replaced with `Indexor` objects.  These handle creating paths, assigning versions, and reading/writing most of the things the `LocalDataService` used to do.  The `LocalDataService` now only has special jurisdiction over things like StateConfigs and preparing the state roots.

There were basically three ways to specify a version before.  Either by adding a version to a record, or by passing a version as an argument, or by not doing either and then it would find a next version.  This meant there were multiple sources of truth on what version should be written.  

From now on, in order to create a versioned object, it **SHOULD ALWAYS** be created by a call to the appropriate creation method inside `Indexer`.  This call can receive an optional version, or no version for `Indexer` to automatically create the object at the next version.   This ensures objects always have a valid version number.

The next most important work is in the `Indexor` service, which should handle everything that was removed from `LocalDataService`.

I added three DAOs to represent basic `IndexEntry`, `Record`, and `CalculationParameters` objects.  The particular expressions of these in the different workflows were made to inherit from the basic objects.  This makes `Indexer` flexible and able to accept additional features.

All three of these further derive from `VersionedObject`, which is a DAO storing the version as a property,  that has been protected as much as is possible within python, to prevent the version being reassigned an invalid value.

Devs should **NEVER** create `IndexEntry`s, `Record`s, or `CalculationParameter`s objects, or their children, directly.  **ALWAYS** create them through `Indexer`.  In the future, this should be enforced by overriding the constructor for `VersionedObject` so that only an `Indexer` object can build one. 

When a GroceryListItem corresponds to output from a previous version, the builder requries a version.  This can be supplied by the `Indexor` to get the most recent version (i.e. `thisOrCurrentVersion()`).

I removed the idea of `"*"` as a version number.  This had emerged briefly as a neat way of requesting the most recent version, but with the `Indexor` is unnecessary and could lead to confusion.

Internally, a version must **always** be an integer.  Errors are represented as `None`, and the default state is assigned the special value `-1`.  Externally, errors and default state will write out with a special string specified in the `application.yml`.  This string could be a number, such as having the default version at 0.

### Dev testing

Please review the code and associated unit tests carefully.  I highly suggest pulling the branch and pulling up `LocalDataService` in one window and `test_LocalDataService` in another and going through and making sure the tests make sense and do what they need to.

I highly suggest doing the same with `Indexer` and `test_Indexer`: pull them up side-by-side and go through them.

This is going to require having a custom dev/Calibration folder that you can write to, preferably the one constructed from the tarball.

First, clear the state folder so that the state is unitialized.  Launch SNAPRed, and start with Calibration.
1. Initialize the state.  When it is finished:
   - examine the "CalibrationIndex.json" file.  It should have an entry for the default version.  This is used internally to provide properties when reading StateConfig.  Make sure the default version is NOT `-1`.
   - Make sure there is a directory `v_0000`.  Inside there is a "CalibrationParameters.json", a "CalibrationRecord.json", and a default diffcal table.
   -  Open the "CalibrationParameters.json" file.  It should have all the parts, and its version should be `0`.
   - Open the file "CalibrationRecord.json" file.  It should  have runNumber as "default" and version `0`.
2. Continue until you get peaks, then continue to assessment.
3. Check the assessment dropdown.  There should be **nothing** there.  The default version should *NOT* be showing here.
4. Go to save.  Try to save the version as "matt".  Try to save version as -1.  Delete the version and continue.
5. It should save as `v_0001`.  There should be **NO** warnings.
   - examine the "CalibrationIndex.json" file.  It should now have an additional entry for version 1.
   - Examine the "CalibrationRecord.json" file for `v_0001`.  It should look like a legit calibration record,  Make sure the "version" field is `1`.   The "calculationParameters" field should also have version `1`.  The workspaces should have versions in their names, and their names correspond to the saved workspaces.
   - Open the "CalibrationParameters.json" file.  It should have all the parts, and its version should be `1`.
6. Go back and run calibration again.  At assessment screen, load `v0001`.  Make sure it loads.
7. Run at least one iteration.
7. At save, enter the version as a random number between 2 and 120.  Make sure it saves at that version, and do the same checks as for version `1`, but looking for the new version number.

Now move to Normalization
1. There is no default, so go straight to loading.  Use the same run number for background to speed it up.
2. Click past peak view.
3. At save, try version "matt", try version "-1", delete version and save.
4. Check the "NormalizationIndex.json" file.  It should have an entry for version `1`.
5. Check that the dir `v_0001` exists and has "NormalizationParameters.json", "NormalizationRecord.json", and three workspaces.
6. On the json files, do the same checks as for Calibration.

Run Reduction.  It should load the latest calibration and normalization with no warnings.

**BONUS POINTS**
Edit your `application.yml` to the following:
``` yaml
version:
  friendlyName:
    error: "not so good"
    default: "X"
  start: 0
```
And then clear the state folder and run the diffcal checks again.  
This time the default should save with version `X` and the version numbering should start at `0`.

### CIS testing

This is an enabler.  But the CIS can test by following similar steps as above.

## Link to EWM item

[EWM#5612](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5612)


src/snapred/backend/data/LocalDataService copy.py
walshmm, line 1: should this be here?

src/snapred/backend/data/DataFactoryService.py
```diff
@@ -60,30 +60,11 @@ def getSampleFilePaths(self):
...
-        else:
-            # lookup and package data
-            reductionState = ReductionState(
-                instrumentConfig=self.getInstrumentConfig(runId),
-                stateConfig=self.getStateConfig(runId, useLiteMode),
-            )
-            self.cache[runId] = reductionState
-
-        return reductionState
-
```
rboston628, line 81: This was not deleted, it was moved to below with the other reduction methods.  I'm not sure how we got two reduction methods sections.

tests/unit/backend/data/test_DataFactoryService.py
```diff
@@ -88,19 +105,11 @@ def test_constructStateId(self):
         actual = self.instance.constructStateId(arg)
         assert actual == self.expected(arg)
 
-    def test_getCalibrationState(self):
-        actual = self.instance.getCalibrationState("123", False)
-        assert actual == self.expected("123", False)
-
```
rboston628, line 94: Was only moved to below

tests/unit/backend/data/test_DataFactoryService.py
```diff
@@ -88,19 +105,11 @@ def test_constructStateId(self):
...
     def test_getGroupingMap(self):
         arg = mock.Mock()
         actual = self.instance.getGroupingMap(arg)
         assert actual == self.expected(arg)
 
-    def test_checkCalibrationStateExists(self):
-        actual = self.instance.checkCalibrationStateExists("123")
-        assert actual == self.expected("123")
-
```
rboston628, line 103: Was only moved to below

src/snapred/backend/data/LocalDataService.py
```diff
@@ -257,34 +244,6 @@ def _constructNormalizationStatePath(self, stateId, useLiteMode) -> Path:
...
-        return self._appendVersion(self._constructNormalizationStatePath(stateId, useLiteMode), version)
-
-    @validate_arguments
-    def _constructCalibrationParametersFilePath(self, runId: str, useLiteMode: bool, version: Version) -> Path:
-        return self._constructCalibrationDataPath(runId, useLiteMode, version) / "CalibrationParameters.json"
-
-    @validate_arguments
-    def _constructNormalizationParametersFilePath(self, runId: str, useLiteMode: bool, version: Version) -> Path:
-        return self._constructNormalizationDataPath(runId, useLiteMode, version) / "NormalizationParameters.json"
-
```
rboston628, line 287: this is delegated to the indexor

src/snapred/backend/data/LocalDataService.py
```diff
@@ -301,222 +260,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
-        }
-        return expressions[symbol](int(runNumber1), int(runNumber2))
-
-    def _isApplicableEntry(self, calibrationIndexEntry, runId):
-        """
-        Checks to see if an entry in the calibration index applies to a given run id via numerical comparison.
-        """
-
-        symbol, runNumber = self._parseAppliesTo(calibrationIndexEntry.appliesTo)
-        return self._compareRunNumbers(runId, runNumber, symbol)
```
rboston628, line 362: These methods were delegated to the Indexor

src/snapred/backend/data/LocalDataService.py
```diff
@@ -301,222 +260,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
-        write_model_list_pretty(calibrationIndex, indexPath)
-
-    def writeNormalizationIndexEntry(self, entry: NormalizationIndexEntry):
-        stateId, _ = self._generateStateId(entry.runNumber)
-        normalizationPath: Path = self._constructNormalizationStatePath(stateId, entry.useLiteMode)
-        indexPath: Path = normalizationPath / "NormalizationIndex.json"
-        # append to index and write to file
-        normalizationIndex = self.readNormalizationIndex(entry.runNumber, entry.useLiteMode)
-        normalizationIndex.append(entry)
-        write_model_list_pretty(normalizationIndex, indexPath)
```
rboston628, line 421: These methods are delegated to the Indexor

src/snapred/backend/dao/calibration/Calibration.py
rboston628, line 1: This now inherits from the `CalculationParameters` object which supplies most of its fields and it s validators

tests/unit/backend/data/test_DataFactoryService.py
rboston628, line 1: Note that some of the tests were rearranged, either to be logically arranged or to better match their order in the service

tests/util_tests/test_state_helpers.py
```diff
@@ -90,12 +90,17 @@ def initPVFileMock() -> mock.Mock:
...
 @mock.patch.object(LocalDataService, "_defaultGroupingMapPath")
 @mock.patch.object(LocalDataService, "readInstrumentConfig")
 @mock.patch.object(LocalDataService, "_readPVFile")
 def test_state_root_override_enter(
-    mockReadPVFile, mockReadInstrumentConfig, mockDefaultGroupingMapPath, mockGenerateStateId
+    mockReadPVFile,
+    mockReadInstrumentConfig,
+    mockDefaultGroupingMapPath,
+    mockGenerateStateId,
+    mockWriteDiffCalTable,  # noqa ARG001
```
rboston628, line 103: I believe this was actually calculating the diffcal tabels for both resolutions as part of the test

tests/unit/backend/data/test_GroceryService.py
rboston628, line 1: These changes, including the `self.diffCalOutputName` lines, are all to facilitate version numbers on diffcal output requests.

src/snapred/backend/dao/calibration/Calibration.py
rboston628, line 1: This now inherits most of its fields from `CalculationParameters`

src/snapred/backend/dao/normalization/Normalization.py
rboston628, line 1: This now inherits most of its fields from `CalculationParameters`

src/snapred/backend/dao/calibration/CalibrationIndexEntry.py
rboston628, line 1: This now inherits most of its fields from `IndexEntry`

src/snapred/backend/dao/normalization/NormalizationIndexEntry.py
rboston628, line 1: This now inherits most of its fields from `IndexEntry`

src/snapred/backend/dao/calibration/CalibrationRecord.py
rboston628, line 1: This now inherits most of its functionality from `Record`

src/snapred/backend/dao/indexing/Versioning.py
rboston628, line 1: Should this go here?  Or in meta?

src/snapred/backend/dao/normalization/NormalizationRecord.py
rboston628, line 1: This now inherits most of its functionality from `Record`

src/snapred/backend/dao/request/CalibrationAssessmentRequest.py
rboston628, line 1: Previously, the "lists" of `WorkspaceName`s were expected to hold only a single value; with the exception of the entry for the output workspaces, which had two members, one for DSP and one for TOF.  That is no longer used, so this can be changed to a single value.

src/snapred/backend/service/CalibrationService.py
```diff
@@ -196,29 +197,82 @@ def fitPeaks(self, request: FitMultiplePeaksRequest):
...
-        calibrationRecord = self.dataExportService.exportCalibrationRecord(calibrationRecord)
-        calibrationRecord = self.dataExportService.exportCalibrationWorkspaces(calibrationRecord)
-        entry.version = calibrationRecord.version
-        self.saveCalibrationToIndex(entry)
+        record = request.calibrationRecord
+        version = self.dataFactoryService.getThisOrNextCalibrationVersion(
+            record.runNumber,
+            record.useLiteMode,
+            request.version,
+        )
```
rboston628, line 209: Use wither the version from the request, or use the Indexor to find the next version.  Needed to rename the workspaces.

src/snapred/backend/service/CalibrationService.py
```diff
@@ -196,29 +197,82 @@ def fitPeaks(self, request: FitMultiplePeaksRequest):
...
+                case wngt.DIFFCAL_MASK:
+                    savedWorkspaces[key] = wng.diffCalMask().runNumber(record.runNumber).version(version).build()
+                case wngt.DIFFCAL_TABLE:
+                    savedWorkspaces[key] = wng.diffCalTable().runNumber(record.runNumber).version(version).build()
+                case _:
+                    raise RuntimeError(f"Unexpected output type {key} for {wsName}")
+
+            self.groceryService.renameWorkspace(wsName, savedWorkspaces[key])
+
+        record.workspaces = savedWorkspaces
```
rboston628, line 250: This was previously taking place inside the LocalDataService, and required being called twice with nearly duplicated code.  Moving it here makes more sense, and also prevents the duplicate renaming code.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -69,16 +67,18 @@ def _createFileNotFoundError(msg, filename):
...
-    # starting version number -- the first run printed
-    VERSION_START = Config["instrument.startingVersionNumber"]
     # conversion factor from microsecond/Angstrom to meters
     CONVERSION_FACTOR = Config["constants.m2cm"] * PhysicalConstants.h / PhysicalConstants.NeutronMass
 
     def __init__(self) -> None:
         self.verifyPaths = Config["localdataservice.config.verifypaths"]
         self.instrumentConfig = self.readInstrumentConfig()
         self.mantidSnapper = MantidSnapper(None, "Utensils")
+        self._rolodex = {}
```
rboston628, line 79: This is a cache of `Indexor` objects which are accessed using the `indexor` method.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -229,25 +224,8 @@ def _findMatchingFileList(self, pattern, throws=True) -> List[str]:
...
-        Find all directories matching a glob pattern.
-        Optional: throws exception if nothing found.
-        """
-        fileList: List[str] = []
-        for fname in glob.glob(pattern, recursive=True):
-            if os.path.isdir(fname):
-                fileList.append(fname)
-        if len(fileList) == 0 and throws:
-            raise ValueError(f"No directories could be found with pattern: {pattern}")
-        return fileList
```
rboston628, line 243: This was used to look for versions from filenames, and is no longer needed.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -229,25 +224,8 @@ def _findMatchingFileList(self, pattern, throws=True) -> List[str]:
...
-
     ##### PATH METHODS #####
 
-    def _appendVersion(self, root: Path, version: Optional[Version]) -> Path:
-        # Append a version directory to a data path
-        return root / wnvf.fileVersion(version)
```
rboston628, line 249: This facilitated creating paths to versioned objects, and is no longer needed

src/snapred/backend/data/LocalDataService.py
```diff
@@ -259,34 +237,6 @@ def _constructNormalizationStatePath(self, stateId, useLiteMode) -> Path:
...
-        return self._appendVersion(self._constructNormalizationStatePath(stateId, useLiteMode), version)
-
-    @validate_call
-    def _constructCalibrationParametersFilePath(self, runId: str, useLiteMode: bool, version: Version) -> Path:
-        return self._constructCalibrationDataPath(runId, useLiteMode, version) / "CalibrationParameters.json"
-
-    @validate_call
-    def _constructNormalizationParametersFilePath(self, runId: str, useLiteMode: bool, version: Version) -> Path:
-        return self._constructNormalizationDataPath(runId, useLiteMode, version) / "NormalizationParameters.json"
-
```
rboston628, line 289: All of these methods, which were only used inside the `LocalDataService`, have been replaced by calls to `Indexor` methods.  For instance, `_constructNormalizationDataPath(runId, useLiteMode, version)` can be replaced with `normalizationIndexor(runId, useLiteMode).versionPath(version)`.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -303,230 +253,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
         return reductionStateRoot / mode / runNumber
 
-    @validate_call
-    def _constructReductionDataPath(self, runNumber: str, useLiteMode: bool, version: Version) -> Path:
-        return self._appendVersion(self._constructReductionDataRoot(runNumber, useLiteMode), version)
-
-    @validate_call
-    def _constructReductionRecordFilePath(self, runNumber: str, useLiteMode: bool, version: Version) -> Path:
-        recordPath = self._constructReductionDataPath(runNumber, useLiteMode, version) / "ReductionRecord.json"
-        return recordPath
```
rboston628, line 313: As above for the other path methods.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -303,230 +253,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
+                path = self._constructCalibrationStatePath(stateId, useLiteMode)
+            case IndexorType.NORMALIZATION:
+                path = self._constructNormalizationStatePath(stateId, useLiteMode)
+            case IndexorType.REDUCTION:
+                # NOTE the "stateId" should actually be a runNumber for reduction
+                path = self._constructReductionDataRoot(stateId, useLiteMode)
+            case _:
+                mode = "lite" if useLiteMode else "native"
+                path = self._constructCalibrationStateRoot(stateId) / mode
+        return path
```
rboston628, line 281: An `Indexor` must be initialized with a directory, and it will manage the files inside that directory.  This method produces the needed directory for calibration, normalization, and reduction `Indexor`s.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -303,230 +253,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
+        # NOTE the reduction state path is determined by run number, not state id
+        if indexorType is IndexorType.REDUCTION:
+            stateId = runNumber
+        else:
+            stateId, _ = self._generateStateId(runNumber)
+        key = (stateId, useLiteMode, indexorType)
+        if self._rolodex.get(key) is None:
+            path = self._statePathForWorkflow(*key)
+            self._rolodex[key] = Indexor(indexorType=indexorType, directory=path)
+        return self._rolodex[key]
```
rboston628, line 293: This checks the rolodex for the correct `Indexor` corresponding to the run number, resolution, and workflow.  If none found, then it creates that `Indexor`.  The created `Indexor` will automatically read the files in its directory, or if none present create those files.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -303,230 +253,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
-
-    def _isApplicableEntry(self, calibrationIndexEntry, runId):
-        """
-        Checks to see if an entry in the calibration index applies to a given run id via numerical comparison.
-        """
+    def normalizationIndexor(self, runId: str, useLiteMode: bool):
+        return self.indexor(runId, useLiteMode, IndexorType.NORMALIZATION)
 
-        symbol, runNumber = self._parseAppliesTo(calibrationIndexEntry.appliesTo)
-        return self._compareRunNumbers(runId, runNumber, symbol)
```
rboston628, line 368: All this logic was moved to `Indexor`

src/snapred/backend/data/LocalDataService.py
```diff
@@ -303,230 +253,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
-            # filter for latest applicable
-            relevantEntries = list(filter(lambda x: self._isApplicableEntry(x, runId), normalizationIndex))
-            if len(relevantEntries) < 1:
-                return None
-            latestNormalization = relevantEntries[-1]
-            version = latestNormalization.version
-        return version
-
-    def getVersionFromNormalizationIndex(self, runId: str, useLiteMode: bool) -> int:
-        return self._getVersionFromNormalizationIndex(runId, useLiteMode)
```
rboston628, line 412: These methods can be replaced with `indexor(etc.).latestApplicableVersion(runNumber)`

src/snapred/backend/data/LocalDataService.py
```diff
@@ -303,230 +253,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
-
-    def writeCalibrationIndexEntry(self, entry: CalibrationIndexEntry):
-        stateId, _ = self._generateStateId(entry.runNumber)
-        calibrationPath: Path = self._constructCalibrationStatePath(stateId, entry.useLiteMode)
-        indexPath: Path = calibrationPath / "CalibrationIndex.json"
-        # append to index and write to file
-        calibrationIndex = self.readCalibrationIndex(entry.runNumber, entry.useLiteMode)
-        calibrationIndex.append(entry)
-        write_model_list_pretty(calibrationIndex, indexPath)
```
rboston628, line 421: This method was not deleted but moved.  It not only calls the `Indexor` method.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -303,230 +253,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
-
-    def writeNormalizationIndexEntry(self, entry: NormalizationIndexEntry):
-        stateId, _ = self._generateStateId(entry.runNumber)
-        normalizationPath: Path = self._constructNormalizationStatePath(stateId, entry.useLiteMode)
-        indexPath: Path = normalizationPath / "NormalizationIndex.json"
-        # append to index and write to file
-        normalizationIndex = self.readNormalizationIndex(entry.runNumber, entry.useLiteMode)
-        normalizationIndex.append(entry)
-        write_model_list_pretty(normalizationIndex, indexPath)
```
rboston628, line 430: These methods were simply moved, and their logic simplified to only call the `Indexor` methods.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -303,230 +253,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
-        """
-        return self._getLatestVersionNumber(self._constructNormalizationStatePath(stateId, useLiteMode))
+    def writeNormalizationIndexEntry(self, entry: NormalizationIndexEntry, version: Optional[int] = None):
+        self.normalizationIndexor(entry.runNumber, entry.useLiteMode).addIndexEntry(entry, version)
 
-    def _getLatestReductionVersionNumber(self, runNumber: str, useLiteMode: bool) -> int:
-        """
-        Get the version number of the latest set of reduction data files.
-        """
-        return self._getLatestVersionNumber(self._constructReductionDataRoot(runNumber, useLiteMode))
```
rboston628, line 500: These methods handled extracting version numbers from the directory tree, and are no longer needed.  Similar logic is replicated inside `Indexor`

src/snapred/backend/data/LocalDataService.py
```diff
@@ -303,230 +253,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
-        normalizationIndex.append(entry)
-        write_model_list_pretty(normalizationIndex, indexPath)
-
-    @validate_call
-    def getCalibrationRecordFilePath(self, runId: str, useLiteMode: bool, version: Version) -> Path:
-        return self._constructCalibrationDataPath(runId, useLiteMode, version) / "CalibrationRecord.json"
-
-    @validate_call
-    def getNormalizationRecordFilePath(self, runId: str, useLiteMode: bool, version: Version) -> Path:
-        return self._constructNormalizationDataPath(runId, useLiteMode, version) / "NormalizationRecord.json"
```
rboston628, line 438: These were not called anywhere, and can be replaced with `indexor(etc).recordPath(version)`.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -303,230 +253,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
-            except ValidationError as e:
-                logger.error(f"Error parsing {latestFile}: {e}")
-                raise OutdatedDataSchemaError(f"It looks like the data schema for {latestFile} is outdated.") from e
-
-        return record
-
-    @validate_call
-    def _getCurrentNormalizationRecord(self, runId: str, useLiteMode: bool):
-        version = self._getVersionFromNormalizationIndex(runId, useLiteMode)
-        return self.readNormalizationRecord(runId, useLiteMode, version)
```
rboston628, line 529: It already returns the current if no version is passed.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -535,292 +340,155 @@ def writeNormalizationRecord(
...
-        savedWorkspaces[wngt.DIFFCAL_TABLE] = wss
-        wss = []
-        for wsName in workspaces.pop(wngt.DIFFCAL_MASK, []):
-            # Rebuild the workspace name to strip any "iteration" number:
-            #   * WARNING: this workaround does not work correctly if there are multiple mask workspaces.
-            ws = wng.diffCalMask().runNumber(record.runNumber).version(record.version).build()
-            wss.append(ws)
-        savedWorkspaces[wngt.DIFFCAL_MASK] = wss
-        savedRecord = deepcopy(record)
-        savedRecord.workspaces = savedWorkspaces
```
rboston628, line 689: All of this logic was being duplicated, first for saving records and secondly for saving workspaces.  It was moved to a single space, inside the `CalibrationService`.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -535,292 +340,155 @@ def writeNormalizationRecord(
...
-            #   * WARNING: this workaround does not work correctly if there are multiple table workspaces.
-            diffCalFilename = Path(
-                wng.diffCalTable().runNumber(record.runNumber).version(record.version).build() + ".h5"
-            )
-            self.writeDiffCalWorkspaces(
-                calibrationDataPath,
-                diffCalFilename,
-                tableWorkspaceName=tableWSName,
-                maskWorkspaceName=maskWSName,
-            )
```
rboston628, line 757: As above, this rename logic was moved to the `CalibrationService`

src/snapred/backend/data/LocalDataService.py
```diff
@@ -849,159 +517,40 @@ def readReductionData(self, runNumber: str, useLiteMode: bool, version: int) ->
...
-        filePath = os.path.join(samplePath, fileName)
-        if not os.path.exists(filePath):
-            raise ValueError(f"the file '{filePath}' does not exist")
-        with open(filePath, "r") as f:
-            calibrantSampleDict = json.load(f)
-        filePath = Path(calibrantSampleDict["crystallography"]["cifFile"])
-        # Allow relative paths:
-        if not filePath.is_absolute():
-            filePath = Path(Config["samples.home"]).joinpath(filePath)
-        return str(filePath)
```
rboston628, line 904: These were not edited, this section was simply moved to below the read / write state methods.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -1081,7 +632,22 @@ def initializeState(self, runId: str, useLiteMode: bool, name: str = None):
...
+                calculationParameters=calibration,
+            )
+            entry = IndexEntry(
+                runNumber="default",
+                useLiteMode=liteMode,
+                version=version,
+                appliesTo=">=0",
+                author="SNAPRed Internal",
+                comments="The default condition for loading StateConfigs if none other found",
+                timestamp=round(time.time()),
```
rboston628, line 650: The `Record` and `IndexEntry` are necessary for the `Indexor` to find the "CalibrationParameters.json" file.

tests/unit/backend/service/test_CalibrationService.py
rboston628, line 1: Most of these changes are just accommodating that the saved workspaces are no longer stored in lists.

tests/unit/backend/data/test_LocalDataService.py
```diff
@@ -91,13 +101,6 @@ def test_fileExists_no():
         assert not LocalDataService().fileExists(nonexistent)
 
 
-def _readInstrumentParameters():
-    instrumentParameters = None
-    with Resource.open("inputs/SNAPInstPrm.json", "r") as file:
-        instrumentParameters = json.loads(file.read())
-    return instrumentParameters
```
rboston628, line 98: Moved below

tests/unit/backend/data/test_LocalDataService.py
```diff
@@ -116,6 +119,36 @@ def test_readInstrumentParameters():
...
+    """This verifies that a broken configuration (from production) can't find all of the files"""
+    # get a handle on the service
+    service = LocalDataService()
+    service.verifyPaths = True  # override test setting
+    prevInstrumentConfig = Config._config["instrument"]["config"]
+    Config._config["instrument"]["config"] = "/this/path/does/not/exist"
+    with pytest.raises(FileNotFoundError):
+        service.readInstrumentConfig()
+    Config._config["instrument"]["config"] = prevInstrumentConfig
+    service.verifyPaths = False  # put the setting back
```
rboston628, line 149: these were moved from somewhere else

tests/unit/backend/data/test_LocalDataService.py
```diff
@@ -128,321 +161,124 @@ def getMockInstrumentConfig():
...
-            assert focusGroup == defaultGroupingMap.liteFocusGroups[n]
-            if not Path(focusGroup.definition).is_absolute():
-                relativePathCount += 1
-        for n, focusGroup in enumerate(groupingMap.nativeFocusGroups):
-            assert focusGroup == defaultGroupingMap.nativeFocusGroups[n]
-            if not Path(focusGroup.definition).is_absolute():
-                relativePathCount += 1
-        assert relativePathCount > 0
-    finally:
-        Config._config["instrument"]["calibration"]["powder"]["grouping"]["home"] = savePath
```
rboston628, line 399: All of these were moved to near the end of the file, to correspond to the location of the methods they test.

tests/unit/backend/data/test_LocalDataService.py
rboston628, line 1: I had started adding comments, but it is really too much of a confusing mess.  Please look at `LocalDataService` and this test side-by-side to make sure the tests make sense.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -908,100 +576,35 @@ def readCifFilePath(self, sampleId: str):
...
-        if normalizationParametersFilePath.exists():
-            logger.warning(f"overwriting normalization parameters at {normalizationParametersFilePath}")
-        normalization.version = int(version)
-        normalizationDataPath: Path = self._constructNormalizationDataPath(
-            normalization.seedRun,
-            normalization.useLiteMode,
-            version,
-        )
-        if not normalizationDataPath.exists():
-            os.makedirs(normalizationDataPath)
```
rboston628, line 1003: All of this logic is handled inside the `Indexor`

src/snapred/backend/data/LocalDataService.py
```diff
@@ -535,292 +340,155 @@ def writeNormalizationRecord(
...
+    def writeReductionRecord(self, record: ReductionRecord, version: Optional[int] = None):
         """
         Persists a `ReductionRecord` to either a new version folder, or overwrites a specific version.
         * side effect: updates version numbers of incoming `ReductionRecord`;
-        * must be called before any call to `writeReductionData`.
+        * side effect: updates version numbers of the `calculationParameters` on the record;
+        * side effect: creates the version directory if none exists;
         """
-        # For the moment, a single run number is assumed:
-        runNumber = record.runNumbers[0]
```
rboston628, line 782: `record.runNumber = record.runNumbers[0]` is automatically performed on reduction records

src/snapred/backend/data/LocalDataService.py
```diff
@@ -535,292 +340,155 @@ def writeNormalizationRecord(
...
-        # There seems no need to write the _nested_ Normalization,
-        # because it's written to a separate file during 'writeNormalizationState'.
-        # However, if it is going to be _nested_, this marks it with the correct version.
-        # (For example, use pydantic Field(exclude=True) to _stop_ nesting it.)
-        record.calibration.version = version
-
-        normalizationPath: Path = self._constructNormalizationDataPath(runNumber, record.useLiteMode, version)
-        # check if directory exists for runId
-        if not normalizationPath.exists():
-            os.makedirs(normalizationPath)
```
rboston628, line 555: All of this is now handled by the `Indexor`

src/snapred/meta/mantid/WorkspaceNameGenerator.py
```diff
@@ -86,7 +85,9 @@ def formatRunNumber(runNumber: str):
     def formatVersion(version: Optional[Version], use_v_prefix: vPrefix = vPrefix.WORKSPACE):
         if version is None:
             return ""
-        if not version == "*":
+        if version != VERSION_DEFAULT:
+            version = str(version).zfill(4)
+        elif isinstance(VERSION_DEFAULT, int):
```
rboston628, line 90: No more globs, but there exists a default version which is where the default diffcal table and the initial calibration parameters file is written.  We have been using `v_0000` as the default, but users might want to call this something else, like `default`, and use `v_0000` as the starting version number.

tests/unit/backend/data/test_LocalDataService.py
```diff
@@ -697,394 +767,421 @@ def test__findMatchingFileList():
...
+# def test_writeCalibrationIndexEntry():
+#     pass
+
+
+# def test_readCalibrationIndexExisting():
+#     pass
+
+
+# def test_readNormalizationIndexExisting():
+#     pass
```
dlcaballero16, line 1032: Delete commented code

tests/unit/backend/data/test_LocalDataService.py
```diff
@@ -697,394 +767,421 @@ def test__findMatchingFileList():
...
-        wngt.DIFFCAL_TABLE: ["_diffract_consts_057514_v0001"],
-        wngt.DIFFCAL_MASK: ["_diffract_consts_mask_057514_v0001"],
-    }
-    with pytest.raises(  # noqa: PT012
-        RuntimeError,
-        match=f"cannot save a workspace-type: {wngt.DIFFCAL_OUTPUT} without a units token in its name",
-    ):
-        localDataService.writeCalibrationWorkspaces(testCalibrationRecord)
+# def test_writeCalibrationWorkspaces_no_units():
+#     pass
```
dlcaballero16, line 1133: delete commented code

tests/unit/backend/data/test_LocalDataService.py
```diff
@@ -1333,57 +1434,17 @@ def test_writeReductionData(createReductionWorkspaces):
...
-
-        # `writeReductionRecord` must be called first
-        # * deliberately _not_ done in this test => <reduction-data root> directory won't exist
-        with pytest.raises(RuntimeError) as einfo:
-            localDataService.writeReductionData(testRecord)
-        msg = str(einfo.value)
-    assert "reduction version directories" in msg
-    assert "do not exist" in msg
+# def test_writeReductionData_no_directories():
+#     pass
```
dlcaballero16, line 1447: delete commented code

tests/unit/backend/data/test_LocalDataService.py
```diff
@@ -1512,341 +1575,146 @@ def test__constructReductionDataFilePath():
...
-    testVersion = randint(1, 20)
-    localDataService = LocalDataService()
-    localDataService._generateStateId = mock.Mock()
-    localDataService._generateStateId.return_value = ("123", "456")
-    localDataService._constructCalibrationStatePath = mock.Mock()
-    localDataService._constructCalibrationStatePath.return_value = Path(Resource.getPath("outputs"))
-    actualPath = localDataService.getCalibrationRecordFilePath("57514", True, testVersion)
-    assert actualPath == Path(Resource.getPath("outputs")) / wnvf.fileVersion(testVersion) / "CalibrationRecord.json"
+# define these here to help the git diff not get confused
+# most of these tests are for deleted methods
```
dlcaballero16, line 1585: not sure if this is still needed, but also a lot of commented code below

dlcaballero16 (approved):
Was able to test and get expected results.

src/snapred/backend/data/LocalDataService.py
ekapadi: Good.  **So long as indexer is not called within these write methods!**

src/snapred/backend/data/LocalDataService.py
```diff
@@ -535,292 +340,155 @@ def writeNormalizationRecord(
...
         """
-        normalizationDataPath: Path = self._constructNormalizationDataPath(
-            record.runNumber, record.useLiteMode, record.version
-        )
+        indexor = self.normalizationIndexor(record.runNumber, record.useLiteMode)
+        version = indexor.thisOrNextVersion(version)
+        normalizationDataPath: Path = indexor.versionPath(version)
         for workspace in record.workspaceNames:
-            filename = workspace + "_" + wnvf.formatVersion(record.version)
+            filename = workspace + "_" + wnvf.formatVersion(version)
```
ekapadi, line 362: All of this crap needs to be moved _back_ into the "save" method that calls this method.

src/snapred/backend/data/LocalDataService.py
ekapadi: Good.  However, please double check that `Normalization` service is also doing this?  (Possibly I was misunderstanding some of the previous changes...  :)  )

src/snapred/backend/data/LocalDataService.py
ekapadi: There still needs to be a `writeWorkspaces` method.  The `CalibrationService` shouldn't need to figure out how to save the workspaces -- that's still `LocalDataService`'s responsibility.  However, it _does_ need to set the workspace names so that they are correct (both in the record, and possibly also on the ADS-resident workspaces themselves).

src/snapred/backend/data/LocalDataService.py
```diff
@@ -535,292 +340,155 @@ def writeNormalizationRecord(
...
+        ext = Config["calibration.diffraction.diagnostic.extension"]
+        filename = Path(wsName + ext)
+        self.writeWorkspace(calibrationDataPath, filename, wsName)
+
+        # write the diffcal table and attached masl
+        tableWSName = record.workspaces[wngt.DIFFCAL_TABLE]
+        maskWSName = record.workspaces[wngt.DIFFCAL_MASK]
+        ext = ".h5"
+        diffCalFilename = Path(tableWSName + ext)
+        self.writeDiffCalWorkspaces(
```
ekapadi, line 425: Sorry!!!  I guess it is still in here.

src/snapred/backend/data/LocalDataService.py
ekapadi: Was it added as a `@property`?  (It possibly should be.  A `PrivateAttribute` or `Optional` attribute probably _not_.)

ekapadi (changes requested):
ongoing...

src/snapred/backend/data/LocalDataService.py
rboston628: The reason was because the method arguments are different  from the key the indexor should associate with.  However, I think I can do this with two separate methods.

src/snapred/backend/dao/request/CalibrationAssessmentRequest.py
```diff
@@ -27,7 +27,7 @@ class CalibrationAssessmentRequest(BaseModel):
     useLiteMode: bool
     focusGroup: FocusGroup
     calibrantSamplePath: str
-    workspaces: Dict[WorkspaceType, List[WorkspaceName]]
+    workspaces: Dict[WorkspaceType, WorkspaceName]
```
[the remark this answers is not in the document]
rboston628: Unfortunately I had to touch this part of the code.  I remember we had discussed removing the nested workspace lists, so since I was here I did that.  I can revert it if it's desired.

src/snapred/backend/data/LocalDataService.py
rboston628: Yes, `groupingMap` doesn't make reference to this method

src/snapred/backend/service/ReductionService.py
```diff
@@ -179,19 +177,28 @@ def fetchReductionGroceries(self, request: ReductionRequest) -> Dict[str, Any]:
         """
         # gather input workspace and the diffcal table
         self.groceryClerk.name("inputWorkspace").neutron(request.runNumber).useLiteMode(request.useLiteMode).add()
-        self.groceryClerk.name("diffcalWorkspace").diffcal_table(request.runNumber).useLiteMode(
+        self.groceryClerk.name("diffcalWorkspace").diffcal_table(request.runNumber, request.version).useLiteMode(
             request.useLiteMode
         ).add()
-        self.groceryClerk.name("normalizationWorkspace").normalization(request.runNumber).useLiteMode(
+        self.groceryClerk.name("normalizationWorkspace").normalization(request.runNumber, request.version).useLiteMode(
```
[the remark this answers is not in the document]
rboston628: the request object only has a single `version` in it.  I agree, it *should* have two, but I'm not going to do that in this PR

src/snapred/backend/dao/normalization/NormalizationRecord.py
```diff
@@ -17,25 +17,24 @@ class NormalizationRecord(BaseModel):
...
+    # specific to normalization records
     backgroundRunNumber: str
     smoothingParameter: float
     peakIntensityThreshold: float
     # detectorPeaks: List[DetectorPeak] # TODO: need to save this for reference during reduction
-    calibration: Calibration
     workspaceNames: List[WorkspaceName] = []
-    version: int = Config["instrument.startingVersionNumber"]
+
     dMin: float
```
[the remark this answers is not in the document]
rboston628: `calibration` is still here, it has jsut been renamed to `calculationParameters`.  The calibration used during normalization can be accessed by its version through `calibrationVersionUsed`

src/snapred/backend/service/NormalizationService.py
```diff
@@ -161,37 +164,42 @@ def normalizationAssessment(self, request: NormalizationRequest):
...
-        normalizationRecord = self.dataExportService.exportNormalizationRecord(normalizationRecord)
-        normalizationRecord = self.dataExportService.exportNormalizationWorkspaces(normalizationRecord)
-        entry.version = normalizationRecord.version
-        self.saveNormalizationToIndex(entry)
+        record = request.normalizationRecord
+        # NOTE the Indexor will handle the version information for us,
+        # but the workspaces must be saved before both record and index are saved
+        self.dataExportService.exportNormalizationRecord(record, version)
+        self.dataExportService.exportNormalizationWorkspaces(record, version)
+        self.saveNormalizationToIndex(entry, version)
```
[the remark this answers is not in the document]
rboston628: Version is not set here, on both record and workspaces

src/snapred/resources/application.yml
```diff
@@ -137,6 +136,11 @@ logging:
 samples:
   home: ${instrument.calibration.sample.home}
 
+version:
+  error: "uninitialized"  # alphanumeric
+  default: 0 # alphanumeric
```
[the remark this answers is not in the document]
rboston628: Version is always an internally represented as an `int`.  The user can choose to have it represented by a string through this setting.  Internally, the default version is `-1`.

If a version wishes, this could be set to `default: "default"`, for instance, and the default version will be named "default" instead of `0`.

src/snapred/backend/dao/indexing/Record.py
```diff
@@ -0,0 +1,64 @@
...
+    The special Nonrecord object should be used in instances where a record is expected but none exists.
+    Use this in place of None.
+
+    """
+
+    runNumber: str
+    useLiteMode: bool
+    version: Version = UNINITIALIZED
+    calculationParameters: CalculationParameters | Literal["No calculation parameters"]
+
```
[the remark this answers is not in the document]
rboston628: The conflict was with making `Nonrecord`.  I have replaced this with `Optional` *plus* a validator that prevents it from being set to `None`, except in the case of the `Nonrecord`.

tests/resources/application.yml
```diff
@@ -160,6 +159,11 @@ test:
 
 cis_mode: false
 
+version:
+  error: "uninitialized"
+  default: "0xF"  # this tests alphanumeric defaults
```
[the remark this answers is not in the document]
rboston628: It's something Malcolm had hinted at, so I just did it now, rather than re-implementing everything later.  As implemented now, it does not impact internal management of versions, where the default is the integer `-1`.  We don't have to tell anyone this is supported, but it is.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -303,230 +253,85 @@ def _constructReductionDataRoot(self, runNumber: str, useLiteMode: bool) -> Path
...
-        This doesn't need to be its own function,
-        but it represents a common pattern in code
-        """
-        if otherThings is None:
-            return max(things, default=self.VERSION_START)
+        return self.normalizationIndexor(runId, useLiteMode).getIndex()
+
+    def readReductionIndex(self, runId: str, useLiteMode: bool):
+        # NOTE currently nothing in the code needs this, but it may as well exist
+        return self.reductionIndexor(runId, useLiteMode).getIndex()
```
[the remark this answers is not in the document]
rboston628: Reduction Indexer is gone.  The path methods depend on a lot of particular data (the `Indexor`'s root path, the version, and the name of the kind of files it writes).   I can think about it more.

src/snapred/backend/dao/indexing/Versioning.py
```diff
@@ -0,0 +1,61 @@
...
+        elif version == VERSION_DEFAULT_NAME:
+            version = VERSION_DEFAULT
+        elif version == VERSION_DEFAULT or version is None:
+            pass
+        elif isinstance(version, integer):
+            version = int(version)
+        elif not isinstance(version, int) or version < VERSION_START:
+            raise ValueError(f"Cannot initialize version as {version}")
+        super().__init__(**kwargs)
+        self.__version = version
```
walshmm, line 30: I feel as through parsing/translation would benefit from being seated in their own methods.

rboston628: I agree.  Let me see what I can do

src/snapred/backend/dao/normalization/NormalizationRecord.py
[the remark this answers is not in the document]
walshmm: What denotes whether something lives on the `NormalizationRecord` vs  its `calculationParameters`?
I would think inputs would go on the parameters (i.e. peakIntensityThreshold), and outputs would either live on the record or be wrapped in another datatype.

This comment is more of a philosophical question rather than a change request, though it may result in a TODO

src/snapred/backend/dao/request/ReductionRequest.py
```diff
@@ -11,7 +11,7 @@ class ReductionRequest(BaseModel):
     useLiteMode: bool
     focusGroup: Union[Optional[FocusGroup], List[FocusGroup]]
     userSelectedMaskPath: Optional[str] = None
-    version: Union[int, Literal["*"]] = "*"
+    version: Optional[int] = None
```
walshmm, line 14: Tangential, but:
1. I'm not sure version should be here yet, we dont currently have plans to support users specifying a version for reduction.  It will just refer to the index for calib/norm
2. We would probably need separate versions for norm and calib if such a feature was added

src/snapred/backend/data/DataFactoryService.py
```diff
@@ -64,7 +64,7 @@ def getGroupingMap(self, runId: str):
 
     @validate_call
     def getCalibrationDataPath(self, runId: str, useLiteMode: bool, version: int):
-        return self.lookupService._constructCalibrationDataPath(runId, useLiteMode, version)
+        return self.lookupService.calibrationIndexor(runId, useLiteMode).versionPath(version)
```
walshmm, line 67: Indexor should probably be renamed to Indexer

src/snapred/backend/data/Indexor.py
```diff
@@ -0,0 +1,367 @@
...
+        """
+        If no version given, defaults to current version
+        """
+        version = self.thisOrCurrentVersion(version)
+        filePath = self.recordPath(version)
+        record = None
+        if filePath.exists():
+            match self.indexorType:
+                case IndexorType.CALIBRATION:
+                    record = parse_file_as(CalibrationRecord, filePath)
```
walshmm, line 308: could the record class not also be an init param?  It would save you these case statements.

src/snapred/backend/dao/request/ReductionRequest.py
rboston628: Yeah, Kort mentioned that as well.  This value is what is used to lookup the calibration/normalization, and so definitely needs two values

src/snapred/backend/data/Indexor.py
rboston628: Are you okay with
``` python
record = parse_file_as(globals()[f"{self.indexorType}Record"], filePath)
```
?

rboston628: It works without the `match`, but it's ugly and not as perspicuous

src/snapred/resources/application.yml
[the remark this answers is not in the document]
walshmm: to clear up possible confusion, we could nest these values under a "friendlyName" heading?
i.e. `version.friendlyName,error`

src/snapred/backend/dao/indexing/Versioning.py
```diff
@@ -0,0 +1,70 @@
+from typing import Any, Optional
+
+from numpy import integer
+from pydantic import BaseModel, computed_field, field_serializer
+from snapred.meta.Config import Config
+
+VERSION_START = Config["version.start"]
+VERSION_NONE_NAME = Config["version.friendlyName.error"]
+VERSION_DEFAULT_NAME = Config["version.friendlyName.default"]
+VERSION_DEFAULT = -1  # SNAPRed Internal flag for default version
```
walshmm, line 10: Unfortunately I had more thoughts on this.  Would it simplify things to regor it as such?

```
class VersionFlag(str, Enum):
    VERSION_START = Config["version.start"]
    VERSION_NONE_NAME = Config["version.friendlyName.error"]
    VERSION_DEFAULT_NAME = Config["version.friendlyName.default"]
    VERSION_DEFAULT = "-1"
    def __str__(self):
        return self.value

class Version:
    value: int | VersionFlag
    def __init__(self, value):
        if isinstance(value, int) and value >= VERSION_START:
            raise ValueError....

class VersionedObject(BaseModel):
    __version: Optional[Version] = None 
   
    ...
    
    def parseVersion(cls, version, *, exclude_none: bool = False) -> int | None:
       v = None
        if a in set(VersionFlag): 
            v = VersionFlag(v)
        else:
            v = int(v)

        if v == VersionFlag.VERSION_NONE_NAME and exclude_none:
            raise ValueError("Attempted to set version to None where not acceptable.")
   

    def writeVersion(cls, version) -> int | str:
        return parseVersion(v)
        
```

walshmm: I wrote up a toy example as well just to confirm some details:
```

# Online Python - IDE, Editor, Compiler, Interpreter

from enum import Enum 

class StrEnum(str, Enum):
    pass

class SomeFlag(StrEnum):
    One = "1"
    Two = "2"
    
    def __str__(self):
        return self.value

def parseA(a):
    if a in set(SomeFlag): 
        a = SomeFlag(a)
        return a
    else:
       return int(a) 

def assertFlag(a :SomeFlag):
    assert isinstance(a, SomeFlag)
    
assertFlag(parseA("1"))
assertFlag(SomeFlag.One)
assert str(SomeFlag.One) == "1"
assert SomeFlag.One == "1"
assert parseA(1) == 1

```

src/snapred/backend/data/Indexer.py
```diff
@@ -0,0 +1,373 @@
...
+
+    @validate_call
+    def __init__(self, *, indexerType: IndexerType, directory: Path | str) -> None:
+        self.indexerType = indexerType
+        self.rootDirectory = Path(directory)
+        self.index = self.readIndex()
+        self.dirVersions = self.readDirectoryList()
+        self.reconcileIndexToFiles()
+
+    def __del__(self):
```
walshmm, line 86: This is just a safety net right?

src/snapred/backend/data/Indexer.py
```diff
@@ -0,0 +1,373 @@
...
+        return versions
+
+    def reconcileIndexToFiles(self):
+        # if a directory has no entry in the index, create one for it
+        versions = self.readDirectoryList()
+        for version in versions:
+            if version not in self.index:
+                record = self.readRecord(version)
+                self.index[version] = self.indexEntryFromRecord(record)
+        # if a version exists in the index with no corresponding directory, delete it
```
walshmm, line 111: Maybe it should be marked as "Not Found"? What if someone accidentally deleted a folder?  This may that less apparent.

src/snapred/backend/data/Indexer.py
```diff
@@ -0,0 +1,373 @@
...
+            if os.path.isdir(fname):
+                version = str(fname).split("_")[-1]
+                if version == VERSION_DEFAULT_NAME:
+                    version = VERSION_DEFAULT
+                version = int(version)
+                versions.add(version)
+        return versions
+
+    def reconcileIndexToFiles(self):
+        # if a directory has no entry in the index, create one for it
```
walshmm, line 105: What if a cis wanted to turn off a calibration temporarily?  Would this subvert that?

src/snapred/backend/dao/indexing/Record.py
```diff
@@ -0,0 +1,56 @@
...
+    @classmethod
+    def indexEntryFromRecord(cls, record) -> IndexEntry:
+        entry = None
+        if record is not None:
+            entry = IndexEntry(
+                runNumber=record.runNumber,
+                useLiteMode=record.useLiteMode,
+                version=record.version,
+                appliesTo=f">={record.runNumber}",
```
walshmm, line 39: The way this is used would create an arbitrary NEW entry in the index, granted with minimal priority, but Im not sure we want to try and generically infer an entry

src/snapred/backend/data/LocalDataService.py
```diff
@@ -323,496 +271,223 @@ def _constructReductionDataFilePath(self, runNumber: str, useLiteMode: bool, ver
...
+        indexer = self.normalizationIndexer(record.runNumber, record.useLiteMode)
+        normalizationDataPath: Path = indexer.versionPath(record.version)
+        if not normalizationDataPath.exists():
+            normalizationDataPath.mkdir(parents=True, exist_ok=True)
         for workspace in record.workspaceNames:
-            filename = workspace + "_" + wnvf.formatVersion(record.version)
             ws = mtd[workspace]
             if ws.isRaggedWorkspace():
-                filename = Path(filename + ".tar")
+                filename = Path(workspace + ".tar")
```
walshmm, line 365: Does this change mean that workspace names come with a version attached now?

src/snapred/backend/data/Indexer.py
rboston628: Yeah, if something does wrong it will save the index before closing

src/snapred/backend/data/LocalDataService.py
rboston628: At this location in the `LocalDataService`, yes.  But they are set inside the `CalibrationService` with the same process.  I think Kort wants us to push it back at the time of workspace name creation, which isn't hard but maybe we can do later.

src/snapred/backend/service/CalibrationService.py
walshmm: I think this should explicitly fail if the given version is not writable.  Ambiguity should be bubbled up if it exists.

src/snapred/backend/service/CalibrationService.py
```diff
@@ -281,11 +334,10 @@ def loadQualityAssessment(self, request: CalibrationLoadAssessmentRequest):
...
             if not wkspaceExists:
-                for wss in calibrationRecord.workspaces.values():
-                    for wsName in wss:
-                        if self.dataFactoryService.workspaceDoesExist(wsName):
-                            wkspaceExists = True
-                            break
+                for wsName in calibrationRecord.workspaces.values():
+                    if self.dataFactoryService.workspaceDoesExist(wsName):
+                        wkspaceExists = True
+                        break
```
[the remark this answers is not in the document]
walshmm: What actually necessitated this change?
There was much back and forth with Robert about allowing a generic approach to saving, as opposed to a flattened linear approach.
It actually looks like this doesnt consult the list at all anymore except for existence checking, which may not coincide with the expected template names.

src/snapred/backend/service/NormalizationService.py
```diff
@@ -162,31 +165,52 @@ def normalizationAssessment(self, request: NormalizationRequest):
...
         entry = request.normalizationIndexEntry
-        version = entry.version
-        normalizationRecord = request.normalizationRecord
-        if version is not None:
-            normalizationRecord.version = version
-        normalizationRecord = self.dataExportService.exportNormalizationRecord(normalizationRecord)
-        normalizationRecord = self.dataExportService.exportNormalizationWorkspaces(normalizationRecord)
-        entry.version = normalizationRecord.version
+        record = request.normalizationRecord
+        version = self.dataFactoryService.getThisOrNextNormalizationVersion(
```
walshmm, line 187: as per my calibration comment on explicit failure.

src/snapred/backend/service/ReductionService.py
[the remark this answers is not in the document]
walshmm: as per other conversation, this version probably shouldnt even exist in this context, as reduction will always refer to latest applicable.

src/snapred/meta/mantid/WorkspaceNameGenerator.py
```diff
@@ -83,17 +82,34 @@ def formatRunNumber(runNumber: str):
...
         if version is None:
-            return ""
-        if not version == "*":
+            version = ""
+        elif version == VERSION_DEFAULT:
+            version = VERSION_DEFAULT_NAME
+        elif str(version).isdigit():
+            version = int(version)
+        else:
+            version = ""
```
walshmm, line 95: this sort of logic feels like it should live in `Versioning` as a static method, then utilized here.

src/snapred/backend/data/Indexer.py
rboston628: Describe the process of turning off a calibration?

src/snapred/backend/data/Indexer.py
rboston628: If someone accidentally deleted a folder... then you no longer have that calibration, and there's no further purpose in retaining its index entry.

src/snapred/backend/service/NormalizationService.py
rboston628: I can add a `validate_call` tag to the method in `Indexer`, and that will cause an error if anything other than `int` or `None` is passed

src/snapred/backend/service/CalibrationService.py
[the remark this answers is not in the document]
rboston628: Nothing necessitated the change.  I had to touch the code making the lists, and I remember both of you mentioning the lists of workspaces was  no longer needed, so I made the nearby change.

src/snapred/backend/service/NormalizationService.py
rboston628: At this point we expect either:
1. version is None, or
2. version was set by the user.
If the version is None, this should set it to the next version.

walshmm (approved):
Workflows complete, loading previous calibrations seem to work, iteration still works.

PR #361, opened 2024-06-03
Save Lite data

dlcaballero16:
## Description of work

Whenever we create lite data, we will attempt to save the lite data as a nxs.h5 file to a location defined by the application.yml file. If the lite data already exists, then just load the lite data.

## Explanation of work

A call to exportWorkspace() was added to LiteDataService whenever reduceLiteData() is called. The file gets saved to /SNS/SNAP/IPTS-{ipts}/shared/lite/ and the file name is SNAP_{runNumber}.lite.nxs.h5. exportWorkspace() ultimately calls Mantid's SaveNexus() algorithm.

### Dev testing
If running locally, run diffraction for run number 46680. Before starting, look in SNS/SNAP/IPTS-24641/shared/lite and verify there is no file there. Run diffraction to the Tweak Peaks step. You should now see this file: SNAP_46680.lite.nxs.h5. Complete the entire diffraction and save. Wait about a minute. Restart diffraction. Get to the Tweak Peaks step again and verify the time stamp of the lite data has not been updated.

If running on analysis, update your application.yml file so that instrument.calibration.home now points to /SNS/SNAP/shared/Calibration_dynamic/. In GrocerySevice.py, add the following right before the if check on line 653:
``` python
loadedFromNative = True
nativeRawWorkspaceName = self._createRawNeutronWorkspaceName(runNumber, False)
```
The purpose of this is to force lite data to be created since the lite data already exists on analysis. Run diffraction with run number 46680. At the Tweak Peaks step, you will see some error messages that should look like this:

ERROR    - Mantid - Unable to open file /SNS/SNAP/IPTS-24641/shared/lite/SNAP_46680.lite.nxs.h5
ERROR    - Mantid - Error in execution of algorithm SaveNexusProcessed:
ERROR    - Mantid - Unable to open File: in "/SNS/SNAP/IPTS-24641/shared/lite/SNAP_46680.lite.nxs.h5"
ERROR    - Mantid - Unable to successfully run SaveNexusprocessed Child Algorithm
ERROR    - Mantid - Unable to successfully run SaveNexusProcessed Child Algorithm

These errors are expected because developers don't have write permissions on analysis. These errors should not stop diffraction and you should be able to continue. Complete diffraction and make sure nothing else is broken.

### CIS testing

Follow the same steps for dev testing locally, but do them on analysis. You should not be able to see any errors related to permissions since CIS should have the correct permissions.

## Link to EWM item

[EWM#3128](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3128)


src/snapred/backend/service/LiteDataService.py
```diff
@@ -12,6 +15,8 @@ class LiteDataService(Service):
     def __init__(self):
         super().__init__()
         self.registerPath("createLiteData", self.reduceLiteData)
+        self.dataExportService = DataExportService()
+        self.dataService = LocalDataService()
```
rboston628, line 19: This service should not be able to access the `LocalDataService`.  It should instead have
``` python
self.dataFactoryService = DataFactoryService()
```

src/snapred/backend/service/LiteDataService.py
```diff
@@ -28,6 +33,7 @@ def reduceLiteData(
         self,
         inputWorkspace: WorkspaceName,
         outputWorkspace: WorkspaceName,
+        runNumber: str,
```
rboston628, line 36: Is there a way to remove this input?  It should be present in the workspace name.

src/snapred/backend/service/LiteDataService.py
dlcaballero16: getIPTS() does not exist in DataFactoryService. Should I add a function in DataFactoryService that calls getIPTS()?

src/snapred/backend/service/LiteDataService.py
dlcaballero16: Yep, updated to just grab from workspace name

src/snapred/backend/service/LiteDataService.py
rboston628: Add it to DataExportService instead, then

rboston628: Actually, don't add a `getIPTS` method, instead add a method that will create the path you create below

src/snapred/backend/data/DataExportService.py
```diff
@@ -33,6 +34,13 @@ def _defaultClass(self, val, clazz):
     def exportCalibrantSampleFile(self, entry: CalibrantSamples):
         self.dataService.writeCalibrantSample(entry)
 
+    def getFullLiteDataFilePath(self, runNumber: str):
+        path = self.dataService.getIPTS(runNumber)
+        pre = "nexus.lite.prefix"
+        ext = "nexus.lite.extension"
+        fileName = Config[pre] + str(runNumber) + Config[ext]
+        return path + fileName
```
rboston628, line 42: Make this a path,
``` python
return Path(path, fileName)
```

src/snapred/backend/service/LiteDataService.py
```diff
@@ -31,13 +33,18 @@ def reduceLiteData(
...
         try:
             data = Recipe().executeRecipe(
                 InputWorkspace=inputWorkspace,
                 LiteDataMapWorkspace=liteDataMap,
                 LiteInstrumentDefinitionFile=instrumentDefinition,
                 OutputWorkspace=outputWorkspace,
             )
+            fullPath = self.dataExportService.getFullLiteDataFilePath(runNumber)
+            path = fullPath.rsplit("/", 1)[0]
+            fileName = fullPath.rsplit("/", 1)[1]
```
rboston628, line 46: Use `fullPath.parts` instead

rboston628 (dismissed):
I made a nice long review with pictures and stuff, and then github decided, no, we don't want you to post that review... so now it's gone.

With 46680, I verified the lite file didn't exist.  At tweak peek view, it did.  I was able to get all the way to saving in both workflows.

When I ran calibration again, the file loaded instantly, because it loaded from the lite data file instead.

On analysis, using run 57514, I was able to see the error message due to lack of permissions.  Despite this error, the program continued and I could get all the way to the saving screen.

rboston628 (approved):

PR #363, opened 2024-06-05
Fix for initialize state issues.

darshdinger:
## Description of work
This PR is for fixing the initialize state workflow.   

## Explanation of work
This PR adds a feature that allows for the simultaneous initialization of both native and lite modes. Along with this, a rather quick fix was found for the issue of two instances of the initialize state workflow being substantiated when a Recoverable error is thrown. The majority of the time in this PR went to stack tracing and finding the location of the issue.

Some helpful debugging commands I picked up along the way [all pdb.set_trace() commands]:
`list .` - to see what breakpoint is being accessed
`w` - to see a stack trace of the executed methods leading up to any given breakpoint

## To test
Insure all pytests pass. Launch SNAPRed with an empty state folder in order to proceed through the initialize state workflow and insure that only one instance is present. After the initialize state workflow is complete, inspect the logging stack trace or the state directory produced to insure that both native and lite modes are initialized. 

## Link to EWM item

[EWM # 5405](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5405)

### DIFF
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -1031,7 +1031,7 @@ class LocalDataService:
             peakTailCoefficient=peakTailCoefficient,
         )
 
-        calibrationReturn = None
+        calibrationReturnValue = None
 
         for liteMode in [True, False]:
             # finally add seedRun, creation date, and a human readable name
@@ -1057,9 +1057,9 @@ class LocalDataService:
             self._writeDefaultDiffCalTable(runId, liteMode)
 
             if useLiteMode == liteMode:
-                calibrationReturn = calibration
+                calibrationReturnValue = calibration
 
-        return calibrationReturn
+        return calibrationReturnValue
 
     def _prepareStateRoot(self, stateId: str):
         """
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -1031,28 +1031,35 @@ class LocalDataService:
             peakTailCoefficient=peakTailCoefficient,
         )
 
-        # finally add seedRun, creation date, and a human readable name
-        calibration = Calibration(
-            instrumentState=instrumentState,
-            name=name,
-            seedRun=runId,
-            useLiteMode=useLiteMode,
-            creationDate=datetime.datetime.now(),
-            version=self.VERSION_START,
-        )
+        calibrationReturn = None
 
-        # Make sure that the state root directory has been initialized:
-        stateRootPath: Path = self._constructCalibrationStateRoot(stateId)
-        if not stateRootPath.exists():
-            # WARNING: `_prepareStateRoot` is also called at `readStateConfig`; this allows
-            #   some order independence of initialization if the back-end is run separately (e.g. in unit tests).
-            self._prepareStateRoot(stateId)
+        for LiteMode in [True, False]:
+            # finally add seedRun, creation date, and a human readable name
+            calibration = Calibration(
+                instrumentState=instrumentState,
+                name=name,
+                seedRun=runId,
+                useLiteMode=LiteMode,
+                creationDate=datetime.datetime.now(),
+                version=self.VERSION_START,
+            )
 
-        # write the calibration state
-        self.writeCalibrationState(calibration, version)
-        # write the default diffcal table
-        self._writeDefaultDiffCalTable(runId, useLiteMode)
-        return calibration
+            # Make sure that the state root directory has been initialized:
+            stateRootPath: Path = self._constructCalibrationStateRoot(stateId)
+            if not stateRootPath.exists():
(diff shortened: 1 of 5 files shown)


src/snapred/backend/data/LocalDataService.py
```diff
@@ -1031,28 +1031,35 @@ def initializeState(self, runId: str, useLiteMode: bool, name: str = None):
...
-            instrumentState=instrumentState,
-            name=name,
-            seedRun=runId,
-            useLiteMode=useLiteMode,
-            creationDate=datetime.datetime.now(),
-            version=self.VERSION_START,
-        )
+        calibrationReturn = None
+
+        for LiteMode in [True, False]:
```
rboston628, line 1036: 1. use `liteMode`.  `LiteMode` looks like a class name.
2. Does this need to be included in this PR?

darshdinger: Done.

src/snapred/ui/handler/SNAPResponseHandler.py
```diff
@@ -33,11 +33,13 @@ def _handle(self, result):
...
-            self.rethrow()
+            self.rethrow(result)
 
+    @staticmethod
     def _isErrorCode(code):
         return code >= ResponseCode.ERROR
 
+    @staticmethod
     def _isRecoverableError(code):
         return ResponseCode.RECOVERABLE <= code < ResponseCode.ERROR
```
ekapadi, line 44: This kind of _logic_ statement depends on precedence rules for parsing.  Personally, I never remember precedence rules so I have _no_ _idea_ what this does.  Does it mean `(ResponseCode.RECOVERABLE <= code) and (code < ResponseCode.ERROR)`  OR does it mean (when `code < ResponseCode.ERROR` => `ResponseCode.RECOVERABLE < True`...  Do you see what I mean?

tests/util_tests/test_state_helpers.py
```diff
@@ -75,6 +106,8 @@ def test_state_root_override_enter(mockReadPVFile, mockReadInstrumentConfig, moc
     mockDefaultGroupingMapPath.return_value = Path(Resource.getPath("inputs/pixel_grouping/defaultGroupingMap.json"))
 
     stateId = "ab8704b0bc2a2342"
+    decodedKey = "some_decoded_key"
+    mockGenerateStateId.return_value = (stateId, decodedKey)
```
ekapadi, line 110: I strongly recommend that you use `None` here as the decoded key (as the "decoded" part is optional).  Otherwise the actual key will be invalid, which might screw someone up in our testing future.

ekapadi (changes requested):

tests/util_tests/test_state_helpers.py
darshdinger: Fixed!

src/snapred/ui/handler/SNAPResponseHandler.py
darshdinger: This is essentially equivalent to: 

```python
(ResponseCode.RECOVERABLE <= code) and (code < ResponseCode.ERROR)
``` 

This means the statement checks if `code` is greater than or equal to `Response.RECOVERABLE` and also less than `ResponseCode.Error`.

It does **NOT** mean:
```python
when code < ResponseCode.ERROR => ResponseCode.RECOVERABLE < True
```

This is left this way in the implementation so that more Recoverable errors can be assigned to response codes that are between greater than 400 (recoverable) and less than 500 (error). State initialization is allocated to 400 or 401 I think. But other recoverable errors and subsequent divergent workflow branches can be allocated properly later on.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -1031,28 +1031,35 @@ def initializeState(self, runId: str, useLiteMode: bool, name: str = None):
...
+                #   some order independence of initialization if the back-end is run separately (e.g. in unit tests).
+                self._prepareStateRoot(stateId)
+
+            # write the calibration state
+            self.writeCalibrationState(calibration, version)
+            # write the default diffcal table
+            self._writeDefaultDiffCalTable(runId, liteMode)
+
+            if useLiteMode == liteMode:
+                calibrationReturn = calibration
```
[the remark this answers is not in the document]
darshdinger: Yes, this makes sense. Fixed!

src/snapred/ui/handler/SNAPResponseHandler.py
ekapadi: I'm fine with this.  It's probably more Pythonic anyway; I have just never seen it before.   I need to take a look at the comparison operator syntax and see what it actually allows.  :)

ekapadi (approved):

PR #365, opened 2024-06-06
Fix labeling of diffcal table

rboston628:
## Description of work

The `GroceryService` was being used like a wrapper of the `WNG` in order to create workspace names for diffcal output.

That isn't a problem, except the expectation was `GroceryService` would only be used to _load_ data.  In this case, it anticipated loading previous calibration tables.

The collision left an incorrect version number on the produced diffcal table.  The version applied to the previously-created record and was added to the name inside `GroceryService`.  However, it was being added to the prospective, newly-created diffcal table instead in the process of generating a name.

The easiest fix was to not use the `GroceryService` as a wrapper of the `WNG`, and to instead use the `WNG` to create the output names.  This also simplified the `fetchGroceryList` method.

## Explanation of work

Removed all reference to `specialOrder` or `isOutput` in the `GroceryListItem` and its builder.

Removed the conditionals checking on `isOutput` from the `fetchGroceryList` method inside `GroceryService`.

Inside calibration service, construct the output names using the `wng` and pass them to `fetchGroceryDict` as additional keyword arguments, which adds them to the dictionary in an exactly equivalent way.

#### First, reproduce the error on `next`

Clear your state folder and run diffcal.  Check the created diffcal table, and it has no version on its name -- this is because there is no record, so the line to add the version is not being executed.  At assessment, there are no versions.  Continue and save.

Run diffcal again.  Check the created diffcal table.  It has `_v0001` tagged to its name.  This is because `GroceryService` loaded a previous record (`v_0001`) and added its version number to the workspace name.  At assessment, try to load `v0001`.  You can't, because it thinks `v0001` is already loaded due to the presence of the errantly-named diffcal table.

#### Second, switch to this branch and see error is gone

Clear your state folder and run diffcal.  Check the created diffcal table, and it has no version on its name.  At assessment, there are no versions.  Continue to save.

Run diffcal again.  The created diffcal table should **NOT** have any version on its name.

At asessment, try to load `v0001`.  It will load with no issue and you will see the associated workspaces with `_v0001` appear in the workspace list.

Save again.

Run normalization, and make sure you can save.

### CIS testing

As above. Inspect the diffcal tables to make sure the values appear reasonable (right number of pixels, values around 5000-8000, etc.).

## Link to EWM item

[EWM#5542](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5542)

### DIFF
--- a/src/snapred/backend/dao/ingredients/GroceryListItem.py
+++ b/src/snapred/backend/dao/ingredients/GroceryListItem.py
@@ -67,11 +67,6 @@ class GroceryListItem(BaseModel):
     # name the property the workspace will be used for
     propertyName: Optional[str]
 
-    # flag to indicate if this is an _output_ workspace:
-    # an output workspace will not be loaded,
-    # it may or may not already exist in the ADS
-    isOutput: bool = False
-
     def builder():
         # NOTE this import is here to avoid circular dependencies -- don't bother trying to move it
         from snapred.meta.builder.GroceryListBuilder import GroceryListBuilder
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -945,7 +945,7 @@ class GroceryService:
         :param groceryDict: a dictionary of GroceryListItems, keyed by a property name
         :type groceryDict: Dict[str, GrocerListItem]
         :param kwargs: keyword arguments will be added to the created dictionary as argName: argValue.
-        Use this to add additional workspaces to the dictionary for easier use in recipes.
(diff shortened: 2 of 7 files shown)


ekapadi (approved):
I verified on "next" that the defect occurs as described (on "analysis.sns.gov").
I verified on this branch that the defect no longer occurs, and that the diffraction-calibration panel works as expected.  Additionally, all unit tests pass (on "analysis.sns.gov" and on a local system).

ekapadi (approved):

PR #368, opened 2024-06-10
SNAPRed Pydantic v2

ekapadi:
## Description of work

Transition the SNAPRed codebase to Pydantic v2.

## Explanation of work

These changes transition the entire SNAPRed codebase to Pydantic v2.7.3.  This work was accomplished by applying the "bump-pydantic" script (see: https://github.com/pydantic/bump-pydantic), followed by _extensive_ "by hand" changes where required.
    
  * Most, but not _all_, pydantic-related deprecation warnings have been addressed.  The remaining ones are either _indirect_ warnings, or are advanced enough that they need to be adjusted on a case-by-case basis.  Note that there are not many of this latter type.  The warnings that have not been addressed are not expected to change the successful _function_ of the code.
    
  * `model_dump()` was not substituted for `dict()` (_yet_) primarily because in many cases, `dict()` is also a non-pydantic method.  The same applies to some applications of the `json()` method.  Note that both of these methods have been overridden in some of the Crystallography classes, along with the implementation of a custom `__init__` method.  In general, defining a custom  `__init__` is not considered to be good practice when working with Pydantic. For this reason, these classes should _probably_ be re-implemented, but this was out-of-scope for this PR.  Regardless, they continue to work as they did.

  *  As we progress in our development work, we can gradually remove these remaining deprecation warnings.

### Dev testing

All of the unit tests should pass, both on "analysis.sns.gov" and on a _local_ system.
In the GUI: the "calibration" and "normalization" panels should function as expected.

******
**Note that a defect with respect to iteration-naming of `diagnostic` workspaces was uncovered during this work -- this defect is also present on "next".**  [EWM#5668](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=5668)
******

## Link to EWM item

[EWM#4720](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=4720)

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,7 @@ channels:
 dependencies:
 - python=3.10
 - pip
-- pydantic>=1.10,<2
+- pydantic>=2.7.3,<3
 - mantidworkbench>=6.8.20231213
 - qtpy
 - pre-commit
@@ -14,7 +14,7 @@ dependencies:
 - pytest-qt
 - pytest-cov
 - flake8
-- erdantic=0.5.1
+- erdantic>=1.0.3
 - versioningit
 - anaconda-client
 - conda-build
@@ -29,4 +29,4 @@ dependencies:
 - pycifrw
 - pyqtwebengine
 - pip:
-  - check-wheel-contents==0.4.0
+  - check-wheel-contents>=0.6.0
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -30,18 +30,18 @@ requirements:
   host:
     - python
     - versioningit
-    - pydantic=1.10
+    - pydantic>=2.7.3,<3
 
   build:
     - setuptools
     - versioningit
-    - {{ pin_compatible("pydantic", min_pin="1.10", max_pin="2") }}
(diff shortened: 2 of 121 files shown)


environment.yml
```diff
@@ -6,15 +6,15 @@ channels:
 dependencies:
 - python=3.10
 - pip
-- pydantic>=1.10,<2
+- pydantic>=2.7.3
```
walshmm, line 9: Looks mostly fine, though maybe we should pin it to below three?  not sure what their timelines are like but would also not be keen on this randomly breaking.   The difficulty with that arises from the other deps that depend on pydantic, as their versioning isnt as straight forward(check wheel content specifically)

darshdinger:
Attached is my review of GUI testing within this branch.
[Pydantic_V2_GUI_test_Notes.txt](https://github.com/user-attachments/files/15791639/Pydantic_V2_GUI_test_Notes.txt)

There were no issues I found just some warnings listed within the .txt file above but looks good and is functional.

PR #371, opened 2024-06-11
Added grouping functions to Reduction Service 

dlcaballero16:
## Description of work

This adds the grouping functions needed to sort SNAPRequests in the Reduction Service.

## Explanation of work

Two grouping functions were added to the ReductionService. One sorts requests based on StateId, and the other sorts on the normalization version. A final function was added that calls the base RequestScheduler class and uses the grouping functions. The expected output is so that the first grouping is sorted by StateId, then sorted by version. A function was added to the DataFactoryService to be able to get the normalization version number from index. For now, the ReductionService does not actually call the grouping function, but will be used later in the future.

### Dev testing

Make sure unit tests pass.

### CIS testing

None.

## Link to EWM item

[EWM#5286](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5286)


src/snapred/backend/data/DataFactoryService.py
```diff
@@ -128,6 +128,10 @@ def getNormalizationDataWorkspace(self, runId: str, useLiteMode: bool, version:
         path = self.getNormalizationDataPath(runId, useLiteMode, version)
         return self.groceryService.fetchWorkspace(os.path.join(path, name) + ".nxs", name)
 
+    @validate_arguments
+    def getNormalizationVersion(self, runId: str, useLiteMode: bool):
+        return self.lookupService._getVersionFromNormalizationIndex(runId, useLiteMode)
```
walshmm, line 133: could you make a wrapper method in localdataservice called `getVersionFromNormalizationIndex` that just invokes `_getVersionFromNormalizationIndex`, typically methods prepended with an '_' are considered private.

walshmm (approved):

PR #374, opened 2024-06-14
Introduce parameter to exclude short wavelengths

darshdinger:
## Description of work
In inspection of normalized diffraction focused sample data, Malcolm found that at low d-spacing values, the observed signal was essentially zero (100% background). This story is meant to act as a fix for this issue. The main work for this story and the majority of the hours went into inspecting the difference between original diffraction focused workspace and the post fix diffraction focused workspace within the Normalization Workflow.  

## Explanation of work
A parameter called `lowWavelengthCrop` was added based on Malcolm's suggestion and set to a default value of 0.05. This parameter was then exposed within the `intializeState` method within `LocalDataService`.

In testing this myself, I ran the Normalization workflow for many different run numbers and parameters, found in most cases that the post fix values for dSpacing were in fact noticeably different:

Pre-fix:
image

Post-Fix:
image

Malcolm will have to come back and insure this is the difference we truly want from a scientific perspective.

### Dev testing
Run this 

To test, please follow these steps:
1. Make sure you are on the `next` branch.
2. Delete all initialized states within Powder directory for your local directory.
3. Run normalization workflow with the following inputs:
    - Run Number: 58882
    - Background: 58813
    - Sample: Silicon_NIST_640D_001.json
    - Grouping: Default
4. When on the `Tweak Parameters` page, go to the main SNAPRed application and from within the workspaces viewer, double click on the workspace title `tof_all_default_lite_s+f-vanadium_058882`.
5. Inspect the lower dSpacing range by using the zoom tool.

Then switch to branch `introduce-parameter-to-exclude-short-wavelength` and follow all the same steps as listed above in order. If you take screenshots of both instances, you should get the same result as within the explanation of the work above. 

**Note:** It is import that you delete states and initialize from scratch both times as this new parameter get substantiated through the initialization workflow.
### CIS testing
Please follow the steps above to run the Normalization Workflow and confirm that values at low d-Spacing are proper.

[EWM # 3520](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3520)

### DIFF
--- a/src/snapred/backend/dao/InstrumentConfig.py
+++ b/src/snapred/backend/dao/InstrumentConfig.py
@@ -27,3 +27,4 @@ class InstrumentConfig(BaseModel):
     delThWithGuide: float
     width: float
     frequency: float
+    lowWavelengthCrop: float = 0.05
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -1036,7 +1036,7 @@ class LocalDataService:
         )
         # then calculate the derived values
         lambdaLimit = Limit(
-            minimum=detectorState.wav - (instrumentConfig.bandwidth / 2),
+            minimum=detectorState.wav - (instrumentConfig.bandwidth / 2) + instrumentConfig.lowWavelengthCrop,
             maximum=detectorState.wav + (instrumentConfig.bandwidth / 2),
         )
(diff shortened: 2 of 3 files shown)


dlcaballero16 (approved):
Tested and got results as expected from given screenshots.

PR #375, opened 2024-06-15
Version integration test :card_index: :bar_chart: 

rboston628:
## Description of work

This follows on from the `Indexor` work in PR #360.  The goal is to create an integration test that runs the calibration workflow from the service layer and verifies that the state folder is created and versions are saved in the correct order.

To accomplish this, the data layer needed to be gently modified to point at SNAPRed internal files rather than files on analysis.  The changes are minimal, and are NOT mocks.  Functions are actually being called, logic runs, and things are written to/read from the disk.

As it runs, it saves diffcals at different versions, and then verifies the correct versions are saved and the `Indexor` is in the correct state.

This test confirms that:
- the API can call to the calibration service without issue
- calibration runs, on synthetic data (with no check for accuracy of results)
- calibration can call to the data layer for saving
- the data layer saves files in the correct locations, with correct names, with correct versions

This is *almost* a system test, excluding the UI, where the system being tested is the ability to write calibration versions.

## Explanation of work

The calibration calculation is actually being performed on the synthetic data, and all elements (fetching groceries, prepping ingredients, calling the algorithms in the recipe) are all actually orchestrated and passing valid values.

The modifications to the data layer were performed by deriving from the data layer classes and overriding a few key functions to point to new locations or new data objects.
- `LocalDataService` was overridden to read a grouping map from the test inputs directory, and to write to a directory inside the test outputs, and otherwise point to needed files in the test inputs.
- `GroceryService` was overridden to return workspaces made from `SyntheticData`, rather than loading them from an `IPTS` folder.
- `SousChef` was overridden to return the diffcal ingredients inside `SyntheticData`, to ensure these are compatible with the workspaces.

The test makes use of  the test instrument ("POP") to perform the calculation.  This allows it to run quickly, while still validating the orchestration.

**IMPORTANT**

There is an apparent bug inside Mantid when saving the diagnostic workspace using `SaveNexusProcessed` which causes a segmentation fault.  I haven't been able to reproduce this bug in any other way than by running this test in pytest.

To avoid the segfault, I overrode the `writeWorkspaces` method to print the diagnostic workspace as a simple text file, rather than using Mantid algorithms.  This prevents the segfault.  When we figure out why the segfault is occurring, this should be changed back.

### Dev testing

Remove the lines to delete the workspaces (`shutil.rmtree`).  Run the test.  At the end, inspect the directory and the created files.  Make sure they all have the versions that we expect, and the workspaces that we expect.

### CIS testing

This is an internal enabler, CIS testing is not needed.

## Link to EWM item

[EWM#5636](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5636)


rboston628:
Do not merged before PR #375 375

darshdinger:
There is a merge conflict with LocalDataService now that needs to be addressed. 

All pytests pass successfully!

Followed test instructions with commenting out the line with `shutil.rmtree` and received a number of files within diff after test successfully ran:

defaultGroupingMap.json:
image

groupingMap.json:
image

CalibrationIndex.json:
image

CalibrationParameters.json:
image

CalibrationRecord.json:
image

CalibrationIndex.json:
image

CalibrationParameters.json:
image

CalibrationRecord.json:
verison: 0

CalibrationParameters.json:
image

CalibrationRecord.json:
version: 7

CalibrationParameters.json:
image

CalibrationRecord.json:
version: 0xF

Other files produced:

- _diffract_consts_default_v0xf.h5
- _diagnostic_column_000123_v0000.nxs.h5
- _diffract_consts_000123_v0000.h5
- _dsp_column_000123_v0000.tar
- _diagnostic_column_000123_v0007.nxs.h5
- _diffract_consts_000123_v0007.h5
- _dsp_column_000123_v0007.tar
- _diffract_consts_default_v0xf.h5

walshmm:
is this really an integration test if we are essentially reimplementing portions for testing alone?  How much overlap is there with the integration testing @ekapadi has been doing as part of [story 5537](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5537)?

rboston628:
> is this really an integration test if we are essentially reimplementing portions for testing alone?

The reimplementations are very limited.  In `LocalDataService` they are limited to redirecting certain file paths, meaning all of the other methods that setup state folders and initialize default objects have to be called.  In `GroceryService`, they are limited to the "`fetch`" methods, meaning the entire process of building a grocery list and processing it still takes place.  In `SousChef` one method was fixed to return compatible ingredients.

The goal was the reimplementations would not impact communication between layers or services.  The change to `SousChef` is the only exception, and is necessary for the workflow to run.

> How much overlap is there with the integration testing @ekapadi has been doing as part of [story 5537](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5537)?

The difference with story 5537 is that work is aimed at ensuring everything runs correctly within the UI workflow.  This is focused in particular on checking saved versions of files, to ensure future changes don't scramble up versioning.

walshmm (approved):

PR #376, opened 2024-06-17
Validate Run Numbers

dlcaballero16:
## Description of work

A new validator package was created and a validator for run numbers was added.

## Explanation of work

The run number validator makes sure a given run number is an actual number and that the run number is within a given range. A warning gets logged when a given run number is not valid.

### Dev testing

In all 3 tabs, you should be able to put in a wrong run number and get a warning that makes sense. "notanumber" should return a warning that says an integer is required. "9999" should result in a warning that the number is too small, and anything over 46342 should result in a warning that the number is too big.

### CIS testing

None.

## Link to EWM item

[EWM#5546](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5546)

### DIFF
--- a/src/snapred/backend/dao/validator/RunNumberValidator.py
+++ b/src/snapred/backend/dao/validator/RunNumberValidator.py
@@ -1,0 +1,26 @@
+from pydantic import BaseModel
+from snapred.backend.log.logger import snapredLogger
+from snapred.meta.Config import Config
+
+logger = snapredLogger.getLogger(__name__)
+
+
+class RunNumberValidator(BaseModel):
+    @classmethod
+    def validateRunNumber(cls, runNumber: str):
+        if not runNumber.isdigit():
+            # raise RecoverableException("Run number must be a positive integer")
+            logger.warning("Run number must be a positive integer")
+            return False
+        number = int(runNumber)
+        minRunNumber = Config["instrument.startingRunNumber"]
+        maxRunNumber = Config["instrument.maximumRunNumber"]
+        if number < minRunNumber:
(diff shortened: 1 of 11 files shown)


src/snapred/backend/dao/validator/RunNumberValidator.py
rboston628, line 1: 1. This doesn't need to be a pydantic model.  It can be a bare class.
2. This should probably live in the `meta` folder and not in the `dao` folder,
3. In the special case of a maximum run number of `0`, the code should allow arbitrarily large run numbers.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -1122,8 +1123,10 @@ def checkCalibrationFileExists(self, runId: str):
         # first perform some basic validation of the run ID
         # - it must be a string of only digits
         # - it must be greater than some minimal run number
-        if not runId.isdigit() or int(runId) < Config["instrument.startingRunNumber"]:
+        if not RunNumberValidator.validateRunNumber(runId):
             return False
+        # if not runId.isdigit() or int(runId) < Config["instrument.startingRunNumber"]:
+        #     return False
```
rboston628, line 1129: Remove commented-out code.

darshdinger (approved):
Reviewed and tested! Looks good. Functions as required.

rboston628:
I ran this independently, and confirmed it requires a valid integer in the calibration run number field.

PR #378, opened 2024-06-18
Enable `parse_file_as`, `parse_obj_as`, `parse_raw_as` methods in `redantic`

rboston628:
## Description of work

Pydantic V2 fatuously removed extremely useful methods that we had come to rely on, and are [completely unwilling](https://github.com/pydantic/pydantic/issues/9634) to re-include the very useful file parsing methods after their users became reliant on them (thanks pydantic guys!)

This adds similar methods inside redantic which should make it easier to read files.

## Explanation of work

I used the pydantic V2 `TypeAdapter` to write methods similar to the `parse_file_as` and related methods.

### Dev testing

Check the unit tests.  If you want, try running some parses in a script or interpreter.

### CIS testing

Not needed, this is purely internal.

## Link to EWM item

This is not from an EWM, though it originated out of the V2 bump and I found these useful during work on the Indexor PR.

### DIFF
--- a/src/snapred/meta/redantic.py
+++ b/src/snapred/meta/redantic.py
@@ -1,8 +1,26 @@
 import json
-from typing import Any, List
+from pathlib import Path
+from typing import Any, List, Type, TypeVar, Union
 
-import pydantic
-from pydantic import BaseModel
+from pydantic import BaseModel, TypeAdapter
+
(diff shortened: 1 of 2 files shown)


darshdinger (approved):
Tested and functional, all unit tests pass as well.

PR #379, opened 2024-06-21
Calibration implement workspace unit validators

darshdinger:
## Description of work
This work is to facilitate the addition of mantid's workspace unit validators within SNAPRed's backend user facing algos. This work is related to the calibration workflow specifically, please look at the EWM link below to get a breakdown of the units validated per algo. Also there is information on other algos involved within the workflow.

## Explanation of work
This turned out to be simpler than originally estimated as we have 28 algos in total within the backend but only a hand full of these turned out to be "user facing". In most cases implementation was simple and straight forward. The `WorkspaceUnitValidator` class is imported from mantid simple api and then used within the workspace definition where appropriate. Pytests were fixed as needed (or not needed in some cases). Please check the EWM story description if needed.

### Dev testing
Insure all pytests pass successfully. Open branch within analysis and insure that Calibration workflow is successfully regression tested from beginning to the end saving step without issues.

### CIS testing
Not needed.

## Link to EWM item

[EWM # 5769](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5769)

### DIFF
--- a/src/snapred/backend/recipe/algorithm/FocusSpectraAlgorithm.py
+++ b/src/snapred/backend/recipe/algorithm/FocusSpectraAlgorithm.py
@@ -5,6 +5,7 @@ from mantid.api import (
     MatrixWorkspaceProperty,
     PropertyMode,
     PythonAlgorithm,
+    WorkspaceUnitValidator,
 )
 from mantid.kernel import Direction, StringMandatoryValidator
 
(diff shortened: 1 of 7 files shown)


dlcaballero16 (approved):
Tested and was able to get to saving step and save with no issues.

PR #380, opened 2024-06-21
Normalization implement workspace unit validators

darshdinger:
## Description of work
This work is to facilitate the addition of mantid's workspace unit validators within SNAPRed's backend user facing algos. This work is related to the normalization workflow specifically, please look at the EWM link below to get a breakdown of the units validated per algo. Also there is information on other algos involved within the workflow.

## Explanation of work
This turned out to be simpler than originally estimated as we have 28 algos in total within the backend but only a hand full of these turned out to be "user facing". In most cases implementation was simple and straight forward. The `WorkspaceUnitValidator` class is imported from mantid simple api and then used within the workspace definition where appropriate. Pytests were fixed as needed (or not needed in some cases). Please check the EWM story description if needed.

## To test
Insure all pytests pass successfully. Open branch within analysis and insure that Normalization workflow is successfully regression tested from beginning to the end saving step without issues.

### CIS testing
Not needed.

## Link to EWM item

[EWM # 5763](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5763)

### DIFF
--- a/src/snapred/backend/recipe/algorithm/FocusSpectraAlgorithm.py
+++ b/src/snapred/backend/recipe/algorithm/FocusSpectraAlgorithm.py
@@ -14,6 +14,14 @@ from snapred.backend.recipe.algorithm.MantidSnapper import MantidSnapper
 
 
 class FocusSpectraAlgorithm(PythonAlgorithm):
+    """
+    This algorithm performs diffraction focusing on TOF data. It converts the
+    input workspace from time-of-flight (TOF) to d-spacing and applies
+    diffraction focusing using a grouping workspace. Optionally, it rebins
+    the output to ensure uniform binning.
+
+    """
+
     def category(self):
         return "SNAPRed Data Processing"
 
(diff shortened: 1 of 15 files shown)


tests/unit/backend/recipe/algorithm/test_FocusSpectraAlgorithm.py
```diff
@@ -205,6 +206,10 @@ def test_failedValidation_inputGroupingIdsMismatch(self):
             NumBanks=5,  # must wrong for test
             BankPixelWidth=2,  # each bank has 4 pixels, 4 banks, 16 total
         )
+
+        # Convert the workspace to TOF units
+        ConvertUnits(InputWorkspace=self.fakeRawData, OutputWorkspace=self.fakeRawData, Target="TOF")
```
rboston628, line 211: In the `CreateSampleWorkspace` method above, you can change `XUnit` to `"TOF"`.

src/snapred/backend/recipe/algorithm/FocusSpectraAlgorithm.py
rboston628, line 1: If I'm correct, the point of this algorithm is to diffraction focus TOF data?  Could you add a docstring explaining that.

rboston628 (approved):

rboston628 (approved):

PR #381, opened 2024-06-21
Added MetadataLookupService to verify list of multiple runs

dlcaballero16:
## Description of work

This new service takes in a list of run numbers that can be passed to a mantid IntArrayProperty, and then that list of runs is verified.

## Explanation of work

The new MetadataLookupService takes in a string of runs that can be accepted by an IntArrayProperty object. You can pass in a list several different ways, including by doing 100:200. When colon separated, it will return a list of all values in between. This list then gets verified by RunNumberValidator and then gets inserted into a list of valid runs. If the list is too big, the return list will be empty. The max size of run numbers is set in the application.yml file by instrument.maxNumberOfRuns. This value is currently set to an arbitrary value of 10.

### Dev testing

Make sure unit tests pass.

### CIS testing

None

## Link to EWM item

[EWM#5006](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5006)

### DIFF
--- a/src/snapred/backend/service/MetadataLookupService.py
+++ b/src/snapred/backend/service/MetadataLookupService.py
@@ -29,16 +29,13 @@ class MetadataLookupService(Service):
     @FromString
     def verifyMultipleRuns(self, runs: str):
         maxRuns = Config["instrument.maxNumberOfRuns"]
-        try:
-            allRuns = IntArrayProperty(name="RunList", values=runs).value
-        except Exception as e:
-            raise e
+        allRuns = IntArrayProperty(name="RunList", values=runs).value
         validRuns = []
         for run in allRuns:
             if RunNumberValidator.validateRunNumber(str(run)):
                 validRuns.append(run)
         if len(validRuns) > maxRuns:
-            logger.warning(f"Maximum vale of {maxRuns} run numbers exceeded")
+            logger.warning(f"Maximum value of {maxRuns} run numbers exceeded")
             validRuns = []
 
         return validRuns
--- a/src/snapred/backend/service/MetadataLookupService.py
+++ b/src/snapred/backend/service/MetadataLookupService.py
@@ -1,0 +1,44 @@
+from mantid.kernel import IntArrayProperty
+
+from snapred.backend.data.DataFactoryService import DataFactoryService
+from snapred.backend.log.logger import snapredLogger
+from snapred.backend.service.Service import Service
+from snapred.meta.Config import Config
+from snapred.meta.decorators.FromString import FromString
+from snapred.meta.decorators.Singleton import Singleton
+from snapred.meta.validator.RunNumberValidator import RunNumberValidator
+
+logger = snapredLogger.getLogger(__name__)
+
+
+@Singleton
+class MetadataLookupService(Service):
+    dataFactoryService = "DataFactoryService"
+
+    # register the service in ServiceFactory please!
+    def __init__(self):
+        super().__init__()
+        self.dataFactoryService = DataFactoryService()
+        self.registerPath("", self.verifyMultipleRuns)
+        return
+
+    @staticmethod
+    def name():
+        return "verifiedRuns"
+
+    @FromString
+    def verifyMultipleRuns(self, runs: str):
+        maxRuns = Config["instrument.maxNumberOfRuns"]
+        try:
+            allRuns = IntArrayProperty(name="RunList", values=runs).value
+        except Exception as e:
+            raise e
+        validRuns = []
+        for run in allRuns:
+            if RunNumberValidator.validateRunNumber(str(run)):
+                validRuns.append(run)
+        if len(validRuns) > maxRuns:
+            logger.warning(f"Maximum vale of {maxRuns} run numbers exceeded")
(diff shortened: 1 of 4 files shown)


src/snapred/backend/service/MetadataLookupService.py
```diff
@@ -0,0 +1,44 @@
...
+            allRuns = IntArrayProperty(name="RunList", values=runs).value
+        except Exception as e:
+            raise e
+        validRuns = []
+        for run in allRuns:
+            if RunNumberValidator.validateRunNumber(str(run)):
+                validRuns.append(run)
+        if len(validRuns) > maxRuns:
+            logger.warning(f"Maximum vale of {maxRuns} run numbers exceeded")
+            validRuns = []
```
rboston628, line 42: Are you sure you want it to return an empty list?

rboston628, line 42: Are you sure you want it to return an empty list?

tests/unit/backend/service/test_MetadataLookupService.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class TestMetadataLookupService(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.instance = MetadataLookupService()
+        super().setUpClass()
+
+    def setUp(self):
+        return super().setUp()
```
rboston628, line 14: If this only calls the super() method then it is not needed.

src/snapred/backend/service/MetadataLookupService.py
```diff
@@ -0,0 +1,44 @@
...
+    def name():
+        return "verifiedRuns"
+
+    @FromString
+    def verifyMultipleRuns(self, runs: str):
+        maxRuns = Config["instrument.maxNumberOfRuns"]
+        try:
+            allRuns = IntArrayProperty(name="RunList", values=runs).value
+        except Exception as e:
+            raise e
```
rboston628, line 35: Is this try/catch necessary here?

src/snapred/backend/service/MetadataLookupService.py
rboston628: Wouldn't it be better to have the for loop limit to 10, and just return the first ten runs?

tests/unit/backend/service/test_MetadataLookupService.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def test_verifyMultipleRuns(self):
```
rboston628, line 16: You need a test to cover the scenario that `IntArrayProperty` can parse the string to a list of run numbers, but the list contains one valid and one invalid run number, and make sure only the valid run is returned.

tests/unit/backend/service/test_MetadataLookupService.py
```diff
@@ -0,0 +1,27 @@
...
+        # Test valid number of runs
+        runs = self.instance.verifyMultipleRuns("46342:46352")
+        assert len(runs) == 10
+
+        # Test exceeding max number of runs
+        runs = self.instance.verifyMultipleRuns("46342:46355")
+        assert len(runs) == 0
```
rboston628, line 23: You can capture and verify the warning message with [`assertLogs`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertLogs)

rboston628: Near the top you need to import the service module specifically with
``` python
ServiceModule = importlib.import_module(MetadataLookupService.__module__)
```
then you can verify the log with
``` python
        with self.assertLogs(logger=ServiceModule.logger, level=logging.WARNING) as cm:
            runs = self.instance.verifyMultipleRuns("46342:46355")
        assert f"whatever the message is" in cm.output[0]
```

src/snapred/backend/service/MetadataLookupService.py
dlcaballero16: So I thought about that, but I wasn't sure if it was better to return the first 10 runs or last 10 runs. I decided that if there's more than 10 runs, then the input is wrong, and the output should be empty so that the user has to try again.

src/snapred/backend/service/MetadataLookupService.py
dlcaballero16: I guess I thought it was, but it seems like IntArrayProperty will throw an exception internally if the input is bad

src/snapred/backend/service/MetadataLookupService.py
rboston628: Would raising a recoverable error work better, then?

dlcaballero16: I think this should? We normally have a box popup whenever we log a warning

rboston628 (approved):
This does not connect to anything yet, so there it should not impact the backend.  Manual testing in the GUI show it runs to completion and saves in calibration and normalization.  The unit tests cover the cases I could think of.

PR #383, opened 2024-06-24
Diagnose failing workspace compare tests

rboston628:
## Description of work

In PR #327, after adding calls to properly compare workspaces, it was discovered that three tests were failing.  These were marked as `xfail` and work to actually fix them delayed for EWM 5043.

Exactly how to fix them relies on diagnosing what is wrong.  This PR diagnosed some problems, fixed two, then found two more.

### `test_LoadGroupingDefinition`

Two tests were failing.  As it happened, both of these were tests of loading the instrument from a `.hdf` file.  I was able to verify, the issue was with the `.hdf` file being loaded, which had the incorrect groupings, as seen below when compared to the xml and nxs grouping workspaces.

Screenshot from 2024-06-24 10-41-41

Fixing the `.hdf` file was pretty easy at this stage, so I replaced it in this PR.

While working on this, I uncommented the tests of loading from the instrument name (e.g. "SNAP" or "SANS").  These were commented-out with the reason being they were "slow".  However, if uncommented they also fail, because they were comparing the full SNAP instrument to the 16-pixel test instrument.

Fixing these ~~should be possible (allegedly)~~ is possible by using mantid's [`amend_config`](https://docs.mantidproject.org/nightly/api/python/mantid/kernel/AmendConfig.html#amend-config) to redirect the [`instrumentDefinition.directory` property](https://docs.mantidproject.org/nightly/concepts/PropertiesFile.html#directory-properties).  ~~I have written an outline of how that might work, but fixing it will be outside the scope of this story.~~  I was able to accomplish this after much trial-and-error.

### `test_DiffractionSpectrumWeightCalculator`

This problem is more difficult to solve, but seems to be due to incorrect peak list in the test.

The test loads a workspace with spectrum peaks, uses the algorithm to calculate the smoothing weights for this workspace, then compares the weight workspace to a workspace of expected results.

Firstly, it is only calculating on a single spectrum.  This is due to the definition of `peaks_json` on L74, which only has a single group.  See below: spectrum 1 looks like a flat line around 0 or 1, but the other spectra are unchanged.
Screenshot from 2024-06-24 12-18-27

Secondly, the weights it is calculating are incorrect.

The expected weights, compared to the spectra peaks, looks reasonable.  They show 0 values underneath some major peaks.
Screenshot from 2024-06-24 12-23-29

The output from the algorithm does not match, and in fact is setting everything to 1.
Screenshot from 2024-06-24 12-21-48

Looking at [the git blame](https://github.com/neutrons/SNAPRed/blame/next/tests/unit/backend/recipe/algorithm/test_DiffractionSpectrumWeightCalculator.py#L65), this was changed two months ago to point to synthetic data.  Before this, it had pointed to a different file.

Loading that file instead of the synthetic data processes all spectra, and is much closer to aligning with the expected value.  The difference is in the low-d-space peaks, which should be filtered out somehow.
image

This suggests there is not any issue with the algorithm itself, but with the input data to it.

I still suggest creating several more tests, and preferably tests not requiring simpler inputs.

### Left for follow-up

- [ ] Fix the problem with the diffraction spectrum weight calculator.  This might require editing and inspecting the algorithm itself.
- [ ] Everywhere present, remove reference to "fakeSNAP.xml" and replace with "fakeSNAP_Definition.xml"
- [ ] Delete "fakeSNAP.xml", delete the extended note in the "fakeSNAP_Definition.xml" file.

### Dev testing

Just make sure the unit tests pass, including the two I fixed.  Make sure the other two I identified are x-failing.

### CIS testing

Not needed, internal issue.

## Link to EWM item

[EWM#5762](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5762)


tests/resources/inputs/testInstrument/facilities.xml
rboston628, line 1: According to [mantid documentation](https://docs.mantidproject.org/nightly/concepts/FacilitiesFile.html#facilities-file), having this file is supposed to enable `amend_config` to find our test instrument definition file.

tests/resources/inputs/testInstrument/fakeSNAPFocGroup_Natural.hdf
rboston628, line 1: This file had invalid grouping numbers.  It has been fixed, so that the original failing two tests now pass.

tests/unit/backend/recipe/algorithm/test_LoadGroupingDefinition.py
```diff
@@ -302,35 +308,35 @@ def test_load_from_xml_file_with_instrument_donor(self):
     def test_load_from_xml_file_with_instrument_file(self):
         self.do_test_load_with_instrument_file("xml")
 
-    # NOTE commented out because slow
-    # def test_load_from_xml_file_with_instrument_name(self):
-    #     self.do_test_load_with_instrument_name("xml")
+    # TODO THIS IS BAD -- EWM 5043
+    @pytest.mark.xfail(reason="Cannot load test instrument from name", strict=True)
+    def test_load_from_xml_file_with_instrument_name(self):
+        self.do_test_load_with_instrument_name("xml")
```
rboston628, line 314: This was a newly-discovered error.  It fails because mantid cannot find the test instrument file by its name.  Or, if using "SNAP" as the name, because SNAP has far more pixels than the test instrument.

tests/unit/backend/recipe/algorithm/test_LoadGroupingDefinition.py
```diff
@@ -302,35 +308,35 @@ def test_load_from_xml_file_with_instrument_donor(self):
...
 
     def test_load_from_nxs_file_with_instrument_file(self):
         self.do_test_load_with_instrument_file("nxs")
 
-    # NOTE commented out because slow
-    # def test_load_from_nxs_file_with_instrument_name(self):
-    #     self.do_test_load_with_instrument_name("nxs")
+    # NOTE this one passes because nexus uses the workspace's instrument
+    def test_load_from_nxs_file_with_instrument_name(self):
+        self.do_test_load_with_instrument_name("nxs")
```
rboston628, line 339: This is not slow, because it does not actually load the instrument by its name.

tests/util/helpers.py
```diff
@@ -192,18 +191,15 @@ def workspacesEqual(Workspace1: str, Workspace2: str, **other_options):
...
-            raise AssertionError(message.column("Message"))
+    # NOTE this can be re-worked to call `assert_wksp_almost_equal` when that
+    # has been fixed to allow exact comparisons.
+    equal, message = CompareWorkspaces(
+        Workspace1=Workspace1,
+        Workspace2=Workspace2,
+        **other_options,
+    )
+    if not equal:
+        raise AssertionError(message.column("Message"))
```
rboston628, line 202: Following on a discussion with @ekapadi, the former method of using mantid's `assert_almost_equal` does not allow exact comparisons, and many tests should be using exact comparisons (see `test_GroceryService` for example; cloned workspaces should not be almsot equal, but exactly equal).  If the `assert_almost_equal` method is fixed to allow exact comparison, then this could be restored to its former glory.  Probably another defect should re-replace all of the `assert_wksp_almost_equal` with this method or a properly-written `CompareWorkspaces`

tests/resources/inputs/testInstrument/fakeSNAPFocGroup_Natural.nxs
rboston628, line 1: This file was updated so that its instrument name is now `fakeSNAP` instead of `fakeSNAP.xml`

tests/resources/inputs/testInstrument/fakeSNAPFocGroup_Natural.hdf
rboston628, line 1: This file was updated so that the groupings match the other files.

tests/unit/backend/recipe/algorithm/test_LoadGroupingDefinition.py
```diff
@@ -249,25 +251,33 @@ def do_test_load_with_instrument_file(self, ext: str):
...
+            ConfigService.setFacility("TestSNAP")
+            loadingAlgo = LoadingAlgo()
+            loadingAlgo.initialize()
+            loadingAlgo.mantidSnapper.cleanup = mock.Mock()
+            loadingAlgo.setProperty("GroupingFilename", self.localGroupingFile[ext])
+            loadingAlgo.setProperty("InstrumentName", "fakeSNAP")
+            loadingAlgo.setProperty("OutputWorkspace", outputWorkspace)
+            assert loadingAlgo.execute()
+        ConfigService.updateFacilities("")
+        ConfigService.reset()
```
rboston628, line 278: This enables mantid to find the test instrument by name, then undoes all the changes so the rest of mantid works.

darshdinger (approved):
All tests pass on branch as required.

PR #386, opened 2024-06-25
Fix broken test of diffraction spectrum weight calculator

rboston628:
## Description of work

From EWM 5043, there was one broken test inside `test_DiffractionSpectrumWeightCalculator.py`.

This one broken test was also the only test.

This adds many more tests, adds validation inside the algorithm to prevent the previous error, and fixes the test that was broken.

## Explanation of work

The test compared the output of the algo to a saved workspace of the expected values.  The problem is those expected values were generated using a list of detector peaks, which no longer resembled the file it had drawn from.

Instead, it had been set to use a list that only specified peaks on a single spectrum, and specified incorrect peaks.

This recreates the correct json file for the detector peaks and loads it to make that test work.

The previous broken test passed a workspace with 6 spectra and a list of `GroupPeakList`s with only a single element; this mismatch was an error.  A new validation check in the algo prevents this, and a new test verifies this is prevented.

Many more tests were also added for the various parts of the algorithm.

### Dev testing

Because this edited the algorithm, make sure both diffcal and normalization can run.  The important step is the tweak peak stage.  Make sure changing values there works as before.

### CIS testing

Not needed, this is an internal issue.

## Link to EWM item

[EWM#5043](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5043)


dlcaballero16 (approved):
Tested diffraction and normalization, was able to tweak parameters and save for both.

PR #390, opened 2024-06-27
Retain diagnostic workspace on iterate.

rboston628:
## Description of work

The diffcal process was recently changed to output a diagnostic workspace with information about the peak fits.

However, if the user elected to iterate, this workspace would be deleted, and could create errors if the user tried to save this older iteration.

The issue was, while the diagnostic workspace group itself was being retained, all of the constituents were deleted, and this signaled to mantid to delete the group as well.

## Explanation of work

The simplest and most elegant solution was to fix `clearADS` inside `GroceryService`, so that if a workspace in the exclude list is also a workspace group, all of the constituents are also excluded.

That was achieved as follows:
``` python
        workspacesToClear = set(mtd.getObjectNames())
        # filter exclude
        workspacesToClear = workspacesToClear - set(exclude)
        # properly handle workspace groups -- also exclude deleting their constituents
        for ws in exclude:
            if mtd[ws].isGroup():
                workspacesToClear = workspacesToClear - set(mtd[ws].getNames())
```
I used sets because it seemed less likely to get in a situation of double-deleting a workspace name from `workspacesToClear`.

### Dev testing

1. **reproduce the error**.  On `next`, run diffcal to assessment stage.  Observe the light-blue workspace group, click the arrow next to it, observe the four workspaces inside it.  Now click to iterate.  It's gone!  Go back through to save and try to save iteration 1.  The program will fail.
2. Switch to this branch, and reproduce the above behavior.  When iterating, the workspace will remain!
3. Iterate a few times for good measure.
4. Save iteration 1.  It should complete fine.
5. After save, continue.  Make sure the *entire* workspace list is cleared.
6. Check that the workspaces saved in the corresponding version folder, and that there is no iteration counter on the workspace names.
7. Run normalization and make sure it saves.

### CIS testing

As above, but use a real run like 58882.

Make sure the workspaces inside the diagnostic workspace group are valuable and make sense.

## Link to EWM item

[EWM#5668](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5668)

### DIFF
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -1031,9 +1031,14 @@ class GroceryService:
         :param cache: whether or not to clear cached workspaces (True = yes, clear the cache), optional (defaults to False)
         :type cache: bool
         """  # noqa E501
-        workspacesToClear = mtd.getObjectNames()
+        workspacesToClear = set(mtd.getObjectNames())
         # filter exclude
-        workspacesToClear = [w for w in workspacesToClear if w not in exclude]
+        workspacesToClear = workspacesToClear - set(exclude)
+        # properly handle workspace groups -- also exclude deleting their constituents
+        for ws in exclude:
+            if mtd[ws].isGroup():
+                workspacesToClear = workspacesToClear - set(mtd[ws].getNames())
+
         # filter caches
         if not cache:
             workspaceCache = self.getCachedWorkspaces()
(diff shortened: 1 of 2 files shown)


dlcaballero16 (approved):
Tested as described and was able to verify the fix works.

PR #396, opened 2024-07-03
Remove double difference in RawVanadiumCorrectionAlgorithm

rboston628:
## Description of work

Malcolm pointed out an apparent scribal error, where the command to subtract the background was repeated inside the algorithm.

## Explanation of work

Deleted the second copy.

### Dev testing

Not needed.

### CIS testing
Ensure this is what you want to do.

## Link to EWM item


mguthriem (approved):
correctly removes the double subtraction that leads to the wrong output.

PR #397, opened 2024-07-03
Add a test to prevent future double-subtraction in raw vanadium

rboston628:
## Description of work

Malcolm noted an error where `Minus` was being erroneously called twice, which was fixed in PR #396.

The unit test for this was already setup in such a way, it would have prevented the issue.

This just adds two checks onto the unit test:
1. the "raw vanadium" should always be positive
2. the "raw vanadium" should have its peak in the same spot as the "signal" workspace

## Explanation of work

The test and accompanying script were cleaned up and refactored a bit.

The test works by:
1. creating a "signal" workspace (which is not at all realistic)
2. creating a "background" workspace
3. adding the two together to create the "sample" workspace, mimicking what an experimentalist would measure

The raw vanadium algorithm is meant to extract out the "background" part and leave only the "signal" part, suitably renormalized.

Therefore the result should be nonnegative, and have its peaks where the signal has its peak.

### Dev testing

Run the CIS script (it will run locally, and takes under a second) and look at the graph, and verify the output (red) is nonnegative, and has a single peak at the "signal"'s peak.

Check the unit test, make sure it is checking this.

Then make sure the test runs.

### CIS testing

Not needed.

## Link to EWM item

[EWM#5938](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5938)

### DIFF
--- a/tests/cis_tests/lazy_raw_vanadium.py
+++ b/tests/cis_tests/lazy_raw_vanadium.py
@@ -34,10 +34,46 @@ sys.path.insert(0, str(Path(SNAPRed_module_root).parent / 'tests'))
 from util.SculleryBoy import SculleryBoy
 
 from snapred.meta.Config import Config, Resource
-Config._config['cis_mode'] = True
+Config._config['cis_mode'] = False
 Resource._resourcesPath = os.path.expanduser("~/SNAPRed/tests/resources/")
 
 class TestRawVanadiumCorrection(unittest.TestCase):
+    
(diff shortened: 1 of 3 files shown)


darshdinger (approved):
All unit tests pass and look good. The CIS test `lazy_raw_vanadium` also successfully runs and prints a figure:
image

PR #398, opened 2024-07-08
Ewm5912 user pixel mask orchestration

darshdinger:
## Description of work
This work is to facilitate the GUI orchestration of the user pixel masks. Within this PR, all the necessary connections are created to allow this feature to be functional and give access to the user.

## Explanation of work
One of the hardest portions of this work was the create a custom multi-select drop down box that allows the user to inspect all of the available pixel masks and also select multiple masks as stated per the story requirements. `ReductionView.py` was updated to inherent classes from `BackendRequestView` all of the other workflow views do. This was updated along with the integration of the multi-select drop down widget called `multiSelectDropDown`. 

`ReductionWorkflow` is also updated partially including Kort's code from his Pixel Mask PR and including some updates I made to handle to connect the new dropDown and retrieve the dictionary of workspace strings and workspace names.

### Dev testing
Testing was done manually, shown to Kort. It is functional and the reduction workflow completed successfully. 

```python
# NOTE: The LOAD button within SNAPRed main UI does not recognize diffcal workspaces. (DEFECT)
``` 

### CIS testing
Please reach out to me for notes!

## Link to EWM item

[EWM # 5912](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5912)


src/snapred/ui/view/BackendRequestView.py
```diff
@@ -28,5 +29,8 @@ def _labeledCheckBox(self, label):
     def _sampleDropDown(self, label, items=[]):
         return SampleDropDown(label, items, self)
 
+    def _multiSelectDropDown(self, label, items=[]):
```
ekapadi, line 32: **A key question here:  do _all_ `BackendRequestView` actually _have_ a multi-select dropdown?**  If so, is it always the "pixel masks" dropdown?  If not, this possibly should be moved to `ReductionRequestView`.

src/snapred/ui/view/reduction/ReductionView.py
ekapadi, line 1: I suggest that we change this name to `ReductionRequestView` in preparation for (almost certainly) adding more features to the reduction workflow panels.

src/snapred/ui/view/reduction/ReductionView.py
```diff
@@ -1,20 +1,22 @@
...
 
+logger = snapredLogger.getLogger(__name__)
+
 
 @Resettable
-class ReductionView(QWidget):
+class ReductionView(BackendRequestView):
     signalRemoveRunNumber = Signal(int)
 
     def __init__(self, pixelMasks=[], parent=None):
```
ekapadi, line 15: `__init__` doesn't have a `pixelMasks` arg anymore.

src/snapred/ui/widget/MultiSelectDropDown.py
```diff
@@ -0,0 +1,74 @@
...
+        self.setModel(QStandardItemModel(self))
+
+    def handleItemPressed(self, index):
+        item = self.model().itemFromIndex(index)
+        if item.checkState() == Qt.Checked:
+            item.setCheckState(Qt.Unchecked)
+        else:
+            item.setCheckState(Qt.Checked)
+
+    def checkedItems(self):
```
ekapadi, line 19: Please do a pass and `annotate` a few of these methods, or at _least_ this one.  For example, I think this is `checkedItems(self): -> List[QString]` which really helps see what's precisely going on.  (No need to annotate _everything_, but return values help.  :)  )

ekapadi (commented):
Only a few minor issues.  I'll be glad to hit "approve" once you've rebased it again (after the backend merge).

src/snapred/ui/view/BackendRequestView.py
darshdinger: I left this integration here because all of our views are derived from the `BackendRequestView`. Previously, `ReducitonRequestView` was made into a QtWidget due to a the lack of the existence of `ReductionService` when `ReductionRequestView` was created. In order to merge, I needed a working model of this view and due to the lack of service backend calls to retrieve information, I couldn't follow the "traditional" SNAPRed pattern of implementing the view.

By 'traditional', I mean specifically this:
```python
class SomeView(BackendRequestView):
   def __init__(self, some arguments taken from the workflow (which are produced by a service call within workflow), parent=None):
# etc.... 
``` 

With the service created, I decided to revert this view to fit the 'standard' model views follow within SNAPRed and adding the functionality of the multi-select drop down GUI widget to `BackendRequestView` to expose this functionality to other potential views later in the future.

src/snapred/ui/view/reduction/ReductionView.py
darshdinger: Done!

src/snapred/ui/view/reduction/ReductionView.py
darshdinger: Why not? Should we follow the pattern or move away from it within Reduction? I'm genially unsure about the best practice here. Hopefully you have some idea?

src/snapred/ui/widget/MultiSelectDropDown.py
darshdinger: Done!

src/snapred/ui/view/reduction/ReductionView.py
darshdinger: Done!

src/snapred/ui/view/reduction/ReductionView.py
darshdinger: Okay! This has been fixed!

ekapadi (approved):
Everything looks good.  All panels have been tested and work as they should.

PR #401, opened 2024-07-11
Configure mypy so it can pass

peterfpeterson:
## Description of work

This configures mypy and fixes a couple of minor issues. Unfortunately, I was unable to convince the build server to use the configuration file.

The intent is to start here and slowly remove ignored directories until everything passes. At that point the mypy pre-commit hook should work similar to [here](https://github.com/pre-commit/pre-commit/blob/main/.pre-commit-config.yaml#L39-L44).

## Dev testing

```sh
python -m mypy .
```
Should run without issue

### CIS testing

Not applicable

## Link to EWM item

[EWM6080](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6080)

This is an alternative attempt at #83.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -41,10 +41,10 @@ jobs:
       with:
         fail_ci_if_error: true
         token: ${{ secrets.CODECOV_TOKEN }} # required
-    - name: mypy checks
-      shell: bash -l {0}
-      run: |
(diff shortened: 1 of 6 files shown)


rboston628 (approved):
I confirmed running `python -m mypy .` does return no errors.  I also confirmed the changed import does not impact the operation of the stand-alone GUI.

PR #402, opened 2024-07-11
Added ability to keep unfocused data for reduction

dlcaballero16:
## Description of work

User can now keep unfocused data during Reduction and can convert workspace to user specified units.

## Explanation of work

The reduction frontend was modified so that the Keep Unfocused Data checkbox is enabled and connected to the backend.  There is also a dropdown box so that the user can select what the units of the workspace should be in.

For the backend, 2 new fields were added to the ReductionIngredients: `keepUnfocused` and `convertUnitsTo`. The first is a bool and the second is a string that determines what the units should be. The ReductionRecipe was updated to process these new variables and make sure the intended outcome happens.

### Dev testing

1. Make sure you have a run that you are able to run Reduction on. 
2. On the Reduction tab, enter a run number. Make sure the `ConvertUnits` dropdown is still disabled. 
3. Now click the `Retain Unfocused Data` box. The dropdown should be enabled. 
4. Try to continue. You should see an error message saying to select units. 
5. Select TOF from the drop and click continue. 
6. When reduction finishes, look at the main SNAPRed window that has all the workspaces. You should now see several workspaces that start `output_tof_unfoc_<run number>...`.  Verify those workspaces have units of TOF. 
7. Rerun reduction but make sure `Retain Unfocused Data` is not checked this time. Make sure those workspaces do not exist anymore. 
8. To be really thorough, verify all units work. Here's what the dropdown should match to:

TOF                         = tof
dSpacing                 = dsp
Wavelength             = lam
MomentumTransfer = qsp

### CIS testing

Do the above, but verify the unfocused workspaces is the correct data that you are expecting.

## Link to EWM item

[EWM#793](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=793)

### DIFF
--- a/src/snapred/backend/dao/ingredients/ReductionIngredients.py
+++ b/src/snapred/backend/dao/ingredients/ReductionIngredients.py
@@ -26,6 +26,9 @@ class ReductionIngredients(BaseModel):
     calibrantSamplePath: str
     peakIntensityThreshold: float
 
+    keepUnfocused: bool
+    convertUnitsTo: str
+
     #
     # FACTORY methods to create sub-recipe ingredients:
     #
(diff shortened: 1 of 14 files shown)


darshdinger (approved):
Looks good. Tested and functional based on testing criteria.

PR #403, opened 2024-07-15
Reduction pixel-mask I/O

ekapadi:
## Description of work

When setting up a reduction process, users may optionally specify the construction of a _combined_ reduction pixel mask, by selecting from a list of pixel masks from previous reductions in the same state, and also from any _compatible_ "by hand" pixel masks that they've created using the Mantid workbench.  This PR deals with the requirements needed to guarantee that any masks selected are _compatible_ with the planned reduction process, namely that they are from the same instrument state, and use the same lite-mode setting.  This PR also implements several ancillary features required to perform reduction using pixel masks, such as checking a specified workspace to determine its associated instrument state, and _naming_ the combined pixel mask using the reduction _timestamp_.

## Explanation of work

  * Calculate the State SHA from the workspace itself.

  * 'getCompatibleReductionMasks': include masks from previous compatible reductions, and any compatible user-generated pixel masks.

  * Rework 'util.helpers.createCompatibleMask' so that it doesn't require an 'instrumentFilename' argument.

  * Reduction pixel-mask load and save methods.

  * Add 'builder' attribute to 'WorkspaceName'; add 'tokens' method to retrieve token values.

  * WNG changes: case-sensitive literal tokens.

  * WNG.NameBuilder: add 'wsType' transparently to 'keys' and 'props'.

  * 'FarmFreshIngredients.FocusGroups' -- implemented as a list, but allow reference as a singular "focusGroup" attribute.

  * Reduction data I/O: use timestamp instead of version.

  * Timestamp format: support for ISO formats (, but also support SNAPRed's legacy "milliseconds" format).  In order to support already existing DAO (i.e. on the filesystem): for these DAO only, allow timestamp to be read either as a `float` (current), or as an `int` (legacy) encoding.

### Dev testing

Unit tests have been implemented to test these _back end_ changes.

### CIS testing

It is easiest to test these changes using the associated _front end_ implementation, which will follow this PR _immediately_.

## Link to EWM item

[EWM#4801](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=4801)


darshdinger (approved):
Fully tested and working. Looks good! :)

PR #295, opened 2024-03-28
Standardize diagnostic workspaces within DiffCal workflow

rboston628:
## Description of work

This is another enabler for Story 4580.

This is to fix the output in two of our algorithms.

### `FitMultiplePeaksAlgorithm`

Each spectrum is extracted and has peaks fit separately.  Each call to `FitPeaks` produces a new grouped workspace for that spectrum, containing diagnostic information.

#### Currently
All of these workspaces are grouped together under a new super-group workspace, as below:
image

#### Aim of this PR: 
- Combine all of the `_fitted_X` workspaces into one workspace with many spectra.
- Combine all of the `_fitted_params_X` workspaces into one single table workspace (use the `ConjoinTableWorkspaces` algorithm in this PR)
- Combine all of the `_fitted_peakpositions_X` workspaces into one workspace with many spectra
- Combine all of the `_fitted_params_err_X` workspaces into one single  table workspace

That is, at the end of running `FitPeaks`, there will be a grouped workspace called `fit_peaks_diag_XXXXX_Y` containin
- `fit_peaks_diag_XXXXX_Y_fitted_peakpositions` with peak positions in all spectra
- `fit_peaks_diag_XXXXX_Y_fitted_params` with the parameters for all peaks in all spectra
- `fit_peaks_diag_XXXXX_Y_fitted` with the fit peaks in all spectra
- `fit_peaks_diag_XXXXX_Y_fitted_params_err_0` with the errors of all peaks in all spectra
- 
This will likely require further edits to `CalibrationMetricExtractionAlgorithm` to make sure it still works with calls to  `FitPeaks`

### `GroupDiffractionCalibration`

Each spectrum is extracted and passed separately to `PDCalibration`.  Each call to `PDCalibration` produces a new grouped workspace for that spectrum, containing diagnostic information.

#### Currently
All of the fitted peaks are combined into a single workspace called `_PDCal_diag`, which has the fit peaks in each spectrum (so if using Column, has 6 spectra of fit peaks).  This workspace is a "secret" output, in that it is produced by its existence is not documented in the algorithm's API.  

**Also currently** the diffraction-focsed TOF data is used as a separate output.  Its only purpose is to match against the fit peak group returned by `PDCalibration`, which applies the fits in TOF.

#### Aim of this PR

Replace the `OutputWorkspaceTOF` property with a `DiagnosticWorkspace` property deriving from `WorkspaceGroup` (see `FitPeaks` for a starting point).

The diagnostic workspace be a grouped workspace that will contain:
[...]

PR #412, opened 2024-07-19
Properly iterate diagnostics :repeat_one: 

rboston628:
## Description of work

In PR #295, the diffcal workflow was refactored to output a `WorkspaceGroup` containign diagnostic information from the peak fitting that takes place inside `PDCalibration`.  This being a `WorkspaceGroup`, and not simply a `Workspace`, has led to a few unexpected behaviors when handling it.

One unexpected behavior was solved in PR #390, where child workspaces were being accidentally deleted and needed to be added to a retain list.

In this case, child workspaces are not being renamed when the parent is.  This means the diagnostic `WorkspaceGroup`s for each iteration all contain references to the same set of workspaces, as opposed to the workspaces made in that iteration.

The work is to add a new endpoint which more effectively handles renaming workspaces on an iteration.

## Explanation of work

Adding the new endpoint inside the `WorkspaceService`.

This required a new DAO request object.

This could be handled with a single call to `RenameWorkspaces` for the whole list of workspaces, so a new method was added to `GroceryService`.

### Dev testing

Check the unit tests and make sure they cover the situations meant to be fixed here.

Run diffcal.  At the assess screen, open the diagnostic workspace group (it has a light blue icon) and make sure the names have no iteration label on them.

Click iterate.  Now the parent and all children should have a `_1` on the end.

Iterate a few times until you are satisfied that the workspaces are being renamed.

### CIS testing

Do as above, but using differences in each calibration, and make sure the values in each iteration are different.

## Link to EWM item

[EWM#6018](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6018)

### DIFF
--- a/tests/util_tests/test_state_helpers.py
+++ b/tests/util_tests/test_state_helpers.py
@@ -90,12 +90,17 @@ def initPVFileMock() -> mock.Mock:
     return mock_
 
 
+@mock.patch.object(LocalDataService, "_writeDefaultDiffCalTable")
 @mock.patch.object(LocalDataService, "_generateStateId")
 @mock.patch.object(LocalDataService, "_defaultGroupingMapPath")
 @mock.patch.object(LocalDataService, "readInstrumentConfig")
 @mock.patch.object(LocalDataService, "_readPVFile")
 def test_state_root_override_enter(
-    mockReadPVFile, mockReadInstrumentConfig, mockDefaultGroupingMapPath, mockGenerateStateId
+    mockReadPVFile,
+    mockReadInstrumentConfig,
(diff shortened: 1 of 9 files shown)


tests/util_tests/test_state_helpers.py
rboston628, line 1: These mocks were added on `next` but not in staging.  I think this test actually calculates DIFC values for the entire SNAP instrument if these mocks are not added.

dlcaballero16 (approved):
Tested and functions as expected.

PR #413, opened 2024-07-19
Fix Diffcal  Data Defect

dlcaballero16:
## Description of work

This is to fix the Diffcal defect of having missing data and a loader not being available.

## Explanation of work

WrapLeftovers was updated to just call ResampleX and SaveNexus instead of saving everything to a tar file. ReheatLeftovers was also updated to just use LoadNexus instead of trying to open a tar file. Anywhere else in the code that was using a tar file was updated to use a nxs.h5 file instead.

### Dev testing

Either locally or on analysis, run any diffcal you want and save it. In the output folder, make sure there is no tar file anymore. The tar file should now be replaced by a file that looks like dsp_<grouping>_<runNumber>_<version>.nxs.h5.

### CIS testing

Run diffcal with Column grouping and complete it and save. Check the output folder and make sure the output data is what you expect for Column grouping. The output file will look like dsp_<grouping>_<runNumber>_<version>.nxs.h5.

## Link to EWM item

[EWM#6098](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6098)


tests/unit/backend/data/test_GroceryService.py
```diff
@@ -1047,6 +1059,11 @@ def test_fetch_grocery_list_diffcal_output(self):
                 .buildList()
             )
             diffCalOutputFilename = self.instance._createDiffcalOutputWorkspaceFilename(groceryList[0])
+            wrapLeftovers = WrapLeftovers()
+            wrapLeftovers.initialize()
+            wrapLeftovers.setPropertyValue("InputWorkspace", self.sampleWS)
+            wrapLeftovers.setPropertyValue("Filename", self.sampleTarWsFilePath)
+            wrapLeftovers.execute()
```
rboston628, line 1066: Could you call here, and above, using the function syntax?
``` python
WrapLeftovers(
    InputWorkspace=self.sampleWS,
    Filename=self.sampleTarWsFilePAth,
)
```
You will need to change the import above, to
``` python
import snapred.backend.recipe.algorithm
from mantid.simpleapi import (
    ...
   the stuff already there,
   ...
   WrapLeftovers,
   ReheatLeftovers,
)
```

tests/unit/backend/recipe/algorithm/data/test_Leftovers.py
```diff
@@ -64,6 +51,14 @@ def test_saveLoad():
         reheatLeftovers.setPropertyValue("OutputWorkspace", "reheated")
         reheatLeftovers.setPropertyValue("Filename", filename)
         reheatLeftovers.execute()
-        compareRaggedWorkspaces(mtd["raw"], mtd["reheated"])
-        DeleteWorkspace("raw")
-        DeleteWorkspace("reheated")
+
```
rboston628, line 54: Same here, replace these calls with the function syntax

rboston628 (approved):
All tests run. Manually ran diffcal, normalization, and reduction using run 46680.  Most importantly, checked that loading previous versions works as expected.  Below is v0001 loaded back in, and looks to have loaded correctly.
image

src/snapred/backend/data/LocalDataService.py
```diff
@@ -460,7 +460,7 @@ def writeNormalizationWorkspaces(self, record: NormalizationRecord):
         for workspace in record.workspaceNames:
             ws = mtd[workspace]
             if ws.isRaggedWorkspace():
-                filename = Path(workspace + ".tar")
+                filename = Path(workspace + ".nxs.h5")
```
rboston628, line 463: This single line is the cause for the codecov complaint.  I don't think it's worth worrying over.

PR #414, opened 2024-07-22
Ewm6118 change pvs

darshdinger:
## Description of work
This was a simple fix request by Malcolm. The logic change include operations to first check for `BL3:Chop:Gbl:WavelengthReq` and if it is not found, then check for `BL3:Chop:Skf1:WavelengthUserReq` before throwing an error.

## Explanation of work
The description contains the work done. Other than that, this task was fixing tests heavy.

### Dev testing
Insure all pytests pass and that any given workflow functions properly.

### CIS testing
Same as above.

## Link to EWM item

[EWM # 6118](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6118)


dlcaballero16 (dismissed):

mguthriem:
Can this also be merged into staging?

rboston628 (changes requested):

src/snapred/backend/data/LocalDataService.py
```diff
@@ -731,16 +731,28 @@ def writeNormalizationState(self, normalization: Normalization):
     def readDetectorState(self, runId: str) -> DetectorState:
         detectorState = None
         pvFile = self._readPVFile(runId)
+        wav_value = None
+        wav_key_1 = "entry/DASlogs/BL3:Chop:Gbl:WavelengthReq/value"
+        wav_key_2 = "entry/DASlogs/BL3:Chop:Skf1:WavelengthUserReq/value"
+
```
mguthriem, line 737: @darshdinger I added these lines to the version of next I'm testing pixel masking in. I think the code above is working (the state is identified from the input file) but I get an error because there is an additional function `detectorStateFromWorkspace` that the pixel masking is calling. This is also in `LocalDataService` L762 ish. This function has to be updated in a similar way to `readDetectorState`

rboston628:
This work was moved to PR #418

PR #426, opened 2024-07-30
Retain reduction-workflow output workspaces

ekapadi:
# PR from "staging" (PR#416) onto "next"

## Description of work

This PR includes changes added to facilitate `qtbot`-based testing, **which also happen to satisfy the requirements of this "3.0.1 defect" regarding retention of reduction-workflow output workspaces**.  The changes of this commit are relatively minor, and will be merged to "next" almost immediately.  For this reason, no attempt has been made to separate out the "staging" changes specific to the present defect.

## Explanation of work

This PR includes the following changes:

  * Identify ADS-resident workspaces before the start of each workflow, and do not delete these workspaces at exit.  This allows users to go back-and-forth between the SNAPRed panel, and the Mantid workbench, and not to have their workspaces unexpectedly deleted by SNAPRed.  In addition, simply extending this workspace list where required allowed retention of final result workspaces from the reduction workflow.

  * Lock out other workflow tabs once a workflow has started; 
  
  * Use `@Slot` decorator where appropriate, throughout the codebase.  This allows `pyqt` to check `Signal` argument types, but in addition, it also allows clear identification of which methods are being employed as slots;

  * Add missing `QObject` derivation for several classes with methods that are used as Qt slots.
    
  * Add several new signals specifically to help during testing: these include a signal at the start of each workflow, and a signal when any `Worker` task completes.  By default, connecting these `Signal` to `lambda *args: None` allows `qtbot` to intercept them when required;
  
  * All `QMessageBox` and `ActionPrompt`-style dialogs, including `SuccessPrompt` (formerly `SuccessDialog`) are now called only via static `warning`, `critical`, and `prompt` methods.  This allows these dialogs to be easily mocked without requiring mocking their entire classes;
  
  * Remove 'WorkflowNodePresenter' as _dead_ code.  This removal is in an _interim_ state, now marked by retaining the source file, but renaming it to `WorkflowNodePresenter_.py`.  The reason for this approach is that most probably we _do_ want to distinguish between `WorkflowPresenter` and `WorkflowNodePresenter`, and although it's not being used at present, this issue should probably be fixed.

### Dev testing

Existing unit tests cover these changes, with a few minor modifications.
In addition, the best way to test these changes is to completely run the diffraction-calibration, normalization, and reduction workflow tabs, and observe that the _other_ workflow tabs are disabled once a given workflow has been started.  **Additionally observe that the reduction output-workspaces are retained at the end of the complete process.** 

### CIS testing
Same as above, for Dev testing.

## Link to EWM item

[EWM#6093](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6093)

### DIFF
--- a/docs/source/api/snapred.backend.dao.request.rst
+++ b/docs/source/api/snapred.backend.dao.request.rst
@@ -36,10 +36,18 @@ snapred.backend.dao.request.CalibrationLoadAssessmentRequest module
    :undoc-members:
    :show-inheritance:
 
-snapred.backend.dao.request.ClearWorkspaceRequest module
+snapred.backend.dao.request.ClearWorkspacesRequest module
+---------------------------------------------------------
+
+.. automodule:: snapred.backend.dao.request.ClearWorkspacesRequest
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+snapred.backend.dao.request.ListWorkspacesRequest module
 --------------------------------------------------------
 
-.. automodule:: snapred.backend.dao.request.ClearWorkspaceRequest
+.. automodule:: snapred.backend.dao.request.ListWorkspacesRequest
    :members:
    :undoc-members:
    :show-inheritance:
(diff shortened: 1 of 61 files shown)


walshmm:
confirmed to still work after some regression tests

walshmm (approved):

PR #418, opened 2024-07-26
Ewm6118

darshdinger:
## Description of work
This was a simple fix request by Malcolm. The logic change include operations to first check for `BL3:Chop:Gbl:WavelengthReq` and if it is not found, then check for `BL3:Chop:Skf1:WavelengthUserReq` before throwing an error.

## Explanation of work
The description contains the work done. Other than that, this task was fixing tests heavy.

### Dev testing
Insure all pytests pass and that any given workflow functions properly.

### CIS testing
Same as above.

## Link to EWM item

[EWM # 6118](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6118)

PR #429, opened 2024-07-31
Defect 6119

darshdinger:
## Description of work
This work was to address a defect encountered by Malcolm within the `readDetectorState()` method. The largest issues was the improper retrieval of metadata within the pv log files. 

## Explanation of work
The majority of this work was handled within work for EWM 6118, where logic was implemented to check for two different variations for `WavelengthUserReq` before throwing an exception. The time was also taken in this PR to implement some better exception handling within `StateValidationException` as it was found that in certain cases, the real exceptions were being masked behind incorrect pathing permissions issues. `StateValidationException` is made more robust to allow for more accurate error print out. Another similar method is added to `LocalDataService` and `DataExportService` to aid in the implementation of a tab lock out feature that conveys information to the User preemptively and keeps workflows inaccessible if the user does not have permissions to access data.

### Dev testing
For testing please follow these steps:
1. Start analysis
2. Rebase on `next`
3. git fetch `defect-6119`
4. WITH a modified `application.yml` launch SNAPRed and attempt running diffcal workflow with the following inputs:
    - Run Number: 61388
5. When hitting tab or clicking into a different cell to input more required information, you should notice this error:

image

This is normal as the changes within EWM 6118 within [PR 418](https://github.com/neutrons/SNAPRed/pull/418) have not yet been merged into next.

6. Now delete `61388` and enter `049999` within the run number cell and tab out and notice another error:

image

### CIS testing
Same as above.

## Link to EWM item

[EWM # 6119](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6119)


dlcaballero16 (dismissed):
Tested and works as expected

PR #432, opened 2024-08-01
Ewm6383 truncate edges

dlcaballero16:
## Description of work

This is the temp fix for bad edges when looking at the results for reduction.

## Explanation of work

In ApplyNormalizationRecipe, the dMin/Max values were adjusted for their given constants defined in application.yml.

### Dev testing

On next and on analysis, run DiffCal with run number 59039, grouping Column, the La11B6 sample, and peak intensity threshold 0.01. On the tweak peaks tab, change max chi sq to 2000. Save the run.

Run Normalization with run number 58808 and background run number 58812, and the same grouping and sample used in diffcal. On the next tab, adjust the smoothing parameter as best as you can so that the orange dashed line acts as a line of best fit to the blue line, this doesn't need to be exact. When you save the normalization, make sure the applies to field is set to `>=59039`.

Run reduction for 59039. Inspect the workspace `output_tof_all_lite_copy1_059039_Column` and plot all the spectra, should just be 6. The results should look something like this:

image (2)

Now switch to the new branch. You only need to rerun the reduction step. When you inspect the workspace, it should now look like this:

image (3)

The edges look better and now nothing should be negative.

### CIS testing

Same as dev testing.

## Link to EWM item

[EWM#6383](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6383)


src/snapred/backend/recipe/ApplyNormalizationRecipe.py
```diff
@@ -23,8 +23,11 @@ def chopIngredients(self, ingredients: Ingredients):
         We are mostly concerned about the drange for a ResampleX operation.
         """
         self.pixelGroup = ingredients.pixelGroup
-        self.dMin = self.pixelGroup.dMin()
-        self.dMax = self.pixelGroup.dMax()
+        # The adjustment below is a temp fix, will be permanently fixed in EWM 6262
+        dMin = [x + Config["constants.CropFactors.lowdSpacingCrop"] for x in self.pixelGroup.dMin()]
+        dMax = [x - Config["constants.CropFactors.highdSpacingCrop"] for x in self.pixelGroup.dMax()]
+        self.dMin = dMin
```
ekapadi, line 29: We need a test here to make sure that `dMax > dMin` after this change.  If it's not I'd then
`raise ValueError("d-spacing crop factors are too large -- resultant dMax must be > resultant dMin")`

ekapadi (commented):
Only a very minor change.  Please take a look at that and I'll run through the tests this afternoon.

ekapadi (approved):

ekapadi:
For some reason it didn't take my last comment:  I ran all tests as described, including the unit-test suite.  Everything works as it should.

PR #437, opened 2024-08-12
Ewm6552 fix incorrect error

darshdinger:
## Description of work
This work was to reproduce an error seen by Malcolm within the staging branch where if a normalization has not been performed on a run that is reduced, there is a confusing error that is displayed to the user on L199 in `SousChef` where SNAPRed is unable to retrieve a peak intensity threshold from a (non-existent) normalization record. This work is to fix the error as per request and change it to display something more meaningful like "Normalization Calibration has not been conducted for this state, normalization folder does not exist."

## Explanation of work
This took a bit of time in implementation, I tried looking at an implementation within `SousChef` within next to try and find ways to implement this fix within staging as a fix is present. After spending time trying to create something sensible with the 'old' way of handling records, I found it best to handle this fix in the simplest way possible to remove any added complexity. So the fix is a simple if elif statement for both normalization and calibration records.

### Dev testing
1. Insure that your local state folders are all deleted.
2. Please open up SNAPRed within dev env
3. Run a calibration for run 46680 (any inputs, just has to run all the way through and save)
4. Run Reduction for run 46680
5. This error should occur:
image

### CIS testing
Same as above.

## Link to EWM item

[EWM # 6552](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6552)

### DIFF
--- a/src/snapred/backend/service/SousChef.py
+++ b/src/snapred/backend/service/SousChef.py
@@ -185,6 +185,14 @@ class SousChef(Service):
         normalizationRecord = self.dataFactoryService.getNormalizationRecord(
             ingredients.runNumber, ingredients.useLiteMode, version
         )
+        if calibrationRecord is None:
+            raise ValueError(
+                f"No calibration record found for run {ingredients.runNumber}, please run a calibration for this run."
+            )
+        elif normalizationRecord is None:
+            raise ValueError(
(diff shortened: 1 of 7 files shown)


dlcaballero16 (dismissed):
Tested and error message pops up as expected.

dlcaballero16 (approved):

PR #438, opened 2024-08-12
`unittest.mock.Mock.called_once` does not exist

ekapadi:
Throughout the SNAPRed test framework, assertions have been repeatedly constructed as `assert <mock>.called_once()`, and analogously for `called_once_with`, `not_called`, `called_with`.  Unfortunately, none of these functions exist, and the associated mock returns a mock function, which has a truth-value of True, and the associated assertions all pass.

This commit includes the following changes (for 52 unit tests, some requiring multiple substitutions):

  * Replace all occurrences of `called_once`, `called_once_with`, `not_called`, and `called_with` with the corresponding `assert_called_once` (and etc.) variants.

  * Make any required fixes to the resulting unit tests.

  * Patch the `unittest.mock.Mock` class so that these `call_once` (and etc.) functions are now defined, but are defined in such a manner that any application raises an exception.  To this end, an abstract-syntax-tree (AST) solution was also explored successfully, but it was considered overkill to meet the requirements of this PR.  This AST-approach is something we might want to look into in the future, if we implement a "SNAPRed coding style" commit check.

## Description of work

Throughout the SNAPRed test framework, assertions have been repeatedly constructed as `assert <mock>.called_once()`, and analogously for `called_once_with`, `not_called`, `called_with`.  Unfortunately, none of these functions exist, and the associated mock returns a mock function, which has a truth-value of True, and the associated assertions all pass.

## Explanation of work

This commit includes the following changes (for 52 unit tests, some requiring multiple substitutions):

  * Replace all occurrences of `called_once`, `called_once_with`, `not_called`, and `called_with` with the corresponding `assert_called_once` (and etc.) variants.

  * Make any required fixes to the resulting unit tests.

  * Patch the `unittest.mock.Mock` class so that these `call_once` (and etc.) functions are now defined, but are defined in such a manner that any use of the functions raises an exception.  To this end, an abstract-syntax-tree (AST) solution was also explored successfully, but it was considered overkill to meet the requirements of this PR.  This AST-approach is something we might want to look into in the future, if we implement a "SNAPRed coding style" commit check.
### Dev testing

These changes are to the **unit-test framework** itself, and to **specific unit tests**.
These changes are SNAPRed-internal.  No changes should be observed in terms of SNAPRed's functionality. 

### CIS testing

As mentioned above, these changes are SNAPRed-internal.

## Link to EWM item

[EWM#6612](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6612)

### DIFF
--- a/src/snapred/backend/dao/indexing/CalculationParameters.py
+++ b/src/snapred/backend/dao/indexing/CalculationParameters.py
@@ -6,7 +6,7 @@ from snapred.backend.dao.state.InstrumentState import InstrumentState
 
 # NOTE: the request __init__ loads CalibrationExportRequest, which imports Calibration,
 #       which imports this module, which causes a circular import situation.
-#       In the future, need to remove the circualr import so that full set of ingredients
+#       In the future, need to remove the circular import so that full set of ingredients
 #       can be preserved
 # from snapred.backend.dao.request.FarmFreshIngredients import FarmFreshIngredients
 
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -3,6 +3,7 @@ import json
 import os
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
+from collections.abc import Iterable
 
 import numpy as np
 from mantid.dataobjects import MaskWorkspace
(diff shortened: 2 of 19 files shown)


dlcaballero16 (approved):
Unit tests now pass with all the updates.

PR #439, opened 2024-08-13
pull in changes from next to fix double error message

walshmm:
## Description of work

As the title states.

## To test

Prep your calibration folder such that you have an initialized calibration folder but no normalization.
Then attempt to perform a reduction.

This should net ONLY 1 error box.
Staging currently nets 2.

`runnumber: 59039`

[EWM#6095](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6095)

### DIFF
--- a/docs/source/api/snapred.ui.view.rst
+++ b/docs/source/api/snapred.ui.view.rst
@@ -52,7 +52,7 @@ snapred.ui.view.DiffCalTweakPeakView module
 .. automodule:: snapred.ui.view.DiffCalTweakPeakView
    :members:
    :undoc-members:
-   :exclude-members: signalRunNumberUpdate, signalValueChanged, signalUpdateRecalculationButton, signalPeakThresholdUpdate, signalMaxChiSqUpdate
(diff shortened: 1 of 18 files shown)


darshdinger (approved):
Tested with 46680 instead of 59039 instead since I wasn't able to get enough peaks. Seems functional, I only get one error box stating that the Normalization does not exist.

On another note, the continue anyway within DiffCal workflow doesn't seem to be working. It throws a non implemented error which is okay but this thrown error doesn't give the user any feedback. There is a stacktrace within the console but SNAPRed simply just crashes, there is not popup feedback.

walshmm:
> Tested with 46680 instead of 59039 instead since I wasn't able to get enough peaks. Seems functional, I only get one error box stating that the Normalization does not exist.
> 
> On another note, the continue anyway within DiffCal workflow doesn't seem to be working. It throws a non implemented error which is okay but this thrown error doesn't give the user any feedback. There is a stacktrace within the console but SNAPRed simply just crashes, there is not popup feedback.

I think I need to address that, sounds like it breaks continue anyway in that case, thanks for testing this!

darshdinger (approved):

PR #442, opened 2024-08-15
Fixes to the iteration dropdown menu.

darshdinger:
## Description of work
This work encapsulate a fix for the iteration dropdown. It was found that when a calibration was performed iterated and completed and then a subsequent calibration was started, the iteration dropdown menu would still be present with filled values. This work implements some hooks as stated within the solution on the EWM post to fix this issue. 

## Explanation of work
Given above.

### Dev testing
With a clean state directory, please run through the calibration workflow on `next` and reproduce the issue stated above. 
Following this, rerun the calibration workflow with another clean state directory on this branch and insure that the iteration dropdown issue has been resolve.

Ensure all pytests pass.

### CIS testing
Same as dev testing.

## Link to EWM item

[EWM # 6388](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6388)

### DIFF
--- a/src/snapred/ui/workflow/WorkflowImplementer.py
+++ b/src/snapred/ui/workflow/WorkflowImplementer.py
@@ -52,7 +52,7 @@ class WorkflowImplementer(QObject):
 
         self.resetHooks = []
 
-    def addResetHook(self, hook):
+    def addResetHook(self, hook: callable[[], None]):
         self.resetHooks.append(hook)
 
     def iterate(self, workflowPresenter):
--- a/src/snapred/ui/workflow/WorkflowImplementer.py
+++ b/src/snapred/ui/workflow/WorkflowImplementer.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Callable, List
 
 from qtpy.QtCore import QObject, Signal
 
@@ -52,7 +52,7 @@ class WorkflowImplementer(QObject):
 
         self.resetHooks = []
 
-    def addResetHook(self, hook: callable[[], None]):
+    def addResetHook(self, hook: Callable[[], None]):
         self.resetHooks.append(hook)
 
     def iterate(self, workflowPresenter):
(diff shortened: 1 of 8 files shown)


src/snapred/ui/workflow/WorkflowImplementer.py
```diff
@@ -50,6 +50,11 @@ def __init__(self, parent=None):
         self.continueAnywayFlags = ContinueWarning.Type.UNSET
         self.responseHandler = SNAPResponseHandler(self.parent)
 
+        self.resetHooks = []
+
+    def addResetHook(self, hook):
```
ekapadi, line 55: `def addResetHook(self, hook: Callable[[], None]):`  : I think it's always important to specify the expected signature and return value of any function passed as an argument.

src/snapred/ui/workflow/WorkflowImplementer.py
```diff
@@ -91,6 +96,9 @@ def reset(self, retainOutputs=False):
         self.outputs = []
         self.collectedOutputs = []
 
+        for hook in self.resetHooks:
```
ekapadi, line 99: See also, all of the `startLambda, resetLambda, iterateLambda, ...` at `Workflow.py`.  However, I think your approach of using (possibly multiple) _hook_ functions is probably _better_.  (And, that naming messes with the actual meaning of `lambda` as being _anonymous_.)

ekapadi (commented):
I'm testing right now.  I had only the one requested change about the `Callable` annotation.

src/snapred/ui/workflow/WorkflowImplementer.py
darshdinger: Yeah, Michael had mentioned this but I figured this way would be a bit more generalized.

ekapadi (approved):
Everything looks good now!  I tested this and it behaves correctly.

PR #443, opened 2024-08-19
Don't cross the streams!

rboston628:
## Description of work

SNAPRed is configured to log all of mantid's info, plus all of SNAPRed's info, which results in thousands of lines when even jsut starting up.

This splits the SNAPRed and mantid logger.  Everything from SNAPRed will be logged to the logging panel.  From mantid, only logs at WARNING or higher will be logged to the panel, while all logs at INFO or higher will go to an external file.

These level settings, and the name of the file, can be adjusted from the `application.yml`

This reduces the screen output from SNAPRed significantly.

## Explanation of work

The SNAPRed logger used to first setup the mantid logger to print similarly to the other logs.  It no longer does this.

A separate logger object is created for mantid logs.  It uses a `StreamHandler` and a `FileHandler` object to send the logs to both a file and to stdout.  The levels for these can be set independently, as can the output file.

One algorithm, `LoadGroupingDefinition`, was accidentally setup to use standard python logger, and not the SNAPRed logger.  That was fixed.

### Dev testing

~~See testing instructions on [mantid PR 37852](https://github.com/mantidproject/mantid/pull/37852) for setting this up to run with the new mantid feature.~~   No longer part of this PR.

Open the SNAPRed GUI.  On the terminal, you should only see the splash and a message "Welcome User!  Happy Reducing!"
image

Check the output log file (should be `mantidlog.txt`).  There should be some other messages there at the INFO level.  If this is empty, or contains DEBUG information, then this fails.

Try running calibration with 46680.  In the terminal logs, you will see a few mantid WARNING messages in yellow.  They should be obvious.  In the log file, there will be a bunch of information, some at INFO and some at WARNING level.  There should be nothing at the DEBUG level in the file.

Try changing the log levels in the  `application.yml` , and make sure mantid logs are printed at the correct level in the two places in a way that makes sense.

Open workbench and run SNAPRed from the interface.  Make sure the SNAPRed logs appear.  The mantid logs will also appear, and there just isn't anything we can do about that.

Close the SNAPRed interface and try running any algorithm in workbench, to make sure it is still writing logs.

### CIS testing

``` python
```

## Link to EWM item

[EWM#6257](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6257)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ conda.recipe/
 src/snapred/resources/dev.yml
 docs/source/developer/dao.rst
 docs/source/developer/snapred/*
+mantidlog.txt
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/src/snapred/backend/log/logger.py
+++ b/src/snapred/backend/log/logger.py
@@ -18,15 +18,17 @@ class CustomFormatter(logging.Formatter):
     _bold_red = "\x1b[31;1m"
     _reset = "\x1b[0m"
     _host = socket.gethostname().split(".")[0]
-    _format = _host + " - " + Config["logging.SNAP.format"]
 
-    FORMATS = {
-        logging.DEBUG: _grey + _format + _reset,
-        logging.INFO: _grey + _format + _reset,
-        logging.WARNING: _yellow + _format + _reset,
-        logging.ERROR: _red + _format + _reset,
-        logging.CRITICAL: _bold_red + _format + _reset,
-    }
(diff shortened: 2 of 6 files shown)


rboston628:
Do NOT merge before [mantid PR 37852](https://github.com/mantidproject/mantid/pull/37852)

dlcaballero16 (approved):
Tested and works as expected

PR #446, opened 2024-08-22
Fixed issue with threshold value not being set in certain cases

dlcaballero16:
## Description of work

This is to fix an issue with the PeakIntensityFractionThreshold in a CalibrantSample not correctly setting the default value in certain cases.

## Explanation of work

Updated the CalibrantSample DAO to set the default value from Config. Also updated SousChef to handle a case where the calibrant path might be None when getting the threshold value.

### Dev testing

Do full runs of DiffCal and Normalization and make sure nothing new is broken. You should be able to use calibrant sample files that don't have the threshold field set, and it should use the default value in application.yml.

### CIS testing

See dev testing.

## Link to EWM item

[EWM#3146](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=3146)


ekapadi (approved):
Ran all tests, including the full set of workflow panels.  Everything works as it should.

PR #448, opened 2024-08-23
write correct calibrant to norm record and then use it in reduction

walshmm:
## Description of work

Pretty simple fix, we need to record the calibrant sample used in normalization and use that in reduction instead of the one use during Calibration.

This is because it is used in preparing the Normalization as part of reduction, and no where else, thus it should be the calibrant from normalization.

## To test

Tests pass, the workflows complete as normal.

I used the typical inputs
```
For reduction/calib
Runnumber: 59039 

for calib:
sample: LA11b6
Grouping: Column
Peak Intensity Thresh: 0.01
Chi: 2000
dmin: 1

for norm:
rnunumber: 58810
bgrunnumber: 58813
```
## Link to EWM item

[EWM#6898](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6898)

### DIFF
--- a/src/snapred/backend/dao/normalization/NormalizationRecord.py
+++ b/src/snapred/backend/dao/normalization/NormalizationRecord.py
@@ -28,6 +28,7 @@ class NormalizationRecord(BaseModel):
     workspaceNames: List[WorkspaceName] = []
     version: int = Config["instrument.startingVersionNumber"]
     crystalDBounds: Limit[float]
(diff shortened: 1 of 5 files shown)


dlcaballero16 (approved):
Tested and works as expected

PR #450, opened 2024-08-28
Reduction save panel

ekapadi:
## Description of work

At the end of the reduction process for each run number, the complete reduction output data and metadata are saved to a path under the IPTS directory associated with that run number.  This saving step occurs automatically prior to the display of the actual reduction "save" panel. In addition, any cleanup of ADS-resident workspaces occurs at the same time.  The final reduction "save" panel now shows a helpful message indicating that the data has been saved, and to which output path (shown as the reduction-output state root).

In the case that the user does not have write permissions to the expected save directory, a continue-anyway prompt is displayed at the start of the reduction process. In this case, the message in the "save" panel reminds the user that they need to save their output workspaces.

## Explanation of work

  * Automatic save of reduction output-workspace data and metadata at the end of the reduction process for each run;

  * Checks of write-permissions and any associated notification prior to the start of the reduction process;

  * Analogous write-permissions checks prior to the start of the diffraction-calibration and normalization-calibration processes;  Note that unlike reduction, users without the required write permissions for diffraction-calibration or normalization-calibration are not allowed to continue with these processes.  Regardless, the continue-anyway mechanisms have been applied to this purpose, as this then allows any modification of all of this behavior to be centralized.

  * Same-state verification of run numbers entered in the reduction-request panel.  At present, in order to facilitate pixel-mask selection, all run numbers must be from the same state.  This requirement can eventually be omitted through the addition of a more-complicated reduction-panel sequence involving pixel-mask selection and assignment on a per-state basis.

  * Reduction output filenames now include the run number, rather than the state-id.  (Note that this is the first time that these output steps have actually been hooked up, so this change shouldn't affect anything.)

  * Adjustments to the text format in some continue-anyway message boxes in order to make better use of screen real estate.  Some text is smaller, and some _bold_ formatting is now used.

  * Fixups to `ValueError` input-validation behavior in the reduction panels so that these exceptions actually result in message boxes, rather than core dumps.  Note that the previous code had been written under the assumption that all of these input-validation steps were occurring in sections wrapped by `SNAPResponseHandler`, and this is most definitely _not_ the case.

  * Fixups to continue-anyway logic.  This allows multiple continue-anyway messages to pop-up in the correct sequence, and prevents these queries from occurring multiple times.

  * Implementation of a more general service-layer `checkWritePermissions` method, which actually checks whether or not the user _could_ correctly _create_ a given path, provided at least one of its parent components exists.

### Dev testing
Additional unit tests have been added to cover the changes of this commit.  Several _existing_ unit tests were modified so that they made a bit more sense (e.g. continue-anyway logic).

**Automatic save process:** verify that the expected files are written to the correct location on disk.  Ironically, the easiest way to get this location is to look at the message in the final "save" panel.  Note that the _contents_ of these files has been verified elsewhere, and is _not_ part of this pull request.

**In order to test the write permissions tests and notifications:**
    Start with an SNS-mirror section, containing your favorite run numbers (e.g. 46680 and 61991 -- two different states).  Step by step, going through the workflows: (1) demonstrate that the panel succeeds as expected in normal operation; (2) `chmod 555 "instrument.calibration.powder.home" / <state id>` -- demonstrate that the notification pops up, and you aren't actually allowed to continue with diffraction-calibration or normalization-calibration workflows, and (3) for reduction, `chmod 555 <read out the save directory root from the successful process test>` -- a notification should pop up, and after continuing, the "save" panel should show an appropriate reminder.

### CIS testing
Same as for dev testing.

## Link to EWM item

[EWM#6323](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=6323)

### DIFF
--- a/src/snapred/backend/dao/request/NormalizationRequest.py
+++ b/src/snapred/backend/dao/request/NormalizationRequest.py
@@ -1,7 +1,10 @@
+from typing import Optional
+
 from pydantic import BaseModel
 
 from snapred.backend.dao.Limit import Limit, Pair
 from snapred.backend.dao.state.FocusGroup import FocusGroup
+from snapred.backend.error.ContinueWarning import ContinueWarning
 from snapred.meta.Config import Config
 
 
@@ -27,3 +30,5 @@ class NormalizationRequest(BaseModel, extra="forbid"):
     )
     nBinsAcrossPeakWidth: int = Config["calibration.diffraction.nBinsAcrossPeakWidth"]
     fwhmMultipliers: Pair[float] = Pair.model_validate(Config["calibration.parameters.default.FWHMMultiplier"])
+
+    continueFlags: Optional[ContinueWarning.Type] = ContinueWarning.Type.UNSET
--- a/src/snapred/backend/service/NormalizationService.py
+++ b/src/snapred/backend/service/NormalizationService.py
@@ -22,7 +22,6 @@ from snapred.backend.dao.response.NormalizationResponse import NormalizationResp
 from snapred.backend.data.DataExportService import DataExportService
 from snapred.backend.data.DataFactoryService import DataFactoryService
 from snapred.backend.data.GroceryService import GroceryService
-from snapred.backend.error.ContinueWarning import ContinueWarning
 from snapred.backend.log.logger import snapredLogger
 from snapred.backend.recipe.GenericRecipe import (
     FocusSpectraRecipe,
@@ -177,7 +176,9 @@ class NormalizationService(Service):
         # check that the user has write permissions to the save directory
         if not self.checkWritePermissions(request.runNumber):
             raise RuntimeError(
-                "<font size = ""2"" >"
+                "<font size = "
+                "2"
+                " >"
                 + "<p>It looks like you don't have permissions to write to "
                 + f"<br><b>{self.getSavePath(request.runNumber)}</b>,<br>"
                 + "which is a requirement in order to run the normalization-calibration workflow.</p>"
@@ -185,7 +186,7 @@ class NormalizationService(Service):
                 + "<br><b>instrument.calibration.powder.home</b> entry in SNAPRed's <b>application.yml</b> file.</p>"
                 + "</font>"
             )
-        
+
     def _sameStates(self, runnumber1, runnumber2):
         stateId1 = self.dataFactoryService.constructStateId(runnumber1)
         stateId2 = self.dataFactoryService.constructStateId(runnumber2)
--- a/src/snapred/backend/service/NormalizationService.py
+++ b/src/snapred/backend/service/NormalizationService.py
@@ -9,6 +9,7 @@ from snapred.backend.dao.normalization import (
     Normalization,
 )
 from snapred.backend.dao.request import (
+    CalibrationWritePermissionsRequest,
     CreateNormalizationRecordRequest,
     FarmFreshIngredients,
     FocusSpectraRequest,
@@ -21,6 +22,7 @@ from snapred.backend.dao.response.NormalizationResponse import NormalizationResp
 from snapred.backend.data.DataExportService import DataExportService
 from snapred.backend.data.DataFactoryService import DataFactoryService
 from snapred.backend.data.GroceryService import GroceryService
+from snapred.backend.error.ContinueWarning import ContinueWarning
 from snapred.backend.log.logger import snapredLogger
 from snapred.backend.recipe.GenericRecipe import (
     FocusSpectraRecipe,
@@ -58,6 +60,7 @@ class NormalizationService(Service):
         self.diffractionCalibrationService = CalibrationService()
         self.sousChef = SousChef()
         self.registerPath("", self.normalization)
+        self.registerPath("validateWritePermissions", self.validateWritePermissions)
         self.registerPath("assessment", self.normalizationAssessment)
         self.registerPath("save", self.saveNormalization)
         self.registerPath("smooth", self.smoothDataExcludingPeaks)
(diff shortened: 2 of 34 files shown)


src/snapred/backend/dao/request/NormalizationRequest.py
```diff
@@ -27,3 +30,5 @@ class NormalizationRequest(BaseModel, extra="forbid"):
     )
     nBinsAcrossPeakWidth: int = Config["calibration.diffraction.nBinsAcrossPeakWidth"]
     fwhmMultipliers: Pair[float] = Pair.model_validate(Config["calibration.parameters.default.FWHMMultiplier"])
+
+    continueFlags: Optional[ContinueWarning.Type] = ContinueWarning.Type.UNSET
```
walshmm, line 34: I wonder if these should live on some parent request object

src/snapred/backend/service/NormalizationService.py
```diff
@@ -147,11 +149,62 @@ def normalization(self, request: NormalizationRequest):
...
+        if not self.checkWritePermissions(request.runNumber):
+            continueFlags |= ContinueWarning.Type.NO_WRITE_PERMISSIONS | ContinueWarning.Type.CONTINUE_NOT_POSSIBLE
+
+        # remove any continue flags that are present in the request by xor-ing with the flags
+        if request.continueFlags:
+            continueFlags = continueFlags ^ (request.continueFlags & continueFlags)
+
+        if continueFlags:
+            raise ContinueWarning(
+                "<p>It looks like you don't have permissions to write to "
```
walshmm, line 188: it would be cool to eventually migrate these templatized messages over to the `application.yml`

src/snapred/backend/service/NormalizationService.py
```diff
@@ -147,11 +149,62 @@ def normalization(self, request: NormalizationRequest):
...
+
+        :param request: a write-permissions request containing the run number and existing continue flags
+        :type request: CalibrationWritePermissionsRequest
+        """
+        # Note: this is split-out as a separate method so it can be checked as early as possible in the workflow.
+        continueFlags = ContinueWarning.Type.UNSET
+
+        # check that the user has write permissions to the save directory
+        if not self.checkWritePermissions(request.runNumber):
+            continueFlags |= ContinueWarning.Type.NO_WRITE_PERMISSIONS | ContinueWarning.Type.CONTINUE_NOT_POSSIBLE
```
walshmm, line 180: I do see the value in a `CONTINUE_NOT_POSSIBLE` for uncertain results failing from a continue anyway, but in this case would it not just be an exception?  Lack of write perms is fatal to the save calibration process.

src/snapred/backend/dao/request/NormalizationRequest.py
ekapadi: **I also think that they should.**  There's a TODO somewhere about adding them to `SNAPRequest`.  However, the problem right now is that we don't derive most of our requests from `SNAPRequest`, which _might_ _also_ be a good idea.

src/snapred/backend/service/NormalizationService.py
ekapadi: I think it's standard practice for most larger codebases to have a central "string table".  I wonder about adding it to `application.yml` however.  The problem is that the end users actually see that file.  (And, for SNAPRed, unfortunately it has started functioning like `Mantid.properties`.)  My point here is that there possibly needs to be a separate layer.

ekapadi: Oh, which reminds me:  maybe we need a method that gives us (from the _templatized_ string) the _un-templatized_ strings (i.e. without the tags, and with "<br>" replaced by "\n")?

src/snapred/backend/service/NormalizationService.py
ekapadi: I had to think about this a bit. It seems to me that the only reason it's _fatal_ is that so much of the rest of SNAPRed depends on the precise location, format, and _name_ of the saved _calibration_ data files + metadata files.  What I'm trying to do here is allow the user to "muck around" in the panels and to see what's possible.  For example, maybe they _do_ want to do their own calibration, but these messages will mean that they know they need to talk to the CIS first, and / or might help them with _which_ paths they need to adjust in the `application.yml`.

ekapadi: My argument for this living in the `ContinueWarning` is _twofold_.  (1) It makes a lot of sense in terms of implementation, as _many_ of these cases might be _changed_ in the future, in order to _allow_ a different level of continuation.  And (2) it also makes a kind-of logical sense, if you consider it as one end of the coninuum -- it's not "ridiculous" that it would belong there!

For example, (if I'm remembering correctly) I ran into case (1) _immediately_ in converting this PR to "staging".  On staging, there is _NO_ allowed continuation for the "no calibration" case, but on "next" there _is_ allowed continuation.

walshmm:
image
after running Calibration and Normalization successfully I then ran reduction and confirmed that the data was saved out to where it was specified
However, load nexus fails to load the file with the following error message:
```
Error in execution of algorithm LoadNexusProcessed:
Cannot open group mantid_workspace_5 of class NXentry (trying to open path /mantid_workspace_5)
Error in execution of algorithm LoadNexus:
Cannot open group mantid_workspace_5 of class NXentry (trying to open path /mantid_workspace_5)
```
image
confirmed new state check error message for normalization
image
image
I didnt actually get the expected calibration/norm write perms message? `self.checkWritePermissions(request.runNumber)` might not be acting as expected on analysis? (tested by using default `application.yml`)

ekapadi:
**These changes should fix the _permissions checking_ issues on analysis.**

ekapadi:
Also see `LocalDataService.readReductionData` for how the workspaces should be reloaded using `LoadNexus`.

walshmm:
image
image
seems to be working as intended now

walshmm:
image
If I open it one entry at a time it loads

walshmm (approved):

PR #451, opened 2024-08-30
Reduction save panel [staging]

ekapadi:
## Description of work
**Modified from the original [PR#450](https://github.com/neutrons/SNAPRed/pull/450) for "next".**

At the end of the reduction process for each run number, the complete reduction output data and metadata are saved to a path under the IPTS directory associated with that run number.  This saving step occurs automatically prior to the display of the actual reduction "save" panel. In addition, any cleanup of ADS-resident workspaces occurs at the same time.  The final reduction "save" panel now shows a helpful message indicating that the data has been saved, and to which output path (shown as the reduction-output state root).

In the case that the user does not have write permissions to the expected save directory, a continue-anyway prompt is displayed at the start of the reduction process. In this case, the message in the "save" panel reminds the user that they need to save their output workspaces.

## Explanation of work

  * Automatic save of reduction output-workspace data and metadata at the end of the reduction process for each run;

  * Checks of write-permissions and any associated notification prior to the start of the reduction process;

  * Analogous write-permissions checks prior to the start of the diffraction-calibration and normalization-calibration processes;  Note that unlike reduction, users without the required write permissions for diffraction-calibration or normalization-calibration are not allowed to continue with these processes.  Regardless, the continue-anyway mechanisms have been applied to this purpose, as this then allows any modification of all of this behavior to be centralized.
    
  * **[specific to staging]:** the continue-anyway part of `ReductionService.validateRequest` when diffraction-calibration or normalization-calibration aren't present has been retained from `next`, although modified to not allow the user to continue.  This provides an earlier warning than the previous one.
   
  * Same-state verification of run numbers entered in the reduction-request panel.  At present, in order to facilitate pixel-mask selection, all run numbers must be from the same state.  This requirement can eventually be omitted through the addition of a more-complicated reduction-panel sequence involving pixel-mask selection and assignment on a per-state basis.

  * Reduction output filenames now include the run number, rather than the state-id.  (Note that this is the first time that these output steps have actually been hooked up, so this change shouldn't affect anything.)

  * Adjustments to the text format in some continue-anyway message boxes in order to make better use of screen real estate.  Some text is smaller, and some _bold_ formatting is now used.

  * Fixups to `ValueError` input-validation behavior in the reduction panels so that these exceptions actually result in message boxes, rather than core dumps.  Note that the previous code had been written under the assumption that all of these input-validation steps were occurring in sections wrapped by `SNAPResponseHandler`, and this is most definitely _not_ the case.

  * Fixups to continue-anyway logic.  This allows multiple continue-anyway messages to pop-up in the correct sequence, and prevents these queries from occurring multiple times.

  * Implementation of a more general service-layer `checkWritePermissions` method, which actually checks whether or not the user _could_ correctly _create_ a given path, provided at least one of its parent components exists.

### Dev testing
Additional unit tests have been added to cover the changes of this commit.  Several _existing_ unit tests were modified so that they made a bit more sense (e.g. continue-anyway logic).

**Automatic save process:** verify that the expected files are written to the correct location on disk.  Ironically, the easiest way to get this location is to look at the message in the final "save" panel.  Note that the _contents_ of these files has been verified elsewhere, and is _not_ part of this pull request.

**In order to test the write permissions tests and notifications:**
    Start with an SNS-mirror section, containing your favorite run numbers (e.g. 46680 and 61991 -- two different states).  Step by step, going through the workflows: (1) demonstrate that the panel succeeds as expected in normal operation; (2) `chmod 555 "instrument.calibration.powder.home" / <state id>` -- demonstrate that the notification pops up, and you aren't actually allowed to continue with diffraction-calibration or normalization-calibration workflows, and (3) for reduction, `chmod 555 <read out the save directory root from the successful process test>` -- a notification should pop up, and after continuing, the "save" panel should show an appropriate reminder.

### CIS testing
Same as for dev testing.

## Link to EWM item

[EWM#6323](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=6323)

### DIFF
--- a/docs/source/api/snapred.ui.view.rst
+++ b/docs/source/api/snapred.ui.view.rst
@@ -30,7 +30,7 @@ snapred.ui.view.DiffCalAssessmentView module
    :show-inheritance:
 
 snapred.ui.view.DiffCalRequestView module
-------------------------------------------------------
+-----------------------------------------
 
 .. automodule:: snapred.ui.view.DiffCalRequestView
    :members:
@@ -105,7 +105,7 @@ snapred.ui.view.NormalizationRequestView module
    :show-inheritance:
 
 snapred.ui.view.NormalizationSaveView module
------------------------------------------------
+--------------------------------------------
 
 .. automodule:: snapred.ui.view.NormalizationSaveView
    :members:
@@ -132,10 +132,10 @@ snapred.ui.view.PromptUserforCalibrationInputView module
    :show-inheritance:
    :exclude-members: dataEntered
 
-snapred.ui.view.reduction.ReductionView module
-----------------------------------------------
+snapred.ui.view.reduction.ReductionRequestView module
+-----------------------------------------------------
 
-.. automodule:: snapred.ui.view.reduction.ReductionView
+.. automodule:: snapred.ui.view.reduction.ReductionRequestView
    :members:
    :undoc-members:
    :exclude-members: signalRemoveRunNumber
--- a/src/snapred/backend/dao/calibration/CalibrationIndexEntry.py
+++ b/src/snapred/backend/dao/calibration/CalibrationIndexEntry.py
@@ -22,7 +22,7 @@ class CalibrationIndexEntry(BaseModel):
     appliesTo: Optional[str] = None
     comments: str
     author: str
-    timestamp: Optional[int] = None
+    timestamp: Optional[float] = None
 
     def parseAppliesTo(appliesTo: str):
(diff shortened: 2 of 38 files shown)


ekapadi:
Added in the new changes for _permissions checking._

walshmm (approved):
All checks from next still passing on this version :+1:

PR #452, opened 2024-09-04
Ewm6834 fix default record

walshmm:
## Description of work

This pr allows:
- snapred to utilize the default calib record written out as part of state init
- reduction to refer to the correct calibrant sample when creating peaks to purge from normalization

This needs a follow up to address when a state is totally uninitialized.

## Explanation of work

I brought over some changes from `staging` for the calibrant sample stuff.  These were actually somewhat essential when interacting with the default record, as it doesnt have a calibrant sample set (nor is it relevant to reduction).

I updated `GroceryService` and `Indexer` to account for the default record, adding a case to attempt to read on if no other record is available.  Doing this, I also changed the way we pull diffcaltable names, and refer to the calibration record filtering on ws type--this simplified referring to the special name default tables recieve and allow me to avoid special logic in this case.

## To test

Initialize a new state.
run reduction
it should prompt you to continue anyway
see that it completes
```
For reduction/calib
Runnumber: 59039 

for calib:
sample: LA11b6
Grouping: Column
Peak Intensity Thresh: 0.01
Chi: 200
dmin: 1

for norm:
rnunumber: 58810
bgrunnumber: 58813
```

this should apply the default calibration during reduction.

## Link to EWM item

[EWM#6834](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6834)

### DIFF
--- a/src/snapred/backend/dao/normalization/NormalizationRecord.py
+++ b/src/snapred/backend/dao/normalization/NormalizationRecord.py
@@ -33,6 +33,7 @@ class NormalizationRecord(Record, extra="ignore"):
     workspaceNames: List[WorkspaceName] = []
     calibrationVersionUsed: int = VERSION_DEFAULT
     crystalDBounds: Limit[float]
+    normalizationCalibrantSamplePath: str
 
     # must also parse integers as background run numbers
     @field_validator("backgroundRunNumber", mode="before")
--- a/src/snapred/backend/dao/request/CreateNormalizationRecordRequest.py
+++ b/src/snapred/backend/dao/request/CreateNormalizationRecordRequest.py
@@ -1,29 +1,18 @@
-from typing import List, Optional
+from pydantic import ConfigDict
 
-from pydantic import BaseModel, ConfigDict
-
-from snapred.backend.dao.indexing.Versioning import VERSION_DEFAULT
-from snapred.backend.dao.Limit import Limit
-from snapred.backend.dao.normalization.Normalization import Normalization
-from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceName
+from snapred.backend.dao.normalization import NormalizationRecord
 
 
-class CreateNormalizationRecordRequest(BaseModel, extra="forbid"):
+class CreateNormalizationRecordRequest(NormalizationRecord, extra="forbid"):
     """
     The needed data to create a NormalizationRecord.
     """
-
-    runNumber: str
-    useLiteMode: bool
-    version: Optional[int] = None
-    calculationParameters: Normalization
-    backgroundRunNumber: str
-    smoothingParameter: float
-    workspaceNames: List[WorkspaceName] = []
-    calibrationVersionUsed: Optional[int] = VERSION_DEFAULT
-    crystalDBounds: Limit[float]
 
     model_config = ConfigDict(
         # required in order to use 'WorkspaceName'
         arbitrary_types_allowed=True,
     )
+
+    @classmethod
+    def parseVersion(cls, version, *, exclude_none: bool = False, exclude_default: bool = False) -> int | None:  # noqa: ARG003
+        return version
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -26,6 +26,7 @@ from snapred.meta.decorators.Singleton import Singleton
 from snapred.meta.mantid.WorkspaceNameGenerator import (
     NameBuilder,
     WorkspaceName,
+    WorkspaceType,
 )
 from snapred.meta.mantid.WorkspaceNameGenerator import (
     WorkspaceNameGenerator as wng,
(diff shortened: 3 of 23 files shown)


src/snapred/backend/data/GroceryService.py
```diff
@@ -222,11 +224,19 @@ def _createDiffcalDiagnosticWorkspaceFilename(self, item: GroceryListItem) -> st
             / (self._createDiffcalOutputWorkspaceName(item) + ext)
         )
 
+    def _createDiffcalTableFilepathFromWsName(
```
ekapadi, line 227: I don't understand this.  I could do
`runNumber, useLiteMode, version = wsName.tokens('runNumber', 'useLiteMode', 'version')`
to get the required field values.  What is the purpose of the redundant arguments?

ekapadi: **Also, and possibly more importantly, what happens when these argument values are in contradiction with each other?**

src/snapred/backend/data/LocalDataService.py
```diff
@@ -842,7 +842,7 @@ def _writeDefaultDiffCalTable(self, runNumber: str, useLiteMode: bool):
         indexer = self.calibrationIndexer(runNumber, useLiteMode)
         version = indexer.defaultVersion()
         grocer = GroceryService()
-        filename = Path(grocer._createDiffcalTableWorkspaceName("default", useLiteMode, version) + ".h5")
+        filename = Path(grocer.createDiffcalTableWorkspaceName("default", useLiteMode, version) + ".h5")
```
ekapadi, line 845: I don't get this: why does "default" subsume the <run number> position (especially as "default" as a string is _not_ a legitimate <run number>), but not just use the `VERSION_DEFAULT` which is provided for that reason particularly?  (And then use the _real_ <run number> which might be useful information.  What am I missing here?)

src/snapred/backend/dao/normalization/NormalizationRecord.py
```diff
@@ -33,6 +33,7 @@ class NormalizationRecord(Record, extra="ignore"):
     workspaceNames: List[WorkspaceName] = []
     calibrationVersionUsed: int = VERSION_DEFAULT
     crystalDBounds: Limit[float]
+    normalizationCalibrantSamplePath: str
```
ekapadi, line 36: I'm tempted to suggest, that because this is a `NormalizationRecord`, the "normalization" part of `normalizationCalibrantSamplePath` is redundant.  And that to clarify the usage, the "normalization" only be added at the `ReductionRecord.normalizationCalibrantSamplePath`.

tests/util/WhateversInTheFridge.py
```diff
@@ -100,6 +100,7 @@ def readNormalizationRecord(self, runId: str, useLiteMode: bool, version: Option
             useLiteMode=useLiteMode,
             version=version,
             calculationParameters=self.calculationParameters_with_stateId("0xdeadbeef"),
+            normalizatinoCalibrantSamplePath="path/to/calibrant",
```
ekapadi, line 103: Mis-spelling `normalizationCalibrationSamplePath` (and also check it at the _source_ in the `NormalizationRecord` declaration).

src/snapred/backend/dao/request/CreateNormalizationRecordRequest.py
```diff
@@ -1,29 +1,18 @@
...
+from snapred.backend.dao.normalization import NormalizationRecord
 
-from snapred.backend.dao.indexing.Versioning import VERSION_DEFAULT
-from snapred.backend.dao.Limit import Limit
-from snapred.backend.dao.normalization.Normalization import Normalization
-from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceName
 
-
-class CreateNormalizationRecordRequest(BaseModel, extra="forbid"):
+class CreateNormalizationRecordRequest(NormalizationRecord, extra="forbid"):
```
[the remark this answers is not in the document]
walshmm: I did think this data object was kinda odd, I didnt look to closely at it.  Looking further into how its used there's a bit of a smell.  
I think this had been just passing a record around up until the indexer stuff?
I'll see if I can rectify this.

src/snapred/backend/data/LocalDataService.py
ekapadi: I realized later that possibly there's no meaningful \<run number\> where this is actually required.  I would like to see these "default" names centralized somewhere (we have the same requirement for \< version \> ), but that doesn't have much to do with this PR.

ekapadi:
**I tested reduction, following l_only_ state initialization (no calibration).  Everything works as it should.**

Then I tested the calibration and normalization panels:  **in normalization _save_ panel, there was a bug (as the new `normalizationCalibrantSamplePath` wasn't being set in `NormalizationWorkflow._saveNormalization`.**  (AND, a key question, why aren't most of the internals of this method part of the `NormalizationService`?!)

**Performed a minor rebase, fixed the bug, and retested.**  Everything appears to be working as it should be.
**I "approve" this PR and it should be ready to merge, once the "read the docs" thing resolves.**

However Darsh will actually need to approve it, since I modified the code.

darshdinger (approved):
Looks good. Reviewed and tested.

PR #454, opened 2024-09-10
Updated mantid version to fix conversion defect

dlcaballero16:
## Description of work

The mantid version was updated to fix an issue with converting units during reduction returning bad data. The bad data had negative values.

### Dev testing

On analysis, run reduction with run 61991, keep unfocused data, and convert units to Momentum Transfer. When reduction is complete, open the converted workspace, and you should see that all the data has positive values along the x-axis. 

### CIS testing

See Dev testing.

## Link to EWM item

[EWM#6788](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6788)

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -7,7 +7,7 @@ dependencies:
 - python=3.10
 - pip
 - pydantic>=2.7.3,<3
-- mantidworkbench>=6.10.20240705
+- mantidworkbench>=6.10.20240909
 - qtpy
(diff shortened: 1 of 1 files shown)


darshdinger (approved):
This results in the expected output behavior.
image

PR #455, opened 2024-09-11
Ewm 6274 skip pixel calibration

dlcaballero16:
## Description of work

This allows user to skip pixel calibration during diffraction calibration. This is needed when pixel calibration fails due to having too many masked pixels.

## Explanation of work

A button was added to the Tweak Peaks tab to skip pixel calibration. The DiffractionCalibrationRequest was modified so there is now a skipPixelCalibration flag. In the calibration service, The diffraction calibration step now checks for percentage of masked pixels, and throws an exception if there's too many masked pixels. The diffcal recipe now checks for the skipPixelCalibration flag to see if PixelCalibration needs to be executed.

### Dev testing

Locally, run diffraction calibration for run 46680 in native mode. On the tweak peaks tab, you should see a Skip Pixel Calibration button in the top right corner. By default, this should be toggled on when in native mode. Toggle the skip pixel calibration to be off. When you try to advance past the Tweak Peaks tab, you  should get a warning message pop up saying there is too many masked pixels. Redo the tweak peaks step and toggle the skip pixel calibration to be on. You should now be able to complete diffraction calibration. Note: running in native mode takes longer than lite mode.

Repeat diffcal in lite mode, but notice now that skipping pixel calibration defaults to off.

### CIS testing

See dev testing

## Link to EWM item

[EWM#6274](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6274)

### DIFF
--- a/src/snapred/backend/dao/ingredients/DiffractionCalibrationIngredients.py
+++ b/src/snapred/backend/dao/ingredients/DiffractionCalibrationIngredients.py
@@ -28,5 +28,6 @@ class DiffractionCalibrationIngredients(BaseModel):
     peakFunction: SymmetricPeakEnum = SymmetricPeakEnum[Config["calibration.diffraction.peakFunction"]]
     maxOffset: float = Config["calibration.diffraction.maximumOffset"]
     maxChiSq: float = Config["constants.GroupDiffractionCalibration.MaxChiSq"]
+    skipPixelCalibration: bool = False
 
     model_config = ConfigDict(extra="forbid")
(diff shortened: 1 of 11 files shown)


rboston628 (approved):
Ran with 46680 in native mode.  At Tweak Peak Peek, I see the toggle for pixel calibration is set to off.
image
I turn this off and continue.  Gears churn, and after a few minutes I receive this error.
image
I toggle skip back on, and run.  Gears churn, then it finishes
image
I go back to beginning and try in lite mode.  The skip toggle is set to off, meaning it has been reset to the correct default.
image
I run, and I get the same error about too many masked pixels.

PR #458, opened 2024-09-15
All workflow panels: happy-path integration tests

ekapadi:
## Description of work

To run these tests, a suitable mirror of the "/SNS" directory structure is required.  The minimum setup required is the tree as shown inline below. As currently setup, each of the tests creates any new files under a temporary directory location. By default these locations are just _under_ the `Config["instrument.calibration.powder.home"]` or the (IPTS-substituted) `Config["instrument.reduction.home"]` that would be used with the `Config["IPTS.root"]` as specified in order to setup the "/SNS" mirror.  This behavior means that the contents of the SNS mirror should not have changed after any test run.

Neither of the diffraction-calibration or the normalization-calibration workflow tests require existing calibration data.  However, the reduction-workflow tests, `test_reduction_panel_happy_path`, does require this data to exist.  If these data need to be updated, they can be generated using the `test_calibration_and_reduction_panels_happy_path` test (but comment out the `@pytest.skip`, and the following lines for `calibration_home_from_mirror` and `reduction_home_from_mirror` so that the generated files will be retained at the correct locations after the test exits).

It is hoped that these tests will prove useful to build out our integration-test framework.

There are a few fine points:

  * Simple dialogs, such as `QMessageBox`, `ActionPrompt`, and `SuccessPrompt` are mocked, using the appropriate _static_ methods (e.g `question`, `warning`, or `critical`).  For all of these cases, the expected message text is known, and any other message text will be treated as a fatal error.

  * Usage such as `
        self._criticalMessageBox = mock.patch(
            "qtpy.QtWidgets.QMessageBox.critical",
            lambda *args, *kwargs: pytest.fail(f"CRITICAL messagebox: {args}", pytrace=False)
        )
        self._criticalMessageBox.start()
        ...
        self._criticalMessageBox.stop() # IMPORTANT: don't forget this!
`
is REQUIRED over using the `mock.patch` as a context manager (e.g. in a `with` clause`).  The problem is that any exception throw, such as a `RecoverableException` or a `ContinueWarning` will break out of any enclosing `with` clause, and at that point, the `mock.patch` won't be working anymore.

  * As recommended by the online `qtbot` documentation, most data input is implemented by directly calling the associated methods (e.g. `setText`), rather than locating the dialog, entering text, and then clicking on the control to (e.g.) enter it.  There is one example of something like the latter, which actually was required for the "Initialize State" dialog, but that example will illustrate why such an approach isn't really recommended.  That example also illustrates why it's really useful to associate widgets (even temporary ones) with instance attributes, so that they can be accessed by name.

  * At several positions in the flow, it's necessary to _wait_ for the GUI state to catch up.  Some of these `waitUntil` have been located using _empirical_ techniques, and they don't always make sense. For successfully running on a laptop, a few of these `waitUntil` are also of longer duration -- in general this only has an effect on total test duration if the test FAILs.

  * A few convenience `pyqt.Signal` have been added, to assist in the detection of the completion of various workflow steps.  Most of these additions have already been merged to "next", as part of other PRs.

  * IMPORTANT: note that the `@Singleton` operator is _not_ mocked out for these tests, nor for any test using the `integration` marker (see the outermost `conftest.py`).  This was necessary in order to get the new `Indexer` class to work with respect to the newly-initialized state.  (Otherwise, each of the services has its own `LocalDataService` instance with its own `_indexer` `lru_cache`, and only _one_ of them ends up knowing about the state initialization until the indexers are reloaded at the next startup.)  FOR THE MOMENT: THIS MEANS THAT `integration` MARKED TESTS MUST BE RUN IN THEIR OWN `pytest` SESSION.

  * A related issue to the previous is that `LocalDataService()._indexer.cache_clear()` must be called after the `Config["instrument.calibration.powder.home"]` is overridden.  Otherwise, any cached indexers will still be operating with respect to the previous calibration-directory locations.

## Explanation of work

  * A `qtbot`-based integration test for the diffraction-calibration workflow.  This integration test executes the "happy path", that is, it assumes input is correctly entered, and that all of the information required to execute the workflow is available.  The one exception to this being that the "initialize state" dialog is exercised.

  * An integration test for the normalization-calibration workflow.  As mentioned, this integration test executes the "happy path".  The "initialize state" dialog is also exercised for this test.

  * An integration test including all three of the diffraction-calibration, normalization-calibration, and the reduction workflows.  By default, this test is marked as `@pytest.skip`, because each of these workflows has its own separate test.  This test is still useful in case a new set of diffraction-calibration and normalization-calibration data need to be generated for the reduction workflow test.

  * A test fixture to override the `Config["instrument.calibration.powder.home"]`, including rebuilding the directory skeleton under a temporary location, without including any existing calibration data. This fixture generates the setup required to run the diffraction-calibration and normalization-calibration tests.

  * A test fixture to override the (IPTS-substituted) `Config["instrument.reduction.home"]`.  It wasn't necessary to include any skeleton data for this fixture, and it primarily allows the generated reduced data to be cleaned up at test exit.
 
  * In the outermost `conftest.py`: `auto-use` test fixtures (at various scopes) to _reset_, or to _completely unwrap_ the `@Singleton` decorator, depending on whether or not integration tests are being run.

For testing:

Repeatedly, I have successfully run this test module on an _old_ laptop with 16GB RAM.  It isn't necessary to run them on `analysis.sns.gov`.  Issues (on a laptop) can sometimes occur if running multiple integration-test modules at once; however, these issues don't seem to occur on analysis.  This has something to do with memory allocation not succesfully being cleaned up between module-scope tests.  I am still working on this issue.

### Dev testing

To run the tests:

  * set `IPTS.root` in `integration_test.yml` to the prepared SNS-mirror path;

  * the commandline should be: `env=integration_test pytest -m integration -k happy_path`

  * on analysis (or any system with enough memory), you can run: `env=integration_test pytest -m integration`

  * `-m integration` is _critical_, as the `@Singleton`-related fixtures in `conftest.py` will act differently depending on it.

**Required minimal SNS-mirror directory tree:**

(WARNING: remember that any test-environment _data_ and _metadata_ files usually begin with the _underscore_, so you need to make sure that the calibration _versioned_ data has this format.)

(**I'm placing a copy of this on `analysis.sns.gov` under Michael's shared area: `/SNS/users/wqp/shared/EWM5537.SNS_root.tar`**  .)

------------------------------

<pre>
SNS_root
&#x2514&#x2500&#x2500 SNAP
&#x251c&#x2500&#x2500 IPTS-24641 &#x2502   &#x251c&#x2500&#x2500 nexus &#x2502   &#x2502   &#x2514&#x2500&#x2500 SNAP_46680.nxs.h5 &#x2502   &#x2514&#x2500&#x2500 shared
&#x2502   &#x251c&#x2500&#x2500 lite
&#x2502   &#x2502   &#x2514&#x2500&#x2500 SNAP_46680.lite.nxs.h5
&#x2514&#x2500&#x2500 shared
&#x251c&#x2500&#x2500 Calibration
&#x2502   &#x251c&#x2500&#x2500 CalibrantSamples
&#x2502   &#x2502   &#x251c&#x2500&#x2500 Diamond_001.json
&#x2502   &#x2502   &#x251c&#x2500&#x2500 EntryWithCollCode120585.cif
&#x2502   &#x2502   &#x251c&#x2500&#x2500 EntryWithCollCode37502_Nickel.cif
&#x2502   &#x2502   &#x251c&#x2500&#x2500 EntryWithCollCode40947_LaB6.cif
&#x2502   &#x2502   &#x251c&#x2500&#x2500 EntryWithCollCode51688.cif
&#x2502   &#x2502   &#x251c&#x2500&#x2500 EntryWithCollCode52054_diamond.cif
&#x2502   &#x2502   &#x251c&#x2500&#x2500 EntryWithCollCode94254_LaB6_neutron_2001.cif
&#x2502   &#x2502   &#x251c&#x2500&#x2500 fakeVanadium.cif
&#x2502   &#x2502   &#x251c&#x2500&#x2500 fakeVanadium.json
&#x2502   &#x2502   &#x251c&#x2500&#x2500 La11B6_NIST_660c_001.json
&#x2502   &#x2502   &#x251c&#x2500&#x2500 Na2Ca3Al2F14.cif
&#x2502   &#x2502   &#x251c&#x2500&#x2500 NAC.json
&#x2502   &#x2502   &#x251c&#x2500&#x2500 nickel_001.json
&#x2502   &#x2502   &#x251c&#x2500&#x2500 Silicon_NIST_640D_001.json
&#x2502   &#x2502   &#x251c&#x2500&#x2500 Silicon_NIST_640d.cif
&#x2502   &#x2502   &#x251c&#x2500&#x2500 vanadium_CollCode171003.cif
&#x2502   &#x2502   &#x251c&#x2500&#x2500 vanadium_cylinder_001.json
&#x2502   &#x2502   &#x251c&#x2500&#x2500 vanadium_niobium.cif
&#x2502   &#x2502   &#x2514&#x2500&#x2500 V-Nb_cylinder_001.json
&#x2502   &#x251c&#x2500&#x2500 Powder
&#x2502   &#x2502   &#x251c&#x2500&#x2500 04bd2c53f6bf6754
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 groupingMap.json
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 lite
&#x2502   &#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 diffraction
&#x2502   &#x2502   &#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 CalibrationIndex.json
&#x2502   &#x2502   &#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 v_0001
&#x2502   &#x2502   &#x2502   &#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 CalibrationParameters.json
&#x2502   &#x2502   &#x2502   &#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 CalibrationRecord.json
&#x2502   &#x2502   &#x2502   &#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 _diagnostic_bank_046680_v0001.nxs.h5
&#x2502   &#x2502   &#x2502   &#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 _diffract_consts_046680_v0001.h5
&#x2502   &#x2502   &#x2502   &#x2502   &#x2502   &#x2502   &#x2514&#x2500&#x2500 _dsp_bank_046680_v0001.nxs.h5
&#x2502   &#x2502   &#x2502   &#x2502   &#x2514&#x2500&#x2500 normalization
&#x2502   &#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 NormalizationIndex.json
&#x2502   &#x2502   &#x2502   &#x2502   &#x2514&#x2500&#x2500 v_0001
&#x2502   &#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 _dsp_all_fitted_van_corr_046680_v0001.nxs
&#x2502   &#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 NormalizationParameters.json
&#x2502   &#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 NormalizationRecord.json
&#x2502   &#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 _tof_all_s+f-vanadium_046680_v0001.nxs
&#x2502   &#x2502   &#x2502   &#x2502   &#x2514&#x2500&#x2500 _tof_unfoc_raw_van_corr_046680_v0001.nxs
&#x2502   &#x2502   &#x251c&#x2500&#x2500 LiteGroupMap.hdf
&#x2502   &#x2502   &#x251c&#x2500&#x2500 PixelGroupingDefinitions
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 defaultGroupingMap.json
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SingleColumn5.hdf
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_2_4.hdf
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_2_4.lite.hdf
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_5degSlice.lite.nxs
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_All.hdf
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_All.lite.hdf
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_All.lite.nxs
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_All.xml
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_Bank.hdf
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_Bank.lite.hdf
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_Bank.lite.nxs
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_Bank.xml
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_Column.hdf
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_Column.lite.hdf
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_Column.lite.nxs
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGroup_Column.xml
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGrp_All.lite.xml
&#x2502   &#x2502   &#x2502   &#x251c&#x2500&#x2500 SNAPFocGrp_Bank.lite.xml
&#x2502   &#x2502   &#x2502   &#x2514&#x2500&#x2500 SNAPFocGrp_Column.lite.xml
&#x2502   &#x2502   &#x2514&#x2500&#x2500 SNAPLite.xml
&#x2502   &#x251c&#x2500&#x2500 SNAPInstPrm.json
</pre>

------------------------------

## Link to EWM item

[EWM#5537](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=5537)

### DIFF
--- a/tests/integration/test_versions_in_order.py
+++ b/tests/integration/test_versions_in_order.py
@@ -220,6 +220,7 @@ class ImitationSousChef(SousChef):
 ###############################################################################
 
 
+# @pytest.mark.integration()  # TODO: why is this mark not used?!
 class TestVersioning(TestCase):
     def setUp(self):
         # NOTE the act of importing InterfaceController will cause test_APIService
@@ -259,6 +260,7 @@ class TestVersioning(TestCase):
     def tearDown(self):
         self.api.serviceFactory.getService = self.old_self
 
+    # @pytest.mark.integration()  # TODO: why is this mark not used?!
     def test_calibration_versioning(self):
         """
         After setting up the data services with the appropriate imitations,
--- a/tests/integration/test_workflow_panels_happy_path.py
+++ b/tests/integration/test_workflow_panels_happy_path.py
@@ -1,0 +1,1347 @@
+import os
+import re
+import tempfile
+from contextlib import ExitStack, suppress
+from pathlib import Path
+from typing import Optional
+from unittest import mock
+
+import pytest
+from mantid.kernel import amend_config
+from qtpy import QtCore
+from qtpy.QtCore import Qt
+from qtpy.QtWidgets import (
+    QMessageBox,
+    QTabWidget,
+)
+from snapred.__main__ import _prepend_datasearch_directories
+
+# I would prefer not to access `LocalDataService` within an integration test,
+#   however, for the moment, the reduction-data output relocation fixture is defined in the current file.
+from snapred.backend.data.LocalDataService import LocalDataService
+from snapred.meta.Config import Config, Resource
+from snapred.ui.main import SNAPRedGUI
+from snapred.ui.view import InitializeStateCheckView
+from snapred.ui.view.DiffCalAssessmentView import DiffCalAssessmentView
+from snapred.ui.view.DiffCalRequestView import DiffCalRequestView
+from snapred.ui.view.DiffCalSaveView import DiffCalSaveView
+from snapred.ui.view.DiffCalTweakPeakView import DiffCalTweakPeakView
+from snapred.ui.view.NormalizationRequestView import NormalizationRequestView
+from snapred.ui.view.NormalizationSaveView import NormalizationSaveView
+from snapred.ui.view.NormalizationTweakPeakView import NormalizationTweakPeakView
+from snapred.ui.view.reduction.ReductionRequestView import ReductionRequestView
+from snapred.ui.view.reduction.ReductionSaveView import ReductionSaveView
+from util.Config_helpers import Config_override
+
+# TODO: WorkflowNodeComplete signal, at end of each node!
+
+
+class InterruptWithBlock(BaseException):
+    pass
+
+
+@pytest.fixture()
+def calibration_home_from_mirror():
+    # Test fixture to create a copy of the calibration home directory from an existing mirror:
+    # * creates a temporary calibration home directory under the optional `prefix` path;
+    #   when not specified, the temporary directory is created under the existing
+    #   `Config["instrument.calibration.powder.home"]`;
+    # * creates symlinks within the directory to required metadata files and directories
+    #   from the already existing `Config["instrument.calibration.powder.home"]`;
+    # * ignores any existing diffraction-calibration and normalization-calibration subdirectories;
+    # * and finally, overrides the `Config` entry for "instrument.calibration.powder.home".
+
+    # IMPLEMENTATION notes:
+    # * The functionality of this fixture is deliberately NOT implemented as a context manager,
+    #   although certain context-manager features are used.
+    # * If this were a context manager, it would  be terminated at any exception throw.  For example,
+    #   it would be terminated by the "initialize state" `RecoverableException`.  Such termination would interfere with
+    #   the requirements of the integration tests.
+    _stack = ExitStack()
+
+    def _calibration_home_from_mirror(prefix: Optional[Path] = None):
+        originalCalibrationHome: Path = Path(Config["instrument.calibration.powder.home"])
+        if prefix is None:
+            prefix = originalCalibrationHome
+
+        # Override calibration home directory:
+        tmpCalibrationHome = Path(_stack.enter_context(tempfile.TemporaryDirectory(dir=prefix, suffix=os.sep)))
+        assert tmpCalibrationHome.exists()
+        _stack.enter_context(Config_override("instrument.calibration.powder.home", str(tmpCalibrationHome)))
+
+        # WARNING: for these integration tests `LocalDataService` is a singleton.
+        #   The Indexer's `lru_cache` MUST be reset after the Config override, otherwise
+        #     it will return indexers synched to the previous `Config["instrument.calibration.powder.home"]`.
+        LocalDataService()._indexer.cache_clear()
+
+        # Create symlinks to metadata files and directories.
+        metadatas = [Path("LiteGroupMap.hdf"), Path("PixelGroupingDefinitions"), Path("SNAPLite.xml")]
+        for path_ in metadatas:
+            os.symlink(originalCalibrationHome / path_, tmpCalibrationHome / path_)
+        return tmpCalibrationHome
+
+    yield _calibration_home_from_mirror
+
+    # teardown => __exit__
+    _stack.close()
+    LocalDataService()._indexer.cache_clear()
+
+
+@pytest.fixture()
+def reduction_home_from_mirror():
+    # Test fixture to write reduction data to a temporary directory under `Config["instrument.reduction.home"]`.
+    # * creates a temporary reduction state root directory under the optional `prefix` path;
+    #   when not specified, the temporary directory is created under the existing
+    #   `Config["instrument.reduction.home"]` (with the substituted 'IPTS' tag).
+    # * overrides the `Config` entry for "instrument.reduction.home".
+
+    # IMPLEMENTATION notes: (see previous).
+    _stack = ExitStack()
+
+    def _reduction_home_from_mirror(runNumber: str, prefix: Optional[Path] = None):
+        if prefix is None:
+            dataService = LocalDataService()
+            originalReductionHome = dataService._constructReductionStateRoot(runNumber)
+
+            # WARNING: this 'mkdir' step will not be reversed at exit,
+            #   but that shouldn't matter very much.
+            originalReductionHome.mkdir(parents=True, exist_ok=True)
+            prefix = originalReductionHome
+
+            tmpReductionHome = Path(_stack.enter_context(tempfile.TemporaryDirectory(dir=prefix, suffix=os.sep)))
+
+            # Ensure that `_createReductionStateRoot` will return the temporary directory,
+            #   while still exercising it's IPTS-substitution functionality.
+            _stack.enter_context(
+                Config_override(
+                    "instrument.reduction.home", Config["instrument.reduction.home"] + os.sep + tmpReductionHome.name
+                )
+            )
+
+            # No `LocalDataService._indexer.cache_clear()` should be required here, but keep it in mind, just in case!
+
+        else:
+            # Specified prefix => just use that, without any substitution.
+            # In this case `_constructReductionStateRoot` will return a path
+            #   which does not depend on the IPTS-directory for the run number.
+            tmpReductionHome = Path(_stack.enter_context(tempfile.TemporaryDirectory(dir=prefix, suffix=os.sep)))
+            _stack.enter_context(Config_override("instrument.reduction.home", str(tmpReductionHome)))
+
+        assert tmpReductionHome.exists()
+        return tmpReductionHome
+
+    yield _reduction_home_from_mirror
+
+    # teardown => __exit__
+    _stack.close()
+
+
+@pytest.mark.integration()
+class TestGUIPanels:
+    @pytest.fixture(scope="function", autouse=True)  # noqa: PT003
+    def _setup_gui(self, qapp):
+        # ---------------------------------------------------------------------------
+        # DEFAULT PATCHES:
+        #   FAIL TEST if any 'warning' OR 'critical' message boxes occur.
+        #   In the test body, these patches are overridden for special cases.
+
+        self._warningMessageBox = mock.patch(
+            "qtpy.QtWidgets.QMessageBox.warning",
+            lambda *args, **kwargs: pytest.fail(
+                "WARNING messagebox:\n" + f"    args: {args}\n" + f"    kwargs: {kwargs}", pytrace=False
+            ),
+        )
+        self._warningMessageBox.start()
+
+        self._criticalMessageBox = mock.patch(
+            "qtpy.QtWidgets.QMessageBox.critical",
+            lambda *args, **kwargs: pytest.fail(
+                "CRITICAL messagebox:\n" + f"    args: {args}\n" + f"    kwargs: {kwargs}", pytrace=False
+            ),
+        )
+        self._criticalMessageBox.start()
+
+        # patch log-warnings QMessage box: runs using `QMessageBox.exec`.
+        self._logWarningsMessageBox = mock.patch(
+            "qtpy.QtWidgets.QMessageBox.exec",
+            lambda self, *args, **kwargs: QMessageBox.Ok
+            if (
+                "The backend has encountered warning(s)" in self.text()
+                and "InstrumentDonor will only be used if GroupingFilename is in XML format." in self.detailedText()
+            )
+            else pytest.fail(
+                "unexpected QMessageBox.exec:\n"
+                + f"    args: {args}\n"
+                + f"    kwargs: {kwargs}\n"
+                + f"    text: '{self.text()}'\n"
+                + f"    detailed text: '{self.detailedText()}'",
+                pytrace=False,
+            ),
+        )
+        self._logWarningsMessageBox.start()
+
+        # Automatically continue at the end of each workflow.
+        self._actionPrompt = mock.patch(
+            "snapred.ui.presenter.WorkflowPresenter.ActionPrompt.prompt",
+            lambda *args: TestGUIPanels._actionPromptContinue(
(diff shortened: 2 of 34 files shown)


src/snapred/resources/application.yml
rboston628, line 1: These look like changes made during testing that should not be merged into the codebase

tests/integration/test_versions_in_order.py
```diff
@@ -220,6 +220,7 @@ def prepDiffractionCalibrationIngredients(self, ingredients: Any) -> Any:
 ###############################################################################
 
 
+# @pytest.mark.integration()  # TODO: why is this mark not used?!
```
rboston628, line 223: I'm not sure what the mark does, but this test was written to run in the usual unit tests

tests/unit/backend/data/test_DataFactoryService.py
```diff
@@ -88,17 +89,10 @@ def tearDown(self):
...
-        with tempfile.NamedTemporaryFile(suffix=".biscuit") as existent:
-            assert DataFactoryService().fileExists(existent.name)
-
-    def test_fileExists_no(self):
-        # assert that a file that does not exist, does not exist
-        with tempfile.TemporaryDirectory() as tmpdir:
-            nonexistent = tmpdir + "/0x0f.biscuit"
-            assert not os.path.isfile(nonexistent)
-            assert not DataFactoryService().fileExists(nonexistent)
+    def test_fileExists(self):
```
rboston628, line 92: Why did you change these tests?

rboston628 (commented):
Are the changes to `@Singleton` necessary for the integration test, or can they be moved to a separate PR?

tests/integration/test_workflow_panels_happy_path.py
```diff
@@ -0,0 +1,1351 @@
...
+    @staticmethod
+    def _actionPromptContinue(title, message, action, parent=None, match=r".*"):  # noqa: ARG004
+        _pattern = re.compile(match)
+        if not _pattern.match(message):
+            pytest.fail(
+                f"unexpected: ActionPrompt.prompt('{title}', '{message}'...)\n"
+                + f"    expecting:  ActionPrompt.prompt(...'{message}'...)"
+            )
+
+    @pytest.mark.integration()
```
dlcaballero16, line 217: Since you have a marker at the class level, the marker here is not required

dlcaballero16:
I would consider updating .github/workflows/actions.yml to explicitly not run the integration tests. So on line 38 of actions.yml, add to the python call so it's `python -m pytest -m "not integration"`

src/snapred/resources/application.yml
ekapadi: You are correct.  Those should not still be in there.

tests/integration/test_versions_in_order.py
ekapadi: I think eventually we'll need different levels of integration tests, so the `integration` mark could then have a "level" attribute (or something like that).  Right now, the mark allows us to bundle all of the integration tests together, and _not_ run them in the same session that the unit tests are run in.  And, this is not what you want for this test, so you've answered my question, and I'll remove the comments.  :)

tests/unit/backend/data/test_DataFactoryService.py
ekapadi: These changes make them consistent with the implementation style of the other tests in `test_DataFactoryService`.  Why I ended up needing to mess with these two tests specifically is a good question, but there was a reason.
I'd go even further and argue that this method shouldn't even exist, and I'm tempted to "nuke" the whole thing: `pathlib.Path.exists()` is sufficient.

ekapadi:
> Are the changes to `@Singleton` necessary for the integration test, or can they be moved to a separate PR?

Yes they're necessary.  No, I don't advise trying to separate them out.  This was actually the most difficult part of getting the integration test to work.  (Everything else was just finicky "timing" issues.)

There are actually several issues:

1. **Way too many classes in SNAPRed are `@Singleton` decorated** -- this includes classes which don't even retain any local data!  I'll note that in _standard practice_ singleton is viewed as an anti-pattern.  As a good counter-example: the implementation of `Indexer` actually uses a more _useful_ approach -- for its singleton aspect it depends on the singleton aspect of its owner (i.e. parent) class (i.e. if it _has_ a singleton aspect).  Unfortunately, when `@Singleton` is turned off, this means that after a state is newly initialized, only that specific `Indexer` has that information, and any indexers already owned by (i.e. in the `lru_cache` within) other services, end up not having the info.  (**For this issue specifically, we possibly should make any `lru_cache` using method _static_.**  I have not done this.  **However, this change is what I'd recommend doing as a separate PR.**  As an aside, another good reason to do this is it facilitates proper _garbage collection_ of `lru_cache`-using classes. )
3. **I needed to be able to launch the integration tests using the normal test framework,** and that meant that there needed to be a way to change the behavior of `@Singleton` for our normal unit tests.  This is the real issue:  we have relied on turning `@Singleton` _off_ because too many of our mocks are set up and are never _removed_.  In any case I arrived at a compromise:  the `@Singleton` is turned on, but only for the duration of the test `scope="function"`.  (This allows us to not have to fix all of our mocks at once.)  **Alternatively, we could use the new methods of `Singleton` to _resume_ our previous behavior of turning the `@Singleton` _off_ completely for the unit tests.**  I don't recommend this.  It should never have been turned off in the first place.
4. **Turning `@Singleton` on allows tests (even unit tests) to be run against classes which _depend_ on the singleton behavior.**
5. **Pytest does not, by design, allow controlling of the import sequence using test markers.**  This means the behavior of `@Singleton` can only be adjusted in certain ways.  Our previous system of entirely mocking out the decorator was implemented as part of the _import_ process, and as such, was somewhat at odds with the pytest design scheme.  (At least as I currently understand it.)

Even considering all of the previous, the solution I came up with still requires these integration tests to be run separately, in their own test _session_.

ekapadi:
> I would consider updating .github/workflows/actions.yml to explicitly not run the integration tests. So on line 38 of actions.yml, add to the python call so it's `python -m pytest -m "not integration"`

I'll add this change.  Note that it will be _redundant_ with what is also in our `pyproject.toml`.

tests/integration/test_workflow_panels_happy_path.py
ekapadi: I'll add this change.

tests/integration/test_versions_in_order.py
rboston628: It might be fine to put this in integration tests, provided we run them as part of the CI/CD.  It is a bit slower than the other unit tests (but still less than 30s), so makes sense to move.
If we get golden data for this test, it could also serve for validation of the diffcal process.

tests/unit/backend/data/test_DataFactoryService.py
rboston628: My understanding is `fileExists()` is meant to be a data service endpoint as part of the layer strategy.  Does this new test check for existing and non-existing files?  I can't tell at a glance.

dlcaballero16 (approved):

PR #462, opened 2024-10-01
Ewm7051 add rebinragged back

dlcaballero16:
## Description of work

RebinRagged loading/saving has been fixed on the Mantid side, so RebinRagged has been added back.

## Explanation of work

In ApplyNormalizationRecipe, we were using ResampleX, which has now been replaced with RebinRagged. Since the loading/saving has been fixed, we no longer need any of the Leftover functions, so those have been removed. In LocalDataService, writeRaggedWorkspace() has been removed and replaced with writeWorkspace().

### Dev testing

Make sure unit tests pass. I would also complete a reduction run and make sure a nexus file gets saved on completion.

### CIS testing

Make sure you can complete a reduction run, and the workspace behaves like you expect it to. See the criteria at the bottom.

## Link to EWM item

[EWM#7051](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=7051)

### Verification

### Acceptance Criteria

- [x] 1. The reduction workflow shall be completed on two run-numbers from a calibrated state (specific example: 61991,61398)
    DiffCal Params for 61991:
        Convergence Threshold: 0.1
        Grouping: All
        Sample: La11B6_NIST_660c_001
        dMin: 0.9
        dMax: 1.18
        Max Chi Sq: 2000
   DiffCal Params for 61398:
        Convergence Threshold: 0.1
        Grouping: All
        Sample: La11B6_NIST_660c_001
        dMin: 1.13
        dMax: 1.23
        Max Chi Sq: 2000
        
- [x] 2. The final reduced workspaces for all pixel groups are ragged workspaces
- [x] 3. The math operations: Minus, Multiply, Divide and Add can be conducted with the two reduced workspaces (from the same pixel grouping).
- [x] 4. The output of these operations are sensible (same dimensions as input, values are real numbers)
- [x] 5. The reduced data can be saved to disk using SaveNexus and reloaded using LoadNexus.


ekapadi (approved):
1. All of the workflow panels still work as expected -- this probably should have been an "acceptance criterion";
2. All of the acceptance criteria are met, and the resulting workspaces and operation values are as expected.  Caveat: reduced-data workspaces with the "all" grouping only contain one spectrum, and therefore are cannot be "ragged".

mguthriem:
Just a comment on the acceptance criteria: for specified run 61991, the sample is not LaB6, it's nist silicon. If anyone tries to run difcal it will fail.

Also, run 61937 is _vanadium_ it's not the best run to test reduction as it doesn't have any peaks, but should work fine for inspecting binning

PR #471, opened 2024-10-10
Fix algo importing so that they may be imported through `mantid.simpleapi` from other algos

rboston628:
## Description of work

Following a design meeting several weeks ago, it was decided `MantidSnapper` is better-used within `Recipe`s, and `Algorithm`s should only be implemented to handle the features requiring direct manipulation of workspace components.

In order to achieve that, we sometimes need for SNAPRed algorithms to import and call other SNAPRed algorithms without `MantidSnapper`.

## Explanation of work

The `__init__.py` file loads all the algorithms then puts them into `mantid.simpleapi`.  However, it does this in order.  If one algorithm first in the order needs one later in the order, that algo is not in `mantid.simpleapi` so the import fails.

This puts a `try-catch` block around the loading, and if an algorithm fails to load from `mantid.simpleapi`, it loads it first then tries again.

### Dev testing

Run all tests locally.

Launch the GUI locally.  Just launching it should be enough.

### CIS testing

N/A

## Link to EWM item

None

### Verification

N/A

### Acceptance Criteria

N/A

### DIFF
--- a/src/snapred/__main__.py
+++ b/src/snapred/__main__.py
@@ -2,6 +2,7 @@ import sys
 
 from mantid.kernel import amend_config
 
+import snapred.backend.recipe.algorithm  # noqa: F401
 from snapred import __version__ as snapred_version
 from snapred.backend.log.logger import snapredLogger
 from snapred.meta.Config import Resource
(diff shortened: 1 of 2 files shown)


dlcaballero16 (approved):
Tests pass and GUI still works.

PR #472, opened 2024-10-11
Enable MantidSnapper and algos to use pointer properties

rboston628:
## Description of work

Will make seamless use of pydantic objects as arguments to algos using pointers instead of string-cassting.

## Explanation of work

Algos can now possess "pointer" properties by using `ULongLongPropertyWithValue`.  

If any algorithm is supposed to take a `ULongLongPropertyWithValue` as an argument, MantidSnapper will automagically handle the casting to/from pointers.  There is currently nothing in mantid or SNAPRed that is using `ULongLongProperty` for any other reason, so this should be safe.

The `pointer.py` module should be used to create and access these objects.  The `create_pointer` function will also store that pointer in a cache so that a leash remains on it until it is accessed again.  The `access_pointer` removes it from the cache, so these pointers are only supposed to be used for passing to/from `PythonAlgorithm`s.

### Dev testing

Run all tests locally.

Perform manual testing of just the diffcal process up to saving (this is only one implicated by changes)

### CIS testing

N/A

## Link to EWM item

N/A

### Acceptance Criteria

N/A

### DIFF
--- a/src/snapred/backend/recipe/algorithm/MantidSnapper.py
+++ b/src/snapred/backend/recipe/algorithm/MantidSnapper.py
@@ -2,6 +2,7 @@ from collections import namedtuple
 
 from mantid.api import AlgorithmManager, Progress, mtd
 from mantid.kernel import Direction
+from mantid.kernel import ULongLongPropertyWithValue as PointerProperty
 
 from snapred.backend.error.AlgorithmException import AlgorithmException
 from snapred.backend.log.logger import snapredLogger
@@ -9,6 +10,7 @@ from snapred.backend.log.logger import snapredLogger
 # must import to register with AlgorithmManager
 from snapred.meta.Callback import callback
 from snapred.meta.Config import Resource
+from snapred.meta.pointer import access_pointer, create_pointer
 
 logger = snapredLogger.getLogger(__name__)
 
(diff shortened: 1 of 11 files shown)


src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -148,7 +144,9 @@ def executeAlgorithm(self, name, outputs, **kwargs):
                     val = val.get()
                 if val is None:
                     continue
-
+                # turn pydantic objects into pointers
+                if isinstance(algorithm.getProperty(prop), PointerProperty):
+                    val = create_pointer(val)
```
walshmm, line 149: do we want to type check `val` as well and ensure its actually a pydantic object?  or should the comment be updated to generically refer to pyobjects?

src/snapred/backend/recipe/algorithm/RemoveEventBackground.py
```diff
@@ -64,18 +74,14 @@ def chopIngredients(self, predictedPeaksList: List[GroupPeakList]):
             self.maskRegions[peakList.groupID] = []
             for peak in peakList.peaks:
                 self.maskRegions[peakList.groupID].append((peak.minimum, peak.maximum))
-
         # get handle to group focusing workspace and retrieve all detector IDs in each group
         focusWSname: str = str(self.getPropertyValue("GroupingWorkspace"))
-        focusWS = self.mantidSnapper.mtd[focusWSname]
-        self.groupIDs: List[int] = [int(x) for x in focusWS.getGroupIDs()]
+        # get a list of the detector IDs in each group
+        self.groupDetectorIDs = access_pointer(GroupedDetectorIDs(focusWSname))
```
walshmm, line 80: This line confuses me, we are accessing the result of GroupedDetectorIDs?  I understand its output is a pointer (after looking up the Algo) but it feels a bit obtuse to have to unwrap it via some other functional import
Pointers are also not common enough in our code for this pattern to be apparent yet.

Could you perhaps do something like this to make it more obvious:
```
from mantid.kernel import ULongLongPropertyWithValue as PointerProperty

...
        
        result : PointerProperty = GroupedDetectorIDs(focusWSname)
        self.groupDetectorIDs = access_pointer(result)
```

src/snapred/backend/recipe/algorithm/RemoveEventBackground.py
```diff
@@ -90,38 +96,35 @@ def PyExec(self):
         self.log().notice("Extracting background")
 
         # get the peak predictions from user input
-        detectorPeaksJson = self.getPropertyValue("DetectorPeaks")
-        typeAdapter = TypeAdapter(List[GroupPeakList])
-        predictedPeaksList = typeAdapter.validate_json(detectorPeaksJson)
+        predictedPeaksList = access_pointer(self.getProperty("DetectorPeaks").value)
```
walshmm, line 99: same as above?

src/snapred/backend/recipe/algorithm/__init__.py
```diff
@@ -12,13 +12,16 @@ def loadModule(x):
     from mantid.api import AlgorithmFactory
     from mantid.simpleapi import _create_algorithm_function
 
-    while True:
+    for i in range(30):
```
walshmm, line 15: could you clarify the why `30` is the desired range in a comment?

src/snapred/meta/pointer.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def create_pointer(thing: Any) -> int:
+    """
+    Returns a memory address to be used as a pointer, and retains a handle to the object
+    to ensure the memory can be accessed later.
+    @param the object whose pointer is required
+    @return the pointer to the object
+    """
+    OBJCACHE[id(thing)] = thing
```
walshmm, line 14: so this is mostly so it doenst get garbage collected?

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -37,36 +39,30 @@ class MantidSnapper:
...
+           kdddxkxxxxxkOOOO0000KKKOdllllloooooooodk0KXXNNNXXXXXKKXXKK00kxdodxO0         xdooolccccc::;';
+                kxxxxxkkkkkkk0XXXXKklccloodddddkOKKXXXXXXKXXKK00OOOkxdoollodxO             xoolcccc::;';
+                     kxddoddk0000K0OdoooooddxkO00KK0000000OOkkddooolllllooodxxk              kdlcc:::;,'
+                            OOkkxxxxxddddddddxxxxxxxxddxxxxxxdollllllllloddxxddk                xl:;;;,':
+                                    xlllldkOOO00000            kdooooddooddxxdddk                  xo:,''
+                                     xlllodk
+                                      kollloO
+                                       kolloO
+                                        0dlo0
+                                          OO
```
walshmm, line 65: fish update!!!

src/snapred/meta/pointer.py
rboston628: correct, without this there are often segfaults

src/snapred/backend/recipe/algorithm/__init__.py
rboston628: It was just a good guess to try alot but without infitnite loops.  Thinking on it, the best number is the length of the module list.

src/snapred/backend/recipe/algorithm/RemoveEventBackground.py
rboston628: I can do this, but it's worth noting that the result is not a "PointerProperty" but an integer, so the type isn't really right... but since type hints are only just that anyway, it works okay.

src/snapred/backend/recipe/algorithm/MantidSnapper.py
rboston628: This is probably the most significant change in this PR

src/snapred/backend/recipe/algorithm/MantidSnapper.py
rboston628: Val should be any python object, not limited to pydantic

walshmm (approved):

PR #480, opened 2024-10-23
Purge bad peaks :mountain: :axe:

rboston628:
## Description of work

The diffcal workflow will not allow users to proceed beyond the peak tweak stage if there are _any_ bad peaks.  This requires users to adjust things to try to make all peaks good.  But sometimes bad peaks just can't go away.

This adds a button which makes the bad peaks go away.

The purged list is then passed forward to all future calculations.

## Explanation of work

Added an extra service endpoint.  The older `calibration/diffraction` endpoint will take a DiffCalRequest, create the ingredients and groceries from it, and pass it to the recipe.  The new `calibration/calibrate` endpoint is expected a request with the ingredients and groceries already in it.  This allows for refining the ingredients before the calibration begins.

Within the diffraction workflow, there is now logic to remove peaks above the $\chi^2$ threshold.

This logic is triggered by the "Purge Bad Peaks" button.

### Dev testing

Run diff cal with the following:
- run: 58882
- sample: silicon
- grouping: Bank

At tweak peak, you should see two groups each with three bad peaks.  Press to purge the peaks.

The red filled regions have vanished.  There are still three good peaks.  Now continue to the saving screen.

Inspect the workspaces present, in particular from the PDCalibration diagnosis group.  In the `_params` table, for instance, you should see six rows for the six peaks (three in two groups).  The column for peak index should go 0, 1, 2 twice, while workspace index 0, 0, 0, 1, 1, 1.  The `_fitted` workspace should show two graphs with three peaks each.  Etc.

Check the other fit peaks outputs (cyan = workspace group = fit peaks output).

Save, run again.  At Tweak Peak, **the previously purged peaks will be back**.

Run with 46680 and Bank, and try purging peaks.  It will fail, as it would result in no peaks on the right grouping.

### CIS testing

Perform the dev steps above, but please also confirm the values, the goodness of fits and DIFC values, and check for some other runs.

## Link to EWM item

[EWM#6242](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6242)

### Verification

### Acceptance Criteria

- [x] TweakPeak view has a button to purge bad peaks
- [x] Clicking this button removes peaks with $\chi^2$ > threshold from the peaks list, then re-graphs with the bad peaks removed
- [x] If removing peaks would result in fewer than two peaks in a spectrum, an error dialog box is raised and no peaks are removed from that spectrum
- [x] The same peaks list, with the removed peaks, is passed forward to all future operations in the diffcal workflow

### DIFF
--- a/src/snapred/ui/workflow/DiffCalWorkflow.py
+++ b/src/snapred/ui/workflow/DiffCalWorkflow.py
@@ -1,4 +1,8 @@
+from mantid.simpleapi import mtd
 from qtpy.QtCore import Slot
+from qtpy.QtWidgets import (
+    QMessageBox,
+)
 
 from snapred.backend.dao import RunConfig
 from snapred.backend.dao.indexing.IndexEntry import IndexEntry
@@ -14,10 +18,12 @@ from snapred.backend.dao.request import (
     FitMultiplePeaksRequest,
     FocusSpectraRequest,
     HasStateRequest,
+    SimpleDiffCalRequest,
 )
 from snapred.backend.dao.SNAPResponse import SNAPResponse
 from snapred.backend.error.ContinueWarning import ContinueWarning
 from snapred.backend.log.logger import snapredLogger
+from snapred.backend.recipe.algorithm.FitMultiplePeaksAlgorithm import FitOutputEnum
 from snapred.meta.Config import Config
 from snapred.meta.decorators.ExceptionToErrLog import ExceptionToErrLog
 from snapred.meta.mantid.AllowedPeakTypes import SymmetricPeakEnum
@@ -87,10 +93,13 @@ class DiffCalWorkflow(WorkflowImplementer):
         self._requestView.litemodeToggle.field.connectUpdate(self._switchLiteNativeGroups)
         self._requestView.runNumberField.editingFinished.connect(self._populateGroupingDropdown)
         self._tweakPeakView.signalValueChanged.connect(self.onValueChange)
+        self._tweakPeakView.signalPurgeBadPeaks.connect(self.purgeBadPeaks)
 
         self.prevFWHM = DiffCalTweakPeakView.FWHM
         self.prevXtalDMin = DiffCalTweakPeakView.XTAL_DMIN
         self.prevXtalDMax = DiffCalTweakPeakView.XTAL_DMAX
+
+        self.peaksWerePurged = False
 
         # 1. input run number and other basic parameters
         # 2. display peak graphs, allow adjustments to view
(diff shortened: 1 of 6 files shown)


src/snapred/backend/service/SousChef.py
```diff
@@ -115,7 +116,7 @@ def prepPixelGroup(self, ingredients: FarmFreshIngredients) -> PixelGroup:
                 timeOfFlight=data["tof"],
                 nBinsAcrossPeakWidth=ingredients.nBinsAcrossPeakWidth,
             )
-        return self._pixelGroupCache[key]
+        return deepcopy(self._pixelGroupCache[key])
```
walshmm, line 119: https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.copy + `deep` kwarg?

rboston628: They aren't all pydantic models, some are lists

walshmm: fair enough

src/snapred/ui/workflow/DiffCalWorkflow.py
```diff
@@ -340,6 +357,54 @@ def _renewFitPeaks(self, peakFunction):
...
+            self._renewFitPeaks(self.peakFunction)
+
+            # update graph with reduced peak list
+            self._tweakPeakView.updateGraphs(
+                self.focusedWorkspace,
+                self.ingredients.groupedPeakLists,
+                self.fitPeaksDiagnostic,
+            )
+
+        except Exception as e:  # noqa BLE001
```
walshmm, line 401: could this not be solved by spinning up a worker like https://github.com/neutrons/SNAPRed/blob/523b781248e9dd24164b7c04a6db736cee72cdaf/src/snapred/ui/presenter/WorkflowPresenter.py#L172-L176 ?

walshmm:
image
image
Presto!
image
image

image
This failed but I'm not sure it was the expected reason?

rboston628:
46680 should use Diamond.  When you purge peaks it should give a "sorry Dave" message and do nothing

src/snapred/ui/workflow/DiffCalWorkflow.py
rboston628: I can try fixing it in this PR, or I can make a defect to fix it

walshmm:
> This story adds a button to the graphical interface of Diffraction Calibration Workflow, in both of the tweak peak views (pre-fit and post-pixelcal) 

Not quite sure what this bit of scope refers to, is there supposed to be a new button in two separate views?

rboston628:
That bit of scope is because the other story I'm working on (7388) will move the view to between pixel calibration and group calibration.  I think this was written assuming 7388 would be finished first.

walshmm:
image
using the diamond sample I managed to fail successfully :confetti_ball:

src/snapred/ui/workflow/DiffCalWorkflow.py
rboston628: Could I spin a defect to handle fixing both of these?

walshmm: I'd be fine with that, its not pressing

walshmm (approved):

PR #487, opened 2024-10-28
Ewm7121 refactor art norm

darshdinger:
## Description of work
The purpose for this PR is the implement the previously created artificial normalization within the `Reduction` workflow. The changes ensure the workflow can handle missing normalization data by applying artificial normalization and introduce enhancements to the user interface, logging, and data handling processes.

## Explanation of work
Key updates include:

1. Artificial Normalization:
* Added the `ArtificialNormalizationRecipe` to `GenericRecipe` that uses the `CreateArtificialNormalizationAlgo` to handle cases when normalization data is missing.
* Automatically extracts parameters and queues the algorithm to create an artificial normalization workspace when needed.

2. Workflow Updates:
* The `Reduction` workflow can now dynamically adjust for missing calibration or normalization data, applying artificial normalization when necessary.
* Updated the UI to prompt users for artificial normalization inputs and seamlessly continue the workflow.

3. Error Handling and Logging:
* Improved logging and error messages to guide users through situations where data is missing.
* Enhanced exception handling to gracefully manage missing states or permissions.

4. Backend Improvements:
* Refactored the `ReductionService` to handle normalization and calibration more flexibly.
* Expanded unit tests to cover new cases with artificial normalization.

## To test
Insure all pytests pass. Please launch SNAPRed UI with a fresh (no states initialized) local copy of `Calibration_next` specified within `application.yml`. `First` run the Diffraction Calibration with the following inputs:

* Run Number: 58882
* Convergence Threshold: 0.1
* Bins Across Peak Width: 10
* Sample: Silicon_NIST_640D_001.json
* Grouping File: Column

Create a new state when prompted. Then continue on through workflow through successful completion.
Now moving onto the Reduction tab, notice there is a new tab called "Artificial Normalization". Press the `Continue` button and notice the UI handling for no inputs:

image

 Finally, `without` running a normalization, run the `Reduction` workflow with the following inputs:

* Run Number: 58882

Continue the workflow to notice the following message pop-up:

image

Continue Past the possible warning for permissions check but clicking `Yes`. Find that the workflow completes successfully. 

`Finally`, re-run this but with a normalization this time and see that the `Artificial Normalization` flow is not used in this case.
### Dev testing
Insure all pytests pass

### CIS testing
See above.

## Link to EWM item

[EWM # 7123](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=7123)

&

[...]

PR #489, opened 2024-10-31
actually applies the artifical norm this time, bases the A.N. on redu…

walshmm:
…ction data as per Malcolm, makes A.N. for EACH group, smoothing field allows for appropriate value >1

## Description of work

This reorgs and applies things so that Artificial Normalization is effective during reduction.

## To test

Refer to #487 for testing params but refer to this ticket for acceptance criteria.

## Link to EWM item

[EWM#8026](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=8026)

### Verification

### Acceptance Criteria

- [x]  Artificial Normalization is actually applied to reduction results. (check this with an extreme artificial normalization.)
- [x]  Smoothing Parameter is a text field to allow freedom to test appropriate parameters for a future slider.
- [x]  Artificial Normalization should be based on the input run for reduction, and one should be made for each group.
- [x]  Due to limitations of the ui, the preview Artificial Normalization will use a Column grouping when deciding the parameters.
- [x]  Lacking a calibration should not prevent a user from proceeding with and Artificial Normalization

### DIFF
--- a/src/snapred/backend/dao/request/ReductionRequest.py
+++ b/src/snapred/backend/dao/request/ReductionRequest.py
@@ -2,6 +2,7 @@ from typing import List, NamedTuple, Optional, Tuple
 
 from pydantic import BaseModel, ConfigDict, field_validator
 
+from snapred.backend.dao.ingredients import ArtificialNormalizationIngredients
 from snapred.backend.dao.state.FocusGroup import FocusGroup
 from snapred.backend.error.ContinueWarning import ContinueWarning
 from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceName
@@ -23,7 +24,7 @@ class ReductionRequest(BaseModel):
     versions: Versions = Versions(None, None)
 
     pixelMasks: List[WorkspaceName] = []
-    artificialNormalization: Optional[str] = None
+    artificialNormalizationIngredients: Optional[ArtificialNormalizationIngredients] = None
 
     # TODO: Move to SNAPRequest
     continueFlags: Optional[ContinueWarning.Type] = ContinueWarning.Type.UNSET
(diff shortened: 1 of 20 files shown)


src/snapred/backend/dao/request/ReductionRequest.py
darshdinger, line 1: This makes a lot more sense.

darshdinger:
# Testing Overview:
--------------------------
1. Attempt Reduction without state initialized leads to this error:
image

2. Proceed with fresh state:
image
Then this error once the processing is complete:
image
The workflow succeeds regardless:
image

3. Trial with diffcal present:
I get this error and no artificial normalization pop up for no normalization present:
image

4. Trial with both normcal and diffcal present:
Reduction seems to proceed as normal:
image


Based on my tests, there are some present issues with the error pop ups for notifying the user.

walshmm:
1. I need to reword the error since it trys to punt you to the state init pop up but you dont have write perms.  It shouldnt be referencing the reduction save path.
2. I think I just need to silence that **warning**?  I'm not sure its actionable from the user or relevant.


These _warnings_ shouldnt be blocking.
Please proceed in reviewing the data and gui functionality as outlined in the AC while I address the inaccurate error message and noisy warning pop up.

darshdinger:
# Testing AC:
------------------

1. Run Reduction without Normalization:
image

2. Testing Smoothing Parameter:
Original - 
image
Trying a different smoothing parameter -
image

The artificial normalization changed a bit in terms of the smoothness but in all cases attempted, it doesn't match up to the diffcal data. Column grouping is present as per AC. Text field is usable and functional as per AC.

4. Final Result:
image
Seems like it has been applied to each grouping schema.

5. As tested before, lacking a calibration does not prevent a user from proceeding with an artificial normalization.

darshdinger (approved):
Through retesting of PR after additional changes results in the expected behavior. The implementation of additional exceptions is functional.

PR #490, opened 2024-11-04
Added use of MetadataLookupService in the frontend

dlcaballero16:
## Description of work

This change adds the use of the MetadataLookupService to the front end so that the multi-run number validation occurs with that service.

## Explanation of work

In the ReductionRequestView, the MetadataLookupService is called to validate multiple run numbers. This allows the user to pass in multiple runs like a Mantid IntArrayProperty.

### Dev testing

You should now be able to enter a list of runs to reduce in 4 different ways:

StartRun-EndRun
StartRun:EndRun
Functionally, these are the same, they will both return a list of runs, each run incrementing by 1 until the end run number is reached.

StartRun:EndRun:step
Instead of incrementing by 1, the list will increment by whatever step is. For example,
1:5:2
will return
1,3,5

item1,item2
Can be combined with ranges like so: 
1,3-5
will return
1,3,4,5

Since this is just testing the input, I would test each of the 4 types of input, but do not worry about actually being able to reduce multiple runs. Just pay attention to the error message and make sure the list of runs is as you expect it. For example, when I input 
46680-46688
as my runs to reduce locally, I get an error that GetIPTS fails, but the error box contains a full list of runs of 46680-46688.

### CIS testing

See Dev testing.

## Link to EWM item

[EWM#6324](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=6324)

### Verification

### Acceptance Criteria

- [ ] acceptance criterion 1
- [ ] acceptance criterion 2


darshdinger:
# Testing:
-------------

1. StartRun-EndRun:
image
Results in:
image

2. StartRun:EndRun:
image
Results in:
image

3. StartRun:EndRun:step
image
Results in:
image

4. Item1,Item2
image
Results in:
image


5. Some incorrect case testing:
image
Results in:
image
**This needs to throw the response that states that the run number is too low.**

image
Results in:
image
**Application crashed, need handling for this.**

I think the work left is just handling incorrect inputs without crashing.

darshdinger:
# ReRun Previous Failed Testing
--------------------------------------------
1. Try with unsupported run numbers:
image
Results in:
image

darshdinger (approved):
Passed the final round of testing. Pop-ups are as expected.

PR #500, opened 2024-11-18
Handle multiple reduction runs

rboston628:
## Description of work

The Artificial Normalization screen requires the workflow to load a screen for adjusting parameters to create a false background.  However, workflow's don't enable going forward or backward or optionally skipping screens, which prevents handling arbitrary cases of Artificial Normalization in run lists.

There are two "happy path" situations which may be considered, and which are enabled in this PR.  
1. user has single run requiring Artificial Normalization
2. user has many runs, all using previously calculated normalizations from disk

The other paths have been turned off by giving the user a warning.

## Explanation of work

The CalibrationService and NormalizationService were updated with the ability to load a set of calibration/normalizations for a list of run numbers.  Also, return a dictionary matching a list of runs to each corresponding version.

If all runs have a matching normalization version (which is not None), then reduction works as normal over the entire list, as intended.

If any of the runs requires artificial normalization, then the workflow will throw an error dialog box requiring the user to enter only a single run.

### Dev testing

**Setup**

Download files 46802, 46803, 46805.  They are in IPTS-24641 and under 1GB.  These are in a new state, so use SNAPRed to create a calibration and a normalization for these runs.  They do not have to be good.  Just use 46803 and whichever sample and step through it until both save.

Run from WORKBENCH and not just stand-alone GUI.

**Hide the normalization**

Hide the normalizations you just made, most easily by changing the directory name.  Run reduction with 46802, 46803, 46805.  SNAPRed will not let you, and require you to clear the run list and try again with a single run.

Run reduction with just 46802.  It will warn about needing artificial normalization.  Continue, create an art norm, then continue to the end of the workflow.

**Unhide the normalization**

Close SNAPRed.

Navigate to the `NormalizationIndex.json` file for this state.  Edit the "appliesTo" field to "46802".

Run normalization with 46802, 46803, 46805.  It will complain that you must have either all or none of the runs needing artificial normalization.

Edit the "appliesTo" field to ">=46802".

Run normalization with 46802, 46803, 46805.  It will run smoothly and without issue.  Observe that the grouping workspaces, calibration tables, and normalizations are not being deleted, until the very end.  At the end, you will have output workspaces for all three runs.

### CIS testing

Follow steps similar to the above, but using the original runs 61991:61993, and valid normalizations and calibrations.  They are located in `/SNS/SNAP/shared/Calibration_next`.

## Link to EWM item

[EWM#8287](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=8287)

See also:
[EWM#8230](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=8230)

### Verification

### Acceptance Criteria

- [x] When running reduction on severall runs (must be >=3 to reproduce original error) which all have valid calibrations and normalizations, SNAPRed automatically reduces all three without stopping on any screens or asking the user for further continuation
- [x] There are no crashes and the Mantid crash screen never comes up (run in workbench to verify this)
- [x] At end of reduction process, output workspaces for all three runs in the list are present.
- [x] The case of Artificial Normalization is handled intelligently (here, because we can only perform on one run at a time, don't let the user try with more than one run)

### DIFF
--- a/src/snapred/backend/dao/request/MatchRunsRequest.py
+++ b/src/snapred/backend/dao/request/MatchRunsRequest.py
@@ -1,0 +1,8 @@
+from typing import List
+
+from pydantic import BaseModel
+
+
+class MatchRunsRequest(BaseModel):
+    runNumbers: List[str]
+    useLiteMode: bool
--- a/src/snapred/backend/service/CalibrationService.py
+++ b/src/snapred/backend/service/CalibrationService.py
@@ -29,6 +29,7 @@ from snapred.backend.dao.request import (
     FocusSpectraRequest,
     HasStateRequest,
     InitializeStateRequest,
+    MatchRunsRequest,
     SimpleDiffCalRequest,
 )
 from snapred.backend.dao.response.CalibrationAssessmentResponse import CalibrationAssessmentResponse
@@ -99,6 +100,7 @@ class CalibrationService(Service):
         self.registerPath("group", self.groupCalibration)
         self.registerPath("validateWritePermissions", self.validateWritePermissions)
         self.registerPath("residual", self.calculateResidual)
+        self.registerPath("fetchMatches", self.fetchMatchingCalibrations)
         return
 
     @staticmethod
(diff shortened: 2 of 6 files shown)


darshdinger:
Code looks fine. Here is the testing:

## Testing
------------
1. Didn't download any runs, did DiffCal and NormCal for run 61991 from scratch:
image

2. Hide Normalization:
image

3. Run Reduction, but it won't let me:
image

4. Run Artificial Norm for only 61991:
image

5. Unhide norm:
image

6. Modify normalizationIndex.py:
image

7. Rerun Reduction for 61991:61993:
image

With workspace:
image

darshdinger (approved):
Looks good.

PR #503, opened 2024-11-19
Cache groupings by state

rboston628:
## Description of work

A grouping should apply to all runs with the same state.  However, currently they are cached by run number.  Caching by run number causes reduction to re-load the same groupings multiple times.

This changes how grouping workspaces are cached inside `GroceryService` to better streamline reduction.

## Explanation of work

Changes `GroceryService` to cache grouping workspaces by state ID, instead of run number.

This requires special handling for the `LiteMap`, which uses a reserved run number.  Most of the changes are to make handling the reserved run numbers more robust.

### Dev testing

To verify that SNAPRed is still correctly handling grouping workspaces inside the grocery cache:

1. Pick two runs from the same state.  I recently found 46802, 46803, 46805, which are relatively small files.
2. Open diffcal panel, run with 46802.  Use any sample, use any grouping (my favorite is Bank).  Continue to tweak peak stage.
3. Verify that the grouping workspace exists in the workspace list, and that its name includes 46802 and Bank (see image).
4. Verify the data is grouped correctly (see image).
image
5. Close the calibration panel window.  Do NOT cancel or continue, just close the window so that the workspace list and grocery cache are maintained.
5. Open diffcal panel again, run with 46803.  Use any sample, pick the _same_ grouping as before (e.g. Bank).  Continue to tweak peak stage.
6. Verify that the data has been grouped correctly (see image).
7. Verify that the grouping workspace with 46802 exists in workspace list (see image).
8. Verify there is NOT a grouping workspace with 46803 in its name (see image).
image

This verifies that `GroceryService` is correctly caching and using grouping workspaces by their state ID and not by their run number.

Now run calibration and normalization  with _any_ run number (e.g. 46680) to verify everything works.

In Reduction, pick a set of three runs from same state with a normalization (I like 46802, 46803, 46805).  Run with these.  As reduction is calculating, observe the workspace list.  The same set of groupings should remain for all three runs, and not be deleted.  The other temporary workspaces should disappear.  These groupings will disappear when the reduction workflow finishes.

**NOTE**: this story does not fix a defect that has already been noted in [Defect 8287](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=8287) related to running with multiple runs and artificial normalizations.  You can **not** run multiple runs in reduction without a prior normalization for the state,

### CIS testing

1. Run calibration with a preferred run
2. Run normalization with preferred set of runs
3. Run reduction with two runs.

## Link to EWM item

There is no story for this, but it came out of work on [Defect 8287](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=8287)

### Verification

N/A

### Acceptance Criteria

N/A

### DIFF
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -25,7 +25,7 @@ from snapred.backend.recipe.FetchGroceriesRecipe import FetchGroceriesRecipe
 from snapred.backend.service.WorkspaceMetadataService import WorkspaceMetadataService
 from snapred.meta.Config import Config
 from snapred.meta.decorators.Singleton import Singleton
-from snapred.meta.InternalConstants import ReservedRunNumber, ReservedStateId
+from snapred.meta.InternalConstants import ReservedRunNumber
 from snapred.meta.mantid.WorkspaceNameGenerator import (
     NameBuilder,
     WorkspaceName,
@@ -863,10 +863,7 @@ class GroceryService:
 
         :rtype: Dict[str, Any]
         """
-        if item.runNumber in ReservedRunNumber.values():
-            stateId = ReservedStateId.forRun(item.runNumber)
-        else:
-            stateId, _ = self.dataService.generateStateId(item.runNumber)
+        stateId, _ = self.dataService.generateStateId(item.runNumber)
         key = self._key(item.groupingScheme, stateId, item.useLiteMode)
         workspaceName = self._createGroupingWorkspaceName(item.groupingScheme, item.runNumber, item.useLiteMode)
         workspaceName = self._loadedGroupings.get(key, workspaceName)
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -25,6 +25,7 @@ from snapred.backend.recipe.FetchGroceriesRecipe import FetchGroceriesRecipe
 from snapred.backend.service.WorkspaceMetadataService import WorkspaceMetadataService
 from snapred.meta.Config import Config
 from snapred.meta.decorators.Singleton import Singleton
+from snapred.meta.InternalConstants import ReservedRunNumber, ReservedStateId
 from snapred.meta.mantid.WorkspaceNameGenerator import (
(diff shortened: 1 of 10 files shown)


ekapadi (commented):
From my comment on slack:

My initial response is that possibly we should name the grouping workspaces based on the first or last 8-characters of the <state SHA>, in what is the <run number> position right now?  In WNG there's already one of the reduction name templates that uses this, so the possibility to do so might be fairly easy to add.  (That being said.  Are we even using the WNG to name the grouping workspaces now?  And unfortunately, if we do this possibly we need to check for a hash collision.  :disappointed: )

tests/resources/application.yml:  state-id needs to be a 16-character hex digest, so naming it anything like "reserved-state" (fighting against slack formatting here)  won't work.   (This is an "issue" waiting to happen when any of us _mock_ something a bit more thoroughly during unit testing.)  As before, I'd also prefer to see any internal "magic" constants NOT in "application.yml":  could the reserved state-ID actually be in one of the DAO python files somewhere?  The exception to this is only stuff that affects the user interface (like possibly our choice of version-numbering scheme).

Everything else looks just fine, but I need to run through your suggested tests.

src/snapred/backend/data/GroceryService.py
```diff
@@ -865,13 +863,17 @@ def fetchGroupingDefinition(self, item: GroceryListItem) -> Dict[str, Any]:
 
         :rtype: Dict[str, Any]
         """
-        key = self._key(item.groupingScheme, item.runNumber, item.useLiteMode)
+        if item.runNumber in ReservedRunNumber.values():
+            stateId = ReservedStateId.forRun(item.runNumber)
+        else:
+            stateId, _ = self.dataService.generateStateId(item.runNumber)
```
rboston628, line 869: Question for reviewer: Should this got here, or inside `LocalDataService.generateStateId`?

PR #505, opened 2024-11-20
Ewm8086 remove background  by smoothing before cross-correlation

rboston628:
## Description of work

This improves `RemoveEventBackground`, and renames it to `RemoveSmoothedBackground`
- it is faster
- it makes use of SmoothDataExcludingPeaks, rather than reimplementing almost the same work
- the data are properly binned without edge effects and without negative values

## Explanation of work

It was determined that inside each group, an averaged background can be calculated by averaging all the spectra in that group then using SmoothDataExcludingPeaks to make a group average background.

This background is subtracted from each individual spectra in the group so that only the peaks remain.  After the subtraction, all negatives are zeroed out.

To prevent complications from weak pixels, the background is suitably scaled prior to subtraction, with the scale factor determined by the mean values across the averaged and individual spectra.  This step is crucial in keeping peaks in pixels with very weak signal.

### Dev testing

You will need an otherwise good calibration run with high background.  I suggest 58882.

Run without remove background, and at tweak peak take a screen shot of the graphs.  Also look at the mask workspace (`diffract_consts_mask_58882`), and note how many bins are masked.

Run with remove background (note: you may have to clear all the workspaces for this to happen).  Compare the new graphs at tweak peak.  Notice the lumpy background is gone, with much sharper peaks.

Check masks.  Fewer pixels have been masked.  Complete the workflow and save.

### CIS testing

First, run the script `check_remove.py`.  This tests the background removal algorithm only.  Inspect the workspaces `peaks_extractDSP_before` and `peaks_extractDSP_after`,  and their focused counterparts (+`_foc`).  Make sure everything is as expected.

Next, run the script `cc-and-smooth.py`.  This mirrors SNAPRed's cross-correlation process, including both the group-by-group cross-correlation and the convergence by median offset.  This tests three different ways of removing the background.  The ones of interest will have the prefix "total" and "peaks".  Look at the CC result (+`_cc`), the offsets, the before and after, etc.

Finally, run `diffcal_pixel_diffraction_background_subtraction_script.py`.  This will run SNAPRed's PixelDiffCalRecipe with background removal turned on.  Inspect the results.

## Link to EWM item

[EWM#8086](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=8086)

### Verification

### Acceptance Criteria

- [x] The algorithm RemoveSmoothedBackground creates and subtracts a smoothed background form each spectrum
- [x] Any negative numbers from the subtraction are set to zero 
- [x] The binning used in the Rebin step should be changed at L59 to self.dBin: float = min([abs(d) for d in dBin])


src/snapred/backend/recipe/PixelDiffCalRecipe.py
```diff
@@ -51,13 +51,14 @@ def chopIngredients(self, ingredients: Ingredients) -> None:
...
-        dMin = ingredients.pixelGroup.dMin()
-        dMax = ingredients.pixelGroup.dMax()
-        dBin = ingredients.pixelGroup.dBin()
+        dMin: List[float] = ingredients.pixelGroup.dMin()
+        dMax: List[float] = ingredients.pixelGroup.dMax()
+        dBin: List[float] = ingredients.pixelGroup.dBin()
         self.overallDMin: float = min(dMin)
         self.overallDMax: float = max(dMax)
-        self.dBin: float = max([abs(d) for d in dBin])
+        self.dBin: float = min([abs(d) for d in dBin])
```
rboston628, line 59: Note: this line satisfies one of the acceptance criteria

src/snapred/backend/recipe/algorithm/RemoveSmoothedBackground.py
```diff
@@ -0,0 +1,198 @@
...
+            y_smooth = smoothWS.readY(wkspindx).copy()
+            y_data = focusWS.readY(wkspindx).copy()
+            scale_denom = sum(y_data) / len(y_data)
+            for detid in self.groupDetectorIDs[groupID]:
+                # subtract off the background and update the date in the workspace
+                y_new = outputWS.readY(detid).copy()
+                scale_num = sum(y_new) / len(y_new)
+
+                y_new = y_new - (scale_num / scale_denom) * y_smooth
+                y_new[y_new < 0] = 0
```
rboston628, line 173: Note: this line satisfies one of the acceptance criteria

rboston628:
MERGE AFTER PR #523

PR #509, opened 2024-11-25
Ewm8253 normcal index being misread

walshmm:
## Description of work

Requires this data change to be merged to pass integration tests:
https://code.ornl.gov/sns-hfir-scse/infrastructure/test-data/snapred-data/-/merge_requests/1

## Explanation of work

It was found that the index was not appropriately failing when a version for a specified run was not found.
This was in part due to its subbing in the "current" version if it received `None` as an input, which it also happened to return when it couldnt find a version.  This then created a scenario in which not finding a record resulted in pulling the `current` version.

The `current` version being arbitrarily the last one done.

## To test

Run through the full range of workflows and see that they work.
Then mess with the calibration/normalization indexes and see how the reduction workflow responds.  Confirm that it matches up with what is expected.

## Link to EWM item

[EWM#8253](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=8253)

### Verification

### Acceptance Criteria

to replicate:
1. Set calibration home = `/SNS/SNAP/shared/Calibration_next`
2. Reduce run 57482 (in lite mode)
3. Data are reduced, using version 2 of normcal in the folder `/SNS/SNAP/shared/Calibration_next/Powder/04bd2c53f6bf6754/lite/normalization/`

However, if the normalizationIndex.json is inspected its content should mean that there is no normcal available for run 57482:

```
[
    {
        "runNumber": "58809",
        "useLiteMode": true,
        "appliesTo": ">=58882",
        "comments": "test of v1.1.0rc1",
        "author": "mg",
        "timestamp": 1729258319.2569995,
        "version": 1
    },
    {
        "runNumber": "58809",
        "useLiteMode": true,
        "appliesTo": ">=58809",
        "comments": " 3mm V-Nb sample",
        "author": "Malcolm",
        "timestamp": 1730229952.2204628,
        "version": 2
    }
]
```
The normcal is only available for run numbers >=58809.

- [x] The expected behaviour would be that SNAPRed gives a warning that normcal does not exist for run 57482 and proceeds to manage that situation.

### DIFF
--- a/src/snapred/backend/dao/indexing/Versioning.py
+++ b/src/snapred/backend/dao/indexing/Versioning.py
@@ -1,79 +1,50 @@
-from typing import Any, Optional
-
-from numpy import integer
-from pydantic import BaseModel, computed_field, field_serializer
-
+from pydantic import BaseModel, ConfigDict, field_validator
 from snapred.meta.Config import Config
+from snapred.meta.Enum import StrEnum
 
 VERSION_START = Config["version.start"]
 VERSION_NONE_NAME = Config["version.friendlyName.error"]
-VERSION_DEFAULT_NAME = Config["version.friendlyName.default"]
 
-# VERSION_DEFAULT is a SNAPRed-internal "magic" integer:
-# * it is implicitely set during `Config` initialization.
-VERSION_DEFAULT = Config["version.default"]
+
+class VersionState(StrEnum):
+    DEFAULT = Config["version.friendlyName.default"]
+    LATEST = "latest"
+    NEXT = "next"
+
+
+# I'm not sure why ci is failing without this, it doesn't seem to be used anywhere
+VERSION_DEFAULT = VersionState.DEFAULT
+
+Version = int | VersionState
 
 
 class VersionedObject(BaseModel):
     # Base class for all versioned DAO
 
-    # In pydantic, a leading double underscore activates
-    # the `__pydantic_private__` feature, which limits the visibility
-    # of the attribute to the interior scope of its own class.
-    __version: Optional[int] = None
+    version: Version
 
-    @classmethod
-    def parseVersion(cls, version, *, exclude_none: bool = False, exclude_default: bool = False) -> int | None:
-        v: int | None
-        # handle two special cases
-        if (not exclude_none) and (version is None or version == VERSION_NONE_NAME):
-            v = None
-        elif (not exclude_default) and (version == VERSION_DEFAULT_NAME or version == VERSION_DEFAULT):
-            v = VERSION_DEFAULT
-        # parse integers
-        elif isinstance(version, int | integer):
-            if int(version) >= VERSION_START:
-                v = int(version)
-            else:
-                raise ValueError(f"Given version {version} is smaller than start version {VERSION_START}")
-        # otherwise this is an error
-        else:
-            raise ValueError(f"Cannot initialize version as {version}")
-        return v
+    @field_validator("version", mode="before")
+    def validate_version(cls, value: Version) -> Version:
+        if value in VersionState.values():
+            return value
 
-    @classmethod
-    def writeVersion(cls, version) -> int | str:
-        v: int | str
-        if version is None:
-            v = VERSION_NONE_NAME
-        elif version == VERSION_DEFAULT:
-            v = VERSION_DEFAULT_NAME
-        elif isinstance(version, int | integer):
-            v = int(version)
-        else:
-            raise ValueError("Version is not valid")
-        return v
+        if isinstance(value, str):
+            raise ValueError(f"Version must be an int or {VersionState.values()}")
 
-    def __init__(self, **kwargs):
-        version = kwargs.pop("version", None)
-        super().__init__(**kwargs)
-        self.__version = self.parseVersion(version)
+        if value is None:
+            raise ValueError("Version must be specified")
 
-    @field_serializer("version", check_fields=False, when_used="json")
-    def write_user_defaults(self, value: Any):  # noqa ARG002
-        return self.writeVersion(self.__version)
+        if value < VERSION_START:
+            raise ValueError(f"Version must be greater than {VERSION_START}")
 
-    # NOTE some serialization still using the dict() method
-    def dict(self, **kwargs):
-        res = super().dict(**kwargs)
-        res["version"] = self.writeVersion(res["version"])
-        return res
+        return value
 
-    @computed_field
-    @property
-    def version(self) -> int:
-        return self.__version
+    # NOTE: This approach was taken because 'field_serializer' was checking against the
+    #       INITIAL value of version for some reason.  This is a workaround.
+    #
+    def model_dump_json(self, *args, **kwargs):  # noqa ARG002
+        if self.version in VersionState.values():
+            raise ValueError(f"Version {self.version} must be flattened to an int before writing to JSON")
+        return super().model_dump_json(*args, **kwargs)
 
-    @version.setter
-    def version(self, v):
-        self.__version = self.parseVersion(v, exclude_none=True)
+    model_config = ConfigDict(use_enum_values=True, validate_assignment=True)
--- a/src/snapred/backend/dao/indexing/Versioning.py
+++ b/src/snapred/backend/dao/indexing/Versioning.py
@@ -1,4 +1,5 @@
 from pydantic import BaseModel, ConfigDict, field_validator
+
 from snapred.meta.Config import Config
 from snapred.meta.Enum import StrEnum
 
--- a/src/snapred/backend/dao/normalization/NormalizationRecord.py
+++ b/src/snapred/backend/dao/normalization/NormalizationRecord.py
@@ -1,9 +1,9 @@
 from typing import Any, List
 
-from pydantic import field_serializer, field_validator
+from pydantic import field_validator
 
 from snapred.backend.dao.indexing.Record import Record
-from snapred.backend.dao.indexing.Versioning import VERSION_DEFAULT, VersionedObject
+from snapred.backend.dao.indexing.Versioning import VERSION_START, Version, VersionedObject
 from snapred.backend.dao.Limit import Limit
 from snapred.backend.dao.normalization.Normalization import Normalization
 from snapred.meta.mantid.WorkspaceNameGenerator import WorkspaceName
@@ -31,7 +31,7 @@ class NormalizationRecord(Record, extra="ignore"):
     smoothingParameter: float
     # detectorPeaks: List[DetectorPeak] # TODO: need to save this for reference during reduction
     workspaceNames: List[WorkspaceName] = []
-    calibrationVersionUsed: int = VERSION_DEFAULT
+    calibrationVersionUsed: Version = VERSION_START
     crystalDBounds: Limit[float]
     normalizationCalibrantSamplePath: str
 
@@ -44,8 +44,4 @@ class NormalizationRecord(Record, extra="ignore"):
     @field_validator("calibrationVersionUsed", mode="before")
     @classmethod
     def version_is_integer(cls, v: Any) -> Any:
-        return VersionedObject.parseVersion(v)
-
-    @field_serializer("calibrationVersionUsed", when_used="json")
-    def write_user_defaults(self, value: Any):  # noqa ARG002
-        return VersionedObject.writeVersion(self.calibrationVersionUsed)
+        return VersionedObject(version=v).version
--- a/src/snapred/backend/data/DataFactoryService.py
+++ b/src/snapred/backend/data/DataFactoryService.py
@@ -161,10 +161,6 @@ class DataFactoryService:
     def getLatestApplicableNormalizationVersion(self, runId: str, useLiteMode: bool):
         return self.lookupService.normalizationIndexer(runId, useLiteMode).latestApplicableVersion(runId)
 
-    @validate_call
-    def getLatestNormalizationVersion(self, runId: str, useLiteMode: bool):
-        return self.lookupService.normalizationIndexer(runId, useLiteMode).latestApplicableVersion(runId)
-
     ##### REDUCTION METHODS #####
 
     @validate_call
--- a/src/snapred/backend/data/DataFactoryService.py
+++ b/src/snapred/backend/data/DataFactoryService.py
@@ -1,11 +1,12 @@
 import os
 from pathlib import Path
-from typing import Dict, List, Optional
+from typing import Dict, List
 
 from pydantic import validate_call
 
 from snapred.backend.dao.calibration.CalibrationRecord import CalibrationRecord
 from snapred.backend.dao.indexing.IndexEntry import IndexEntry
+from snapred.backend.dao.indexing.Versioning import Version, VersionState
 from snapred.backend.dao.InstrumentConfig import InstrumentConfig
 from snapred.backend.dao.normalization.NormalizationRecord import NormalizationRecord
 from snapred.backend.dao.reduction import ReductionRecord
@@ -81,7 +82,7 @@ class DataFactoryService:
         return self.lookupService.calibrationExists(runId, useLiteMode)
 
     @validate_call
-    def getCalibrationDataPath(self, runId: str, useLiteMode: bool, version: int):
+    def getCalibrationDataPath(self, runId: str, useLiteMode: bool, version: Version):
         return self.lookupService.calibrationIndexer(runId, useLiteMode).versionPath(version)
 
     def checkCalibrationStateExists(self, runId: str):
@@ -102,28 +103,22 @@ class DataFactoryService:
         return self.lookupService.calibrationIndexer(runId, useLiteMode).getIndex()
 
     @validate_call
-    def getCalibrationRecord(self, runId: str, useLiteMode: bool, version: Optional[int] = None) -> CalibrationRecord:
+    def getCalibrationRecord(
+        self, runId: str, useLiteMode: bool, version: Version = VersionState.LATEST
+    ) -> CalibrationRecord:
         """
         If no version is passed, will use the latest version applicable to runId
         """
         return self.lookupService.readCalibrationRecord(runId, useLiteMode, version)
 
     @validate_call
-    def getCalibrationDataWorkspace(self, runId: str, useLiteMode: bool, version: int, name: str):
+    def getCalibrationDataWorkspace(self, runId: str, useLiteMode: bool, version: Version, name: str):
         path = self.lookupService.calibrationIndexer(runId, useLiteMode).versionPath(version)
         return self.groceryService.fetchWorkspace(os.path.join(path, name) + ".nxs", name)
 
     @validate_call
-    def getThisOrCurrentCalibrationVersion(self, runId: str, useLiteMode: bool, version: Optional[int] = None):
-        return self.lookupService.calibrationIndexer(runId, useLiteMode).thisOrCurrentVersion(version)
-
-    @validate_call
-    def getThisOrNextCalibrationVersion(self, runId: str, useLiteMode: bool, version: Optional[int] = None):
-        return self.lookupService.calibrationIndexer(runId, useLiteMode).thisOrNextVersion(version)
-
-    @validate_call
-    def getThisOrLatestCalibrationVersion(self, runId: str, useLiteMode: bool, version: Optional[int] = None):
-        return self.lookupService.calibrationIndexer(runId, useLiteMode).thisOrLatestApplicableVersion(runId, version)
+    def getLatestApplicableCalibrationVersion(self, runId: str, useLiteMode: bool):
+        return self.lookupService.calibrationIndexer(runId, useLiteMode).latestApplicableVersion(runId)
 
     ##### NORMALIZATION METHODS #####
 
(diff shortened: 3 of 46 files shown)


src/snapred/backend/dao/indexing/Versioning.py
```diff
@@ -1,78 +1,50 @@
...
+class VersionState(StrEnum):
+    DEFAULT = Config["version.friendlyName.default"]
+    LATEST = "latest"
+    NEXT = "next"
+
+
+# I'm not sure why ci is failing without this, it doesn't seem to be used anywhere
+VERSION_DEFAULT = VersionState.DEFAULT
+
+Version = int | VersionState
```
rboston628, line 18: I want to draw @ekapadi 's attention to this change

src/snapred/backend/data/DataFactoryService.py
```diff
@@ -102,36 +103,30 @@ def getCalibrationIndex(self, runId: str, useLiteMode: bool) -> List[IndexEntry]
...
-        return self.lookupService.calibrationIndexer(runId, useLiteMode).thisOrCurrentVersion(version)
-
-    @validate_call
-    def getThisOrNextCalibrationVersion(self, runId: str, useLiteMode: bool, version: Optional[int] = None):
-        return self.lookupService.calibrationIndexer(runId, useLiteMode).thisOrNextVersion(version)
-
-    @validate_call
-    def getThisOrLatestCalibrationVersion(self, runId: str, useLiteMode: bool, version: Optional[int] = None):
-        return self.lookupService.calibrationIndexer(runId, useLiteMode).thisOrLatestApplicableVersion(runId, version)
+    def getLatestCalibrationVersion(self, runId: str, useLiteMode: bool):
```
rboston628, line 120: change name to `getLatestApplicableCalibrationVersion` for consistency

src/snapred/backend/data/DataFactoryService.py
```diff
@@ -149,28 +144,26 @@ def getNormalizationIndex(self, runId: str, useLiteMode: bool) -> List[IndexEntr
...
 
     @validate_call
-    def getThisOrCurrentNormalizationVersion(self, runId: str, useLiteMode: bool, version: Optional[int] = None):
-        return self.lookupService.normalizationIndexer(runId, useLiteMode).thisOrCurrentVersion(version)
-
-    @validate_call
-    def getThisOrNextNormalizationVersion(self, runId: str, useLiteMode: bool, version: Optional[int] = None):
-        return self.lookupService.normalizationIndexer(runId, useLiteMode).thisOrNextVersion(version)
+    def getLatestApplicableNormalizationVersion(self, runId: str, useLiteMode: bool):
+        return self.lookupService.normalizationIndexer(runId, useLiteMode).latestApplicableVersion(runId)
```
rboston628, line 162: How does this method differ from L165-166 below?

src/snapred/backend/data/DataFactoryService.py
```diff
@@ -149,28 +144,26 @@ def getNormalizationIndex(self, runId: str, useLiteMode: bool) -> List[IndexEntr
...
 
     @validate_call
-    def getThisOrLatestNormalizationVersion(self, runId: str, useLiteMode: bool, version: Optional[int] = None):
-        return self.lookupService.normalizationIndexer(runId, useLiteMode).thisOrLatestApplicableVersion(runId, version)
+    def getLatestNormalizationVersion(self, runId: str, useLiteMode: bool):
+        return self.lookupService.normalizationIndexer(runId, useLiteMode).getLatestApplicableVersion(runId)
```
rboston628, line 166: How does this method differ from L161-162 above?

src/snapred/backend/data/GroceryService.py
```diff
@@ -333,14 +334,14 @@ def createDiffcalTableWorkspaceName(
...
     ) -> WorkspaceName:
         """
-        NOTE: This method will IGNORE runNumber if the provided version is VERSION_DEFAULT
+        NOTE: This method will IGNORE runNumber if the provided version is VersionState.DEFAULT
         """
         wsName = wng.diffCalTable().runNumber(runNumber).version(version).build()
-        if version == VERSION_DEFAULT:
-            wsName = wsName = wng.diffCalTable().runNumber("default").version(VERSION_DEFAULT).build()
+        if version in [VersionState.DEFAULT, VERSION_START]:
+            wsName = wsName = wng.diffCalTable().runNumber("default").version(VersionState.DEFAULT).build()
```
rboston628, line 344: `wsName = wsName = `??

src/snapred/backend/data/Indexer.py
```diff
@@ -129,7 +126,7 @@ def reconcileIndexToFiles(self):
                     f"The following records were expected, but not available on disk: {missingRecords}"
                 )
             else:
-                logger.warn(f"The following records were expected, but not available on disk: {missingRecords}")
+                logger.warning(f"The following records were expected, but not available on disk: {missingRecords}")
```
rboston628, line 129: This warning could be improved by asking the user to consult an IS to locate the records or to edit the index

src/snapred/backend/data/Indexer.py
```diff
@@ -145,7 +142,7 @@ def defaultVersion(self) -> int:
         """
         The version number to use for default states.
         """
-        return VERSION_DEFAULT
+        return VERSION_START
```
rboston628, line 145: Why is the default version beginning at the start?

src/snapred/backend/data/Indexer.py
```diff
@@ -186,65 +183,33 @@ def nextVersion(self) -> int:
...
+            return self.currentVersion() + 1
 
     @validate_call
-    def thisOrLatestApplicableVersion(self, runNumber: str, version: Optional[int]):
-        if self.isValidVersion(version) and self._isApplicableEntry(self.index[version], runNumber):
-            return version
-        else:
-            return self.latestApplicableVersion(runNumber)
+    def getLatestApplicableVersion(self, runNumber: str):
+        return self.latestApplicableVersion(runNumber)
```
rboston628, line 197: do we really need `getLatestApplicableVersion` **and** `latestApplicableVersion`?

src/snapred/backend/data/Indexer.py
```diff
@@ -186,65 +183,33 @@ def nextVersion(self) -> int:
...
-    def isValidVersion(self, version):
+    def validateVersion(self, version):
         try:
-            VersionedObject.parseVersion(version, exclude_none=True)
+            VersionedObject(version=version)
             return True
         except ValueError:
-            return False
+            # This error would only ever result from a software bug.
+            # Saving/Loading/Refering to erroneous "current" versions just serves to obfuscate the error.
```
rboston628, line 205: I'm pretty sure it could also result from someone manually editing either the index or record files

src/snapred/backend/data/Indexer.py
```diff
@@ -277,24 +242,20 @@ def indexPath(self):
         """
         return self.rootDirectory / f"{self.indexerType}Index.json"
 
-    def recordPath(self, version: Optional[int] = None):
+    def recordPath(self, version: int):
```
rboston628, line 245: Should this type be `Version`?

src/snapred/backend/data/Indexer.py
```diff
@@ -277,24 +242,20 @@ def indexPath(self):
...
         """
         Path to a specific version of a calculation record
         """
         return self.versionPath(version) / f"{self.indexerType}Record.json"
 
-    def parametersPath(self, version: Optional[int] = None):
+    def parametersPath(self, version: int):
```
rboston628, line 251: As above

src/snapred/backend/data/Indexer.py
```diff
@@ -277,24 +242,20 @@ def indexPath(self):
...
         """
         return self.versionPath(version) / f"{self.indexerType}Parameters.json"
 
-    @validate_call
-    def versionPath(self, version: Optional[int] = None) -> Path:
-        if version is None:
-            version = VERSION_START
-        else:
-            version = self.thisOrCurrentVersion(version)
+    def versionPath(self, version: int) -> Path:
```
rboston628, line 257: As above

src/snapred/backend/data/Indexer.py
```diff
@@ -342,68 +303,111 @@ def addIndexEntry(self, entry: IndexEntry):
...
-            version=self.thisOrNextVersion(version),
+            version=self._flattenVersion(version),
             **other_arguments,
         )
         record.calculationParameters.version = record.version
         return record
 
-    def _determineRecordType(self, version: Optional[int] = None):
-        version = self.thisOrCurrentVersion(version)
+    def _determineRecordType(self, version: int):
```
rboston628, line 320: Should `version` be ` Version`?

src/snapred/backend/data/Indexer.py
```diff
@@ -342,68 +303,111 @@ def addIndexEntry(self, entry: IndexEntry):
...
         recordType = None
-        if version == VERSION_DEFAULT:
+        if version == self.defaultVersion():
             recordType = DEFAULT_RECORD_TYPE.get(self.indexerType, None)
         if recordType is None:
             recordType = RECORD_TYPE[self.indexerType]
         return recordType
 
-    def readRecord(self, version: Optional[int] = None) -> Record:
+    def readRecord(self, version: int) -> Record:
```
rboston628, line 328: As above

src/snapred/backend/data/Indexer.py
```diff
@@ -342,68 +303,111 @@ def addIndexEntry(self, entry: IndexEntry):
...
+                f"No record found at {filePath} for version {version}, latest version is {self.currentVersion()}"
+            )
         return record
 
+    def _flattenVersion(self, version: Version):
+        flattenedVersion = None
+        if version == VersionState.DEFAULT:
+            flattenedVersion = self.defaultVersion()
+        elif version == VersionState.LATEST:
+            flattenedVersion = self.currentVersion()
```
rboston628, line 347: Why would `latest` return the current version? The adjective `latest` otherwise refers to the most recent calendar date that applies to the run

src/snapred/backend/data/Indexer.py
```diff
@@ -342,68 +303,111 @@ def addIndexEntry(self, entry: IndexEntry):
...
+            raise ValueError(
+                f"No available versions found during lookup using: "
+                f"v={version}, index={self.index}, dir={self.dirVersions}"
+            )
+        return flattenedVersion
+
+    def versionExists(self, version: Version):
+        return self._flattenVersion(version) in self.index
+
+    def writeNewRecord(self, record: Record, entry: IndexEntry):
```
rboston628, line 365: `writeIndexAndRecord`?  (and swap order of arguments)

src/snapred/backend/data/LocalDataService.py
```diff
@@ -142,7 +142,7 @@ def _readInstrumentParameters(self) -> Dict[str, Any]:
             raise _createFileNotFoundError("Instrument configuration file", self.instrumentConfigPath) from e
 
     def readStateConfig(self, runId: str, useLiteMode: bool) -> StateConfig:
-        diffCalibration = self.calibrationIndexer(runId, useLiteMode).readParameters()
+        diffCalibration = self.calibrationIndexer(runId, useLiteMode).readParameters(VersionState.LATEST)
```
rboston628, line 145: This will return the `current` index.  Should it?

src/snapred/backend/data/LocalDataService.py
```diff
@@ -457,7 +457,8 @@ def _indexer(self, stateId: str, useLiteMode: bool, indexerType: IndexerType):
 
     def indexer(self, runNumber: str, useLiteMode: bool, indexerType: IndexerType):
         stateId, _ = self.generateStateId(runNumber)
-        return self._indexer(stateId, useLiteMode, indexerType)
+        indexer = self._indexer(stateId, useLiteMode, indexerType)
+        return indexer
```
rboston628, line 461: Is anything gained by splitting this?

src/snapred/backend/data/LocalDataService.py
```diff
@@ -481,28 +482,37 @@ def writeNormalizationIndexEntry(self, entry: IndexEntry):
...
+            entryParams["version"] = VersionState.NEXT
+        return indexer.createIndexEntry(**entryParams)
 
     def createNormalizationRecord(self, request: CreateNormalizationRecordRequest) -> NormalizationRecord:
         indexer = self.normalizationIndexer(request.runNumber, request.useLiteMode)
         return indexer.createRecord(**request.model_dump())
 
     def normalizationExists(self, runId: str, useLiteMode: bool) -> bool:
-        version = self.normalizationIndexer(runId, useLiteMode).currentVersion()
+        version = self.normalizationIndexer(runId, useLiteMode).getLatestApplicableVersion(runId)
```
rboston628, line 496: Again, should it be `latestApplicableVersion` or `getLatestApplicableVersion`?

src/snapred/backend/data/LocalDataService.py
```diff
@@ -481,28 +482,37 @@ def writeNormalizationIndexEntry(self, entry: IndexEntry):
...
         return version is not None
 
     @validate_call
-    def readNormalizationRecord(self, runId: str, useLiteMode: bool, version: Optional[int] = None):
+    def readNormalizationRecord(self, runId: str, useLiteMode: bool, version: Optional[Version] = VersionState.LATEST):
```
rboston628, line 500: It doesn't need to be `Optional` if it has a default

src/snapred/backend/data/LocalDataService.py
```diff
@@ -481,28 +482,37 @@ def writeNormalizationIndexEntry(self, entry: IndexEntry):
...
         """
         Will return a normalization record for the given version.
         If no version given, will choose the latest applicable version from the index.
         """
         indexer = self.normalizationIndexer(runId, useLiteMode)
-        if version is None:
+        if version is VersionState.LATEST:
```
rboston628, line 506: In one part you use `VersionState.LATEST` to reference the current version, but here it is referencing the latest applicable version.  Maybe we need another value in `VersionState`?

rboston628 (commented):
I got pretty far, but I still have more changes to review

src/snapred/backend/dao/indexing/Versioning.py
```diff
@@ -1,78 +1,50 @@
...
+            raise ValueError(f"Version must be greater than {VERSION_START}")
+
+        return value
+
+    # NOTE: This approach was taken because 'field_serializer' was checking against the
+    #       INITIAL value of version for some reason.  This is a workaround.
+    #
+    def model_dump_json(self, *args, **kwargs):  # noqa ARG002
+        if self.version in VersionState.values():
+            raise ValueError(f"Version {self.version} must be flattened to an int before writing to JSON")
```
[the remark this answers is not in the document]
walshmm: It requires external lookups to figure out what the actual flattened version is.  I'm keeping it decoupled from the indexer.

src/snapred/backend/dao/normalization/NormalizationRecord.py
```diff
@@ -44,8 +44,4 @@ def validate_backgroundRunNumber(cls, v: Any) -> Any:
     @field_validator("calibrationVersionUsed", mode="before")
     @classmethod
     def version_is_integer(cls, v: Any) -> Any:
-        return VersionedObject.parseVersion(v)
-
-    @field_serializer("calibrationVersionUsed", when_used="json")
-    def write_user_defaults(self, value: Any):  # noqa ARG002
-        return VersionedObject.writeVersion(self.calibrationVersionUsed)
+        return isinstance(VersionedObject(version=v).version, int)
```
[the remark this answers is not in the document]
walshmm: oh I see, I'll fix this

src/snapred/backend/data/LocalDataService.py
walshmm: It should not, this needs to be fixed.  I will be removing the LATEST -> current flattening, as that isnt correct and should never be used.  i.e. loading should always require a version, saving would never refer to 'current'

src/snapred/backend/data/LocalDataService.py
walshmm: I guess `latestApplicableVersion`? it doesnt actually have a setter method so a getter doesnt make as much sense

ekapadi (commented):
First of all, I will "ditto" most of Reece's questions, and I'd like to see the answers.
Secondly, this is a clear improvement over the current code, and I think that we should definitely merge these changes.

-------
However, that being said.  If this was completely "my own" codebase I would _simplify_ all of this _with extreme prejudice_.  

For example: version should just be an `int`.  Except for a few "very special" places in the code it should be _required_ and non optional.  It should be validated in one place only.  Outside of that location, `RuntimeError` should be raised if an unexpected value is detected, and rather than masking a defect, the program should shut down.  (It's our _separate_ responsibility as developers to ensure that something like this never happens in released code.)

I don't see much reason to _propagate_ the additional concepts of "start", "next", "latest" and / or "default" (sigh) outside of the very small region of code where the version is actually _initialized_ (for a new record), or looked up (for an existing record).  The fact that we _propagate_ these variants has caused us so much difficulty!

As a separate, but related issue, I am also gradually _removing_ "my own" default arguments on new methods which are intended for _developer_ use.  I've been observing that although sometimes the use of these arguments allows us to prototype more rapidly, in too many cases these default arguments have just gotten us into trouble.

src/snapred/backend/data/Indexer.py
walshmm: The internal vs external representation.  Though this indexer doesnt actually separate them totally.  
The concept of "default" probably shouldnt have bled into indexer, as that is more related to the external ways of using the indexer.  I.e. a Calibration has a default, which can be established using the Indexer, but Normalization does not.  In both cases the indexer should start at some arbitrary version and increment.

We could further flesh out the concept of "friendly names" in the future to support any special versions (an additional mapping field on the index?)

src/snapred/backend/data/Indexer.py
walshmm: but then we wouldnt be able to read said files as they would fail to serialize into a `VersionedObject`, thus would not reach this portion of the code. (fail to read probably means we needed to migrate versions or something)

Since it is validated on write/read, this step only applies to externally supplied input, which means either:
1. We failed to validate user input and it somehow made it this far down without error
2. We wrote code that submitted an invalid version(Bad coversion, typo, Point 1, ect)

src/snapred/backend/data/Indexer.py
walshmm: Disk should only have a flattened int representation, so int is probably best.

src/snapred/backend/data/Indexer.py
walshmm: again not, as above.  If its on disk it better be an int.

src/snapred/backend/data/LocalDataService.py
walshmm: The conversion of latest to current was a mistake.  We will actually be trying to remove these intermediates entirely in the future

src/snapred/backend/data/Indexer.py
walshmm: how about `writeNewVersion`?  These files need to be written in tandem due to current refresh behaviour.  
(semi-related: We should potentially just always read from disk)

src/snapred/backend/dao/indexing/Versioning.py
```diff
@@ -1,79 +1,51 @@
...
-
-from numpy import integer
-from pydantic import BaseModel, computed_field, field_serializer
+from pydantic import BaseModel, ConfigDict, field_validator
 
 from snapred.meta.Config import Config
+from snapred.meta.Enum import StrEnum
 
 VERSION_START = Config["version.start"]
 VERSION_NONE_NAME = Config["version.friendlyName.error"]
```
ekapadi, line 7: I'm wondering if we should put type annotations on these constants?  For example, I'm assuming that `VERSION_START: int` and `VERSION_NONE_NAME: str` and etc.

src/snapred/backend/dao/normalization/NormalizationRecord.py
```diff
@@ -31,7 +31,7 @@ class NormalizationRecord(Record, extra="ignore"):
     smoothingParameter: float
     # detectorPeaks: List[DetectorPeak] # TODO: need to save this for reference during reduction
     workspaceNames: List[WorkspaceName] = []
-    calibrationVersionUsed: int = VERSION_DEFAULT
+    calibrationVersionUsed: Version = VERSION_START
```
ekapadi, line 34: A _little_ bit confusing, as it's not obvious then that this means the default version.  Here I'm wondering if we should just _define_ `VERSION_DEFAULT = VERSION_START`?

ekapadi (commented):
OK, that's interesting.  **I ran your test on "next", and NOW it is recognizing that there's no normalization data available (it's asking if I want to run an artificial normalization).**  So, with this in mind I resurrect my previous question.   Can you get me (the details of) a specific <test that fails> (on next) and <test that succeeds> on your branch?

ekapadi (approved):
In addition to having verified all of the workflows separately,  I had just realized that it's possible to verify the changes of PR#509 using only the "46680" run:

1. (For "next" and also for the PR#509 branch) I re-initialize a new state
2. I run diffcal and normcal
3. for normcal I set "applies to" to ">46690"

=> On "next" this happily completes a reduction and mis-applies the existing normalization.

=> On the PR#509 branch it notes that no applicable normalization is available, and would I like to use an artificial normalization?

**As far as I'm concerned, this PR is now verified and approved.**  I'm certain that when Malcolm tests it he can create a few more complicated normalization cases, and there won't be any problems.  So, with that in mind, my suggestion would be to not spend any more time on it!

PR #514, opened 2024-12-05
Fix workflow presenter to run things from threads

rboston628:
## Description of work

There are some processes which take place inside workflows.  It is important these be able to return errors.   Many of these methods were not sent to separate threads, which meant errors could not be handled through the universal error handling point.

This fixes that, providing a method for all workflows to call to easily send methods to  new thread.

## Explanation of work

Inside `WorkflowPresenter`, split off the functionality that handles actions on continue to handle any kind of action.

Updated the workflows to make use of this functionality, instead of their prior `try-catch` blocks that only logged errors to the output.

### Dev testing

This impacts the following parts of the UI:
 - groupings populated in the drop down
 - recalculation on peak-tweak parameters
 - purging peaks from diffcal workflow

Make sure these parts of the workflow are still completely correctly.

### CIS testing

N/A

## Link to EWM item

There is no ticket for this work , but it supports other work in 

[EWM#7712](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=7712)

### Verification

N/A

### Acceptance Criteria

N/A

### DIFF
--- a/src/snapred/ui/presenter/WorkflowPresenter.py
+++ b/src/snapred/ui/presenter/WorkflowPresenter.py
@@ -1,4 +1,4 @@
-from typing import Callable, List
+from typing import Any, Callable, List, Tuple
 
 from qtpy.QtCore import QObject, Signal, Slot
 from qtpy.QtWidgets import QMainWindow, QMessageBox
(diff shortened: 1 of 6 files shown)


src/snapred/ui/presenter/WorkflowPresenter.py
```diff
@@ -164,10 +164,15 @@ def verifyAndContinue():
             return model.continueAction(self)
 
         # do action
-        self.worker = self.worker_pool.createWorker(target=verifyAndContinue, args=None)
+        continueOnSuccess = lambda success: self.advanceWorkflow() if success else None  # noqa E731
+        self.handleAction(verifyAndContinue, None, continueOnSuccess)
+
+    def handleAction(self, action, args, onSuccess):
```
walshmm, line 170: Please add typehints, I think its important we know onSuccess is a `Callable`

walshmm (approved):

PR #511, opened 2024-12-03
After completion of pixel calibration, show unfocused before and after 

rboston628:
## Description of work

Some users of SNAP want to see unfocused before/after workspaces, to help assess the usefulness of pixel calibration.

## Explanation of work

One of the workspaces already existed, but was only retained in CIS mode.

Another of the workspaces is created at the end of `PixelDiffCalRecipe`.

### Dev testing

Run calibration with 46680.  It will fail, but not before creating the workspaces.

When it finishes, within the list of workspaces should be `dsp_046880_raw_beforeCrossCor` and `dsp_046680_raw_afterCrossCor`.  Check the following:
- both workspaces have 18432 spectra
- both workspaces are in d-Spacing
- both workspaces have the same binning
- both workspaces can be viewed in sliceviewer by right clicking in the ADS

Right click on the `tof_all_lite_raw_046680` workspace, and select to view the detectors.  This will open a table workspace with the column DIFC.  

Then right click on "before" workspace and select the same.  Verify the DIFC columns are identical between the original and the "before"

Right click on the "after" workspace and select the same.  Verify the DIFC columns differ between the original and "after"

### CIS testing

Run this with 58882 and Silicon (or your own choice).

Verify as above.

Also make sure, in slice viewer, that pixel calibration has done anything worthwhile.

## Link to EWM item

[EWM#7712](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=7712)

### Verification

- [x] the author has read the EWM story and acceptance critera
- [x] the reviewer has read the EWM story and acceptance criteria
- [x] the reviewer certifies the acceptance criteria below reflect the criteria in EWM

### Acceptance Criteria

This list is for ease of reference, and does not replace reading the EWM story as part of the review.  Verify this list matches the EWM story before reviewing.

**NOTE** this PR only applies to steps 1 and 2 described in the EWM item

- [x] After completion of pixel calibration and before group calibration, the user has access to before/after workspaces
- [x] The before/after workspaces are unfocused (18432 spectra in lite or 1.2M spectra in native)
- [x] The before/after workspaces functions of d-spacing
- [x] The before/after workspaces have the same binning 
- [x] The before/after workspaces differ in that one has been converted to d-spacing using the input difcal convertion and the second has been converted using the difcal that include the offsets determined via the pixel calibration process
[...]

PR #517, opened 2024-12-06
Allow user to skip pixelcal after pixelcal

rboston628:
## Description of work

If the user performs pixel calibration, but sees that it did not do anything positive, allow the user from the tweak peak stage to elect to skip pixel calibration and recalculate.

## Explanation of work

After all the other changes, this was very trivial.

Add a new button to tweak peak view for skipping pixel calibration, and add a check if the user changes their initial selection.

If the initial selection changes, recalculate with/without pixel calibration.

### Dev testing

Run diffraction with 46680, skip pixel toggle set to False.

It should fail.  In the workspace list, open the two table workspaces, `diffract_consts_default_v0` and `diffract_const_046680`.  Compare, and ensure these are not equal workspaces.

Right-click on `diffract_const_046680` and rename it to something obvious so that it is not overwritten.

Set Skip Pixel Cal toggle to True and continue.  There may be a warning about a workspace already loaded.

Set Skip Pixel Cal toggle to False and press Recalculate.  The same warning will come up.  Close it.

Compare the new `diffract_const_046680` to the original.  They should be identical. This proves the workflow is being re-run without skipping pixel calibration.

Change the grouping and recalculate.  Make sure it works.

Change the grouping **and** a dMin and recalculate.  Make sure it works.

Change **only** a dMin and recalculate.  Make sure it works.

Continue to save.  You'll have to set Skip Pixel Cal to True and change the FWHM to 4 and Recalculate to continue.

### CIS testing

As above, but I suggest 58882.  Make sure to inspect the before/after files in slice viewer.  Make sure you can find them and they make sense.

## Link to EWM item

[EWM#7712](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=7712)

### Verification

- [X] the author has read the EWM story and acceptance critera

### Acceptance Criteria

**NOTE** Following conversation with Malcolm, the behavior of iterating pixelcal is not necessary as there are no parameters to adjust.

- [X] ~~after pixel calibration, user has available two d-spacing unfocussed workspaces reflecting separately the input calibration and the output calibration (i.e. application of the pixel-calibration-generated DIFC values).~~ **already met in PR #511**
- [X] ~~The user can continue to next stage of workflow if desired~~ **already met**
- [ ] Or, the user can repeat pixel calibration, opting to completey reset DIFCs to their original values
- [ ] The user has access to a UI element to select from the above options
- [ ] The user can repeat pixel calibration an arbitrary number of times (note: they can, but it doesn't do much)
- [ ] Conducting N iterations, where N>2 should not cause further increase in RAM usage

### DIFF
--- a/src/snapred/ui/view/DiffCalTweakPeakView.py
+++ b/src/snapred/ui/view/DiffCalTweakPeakView.py
@@ -60,6 +60,9 @@ class DiffCalTweakPeakView(BackendRequestView):
         self.signalRunNumberUpdate.connect(self._updateRunNumber)
         self.signalMaxChiSqUpdate.connect(self._updateMaxChiSq)
 
+        # skip pixel calibration toggle
+        self.skipPixelCalToggle = self._labeledToggle("Skip Pixel Calibration", False)
+
         self.continueAnyway = False
         self.signalContinueAnyway.connect(self._updateContinueAnyway)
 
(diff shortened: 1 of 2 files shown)


dlcaballero16 (approved):
Tested and functions as expected.

PR #519, opened 2024-12-10
fix inputs for Artificial Norm

walshmm:
## Description of work

Standardized the way we read bools from the true/false dropdowns, they were being read incorrectly
This might also inadvertantly located where we were going wrong for this retain workspaces bug. 

## To test

Initialize a state, run reduction on `59039` with art norm, confirm that in the preview window the art norm is updated when changing the flags from true to false.
Check the outputs with and without flags and confirm they are different.

## Link to EWM item

[EWM#8374](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=8374)

### Verification

### Acceptance Criteria

- [ ] This defect is to ensure the values specified in the UI match those that are applied (and the defaults should remain as they are: both true)


walshmm:
this changes went in with pr #520

PR #523, opened 2024-12-20
Update the use of pointers within certain algorithms

rboston628:
## Description of work

In anticipation of the eventual [`PythonObjectionProperty`](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=7807), it is necessary to begin refactoring algorithms to rely more on pointer properties instead of passing large pydantic objects as strings.

It was also deemed necessary that algorithms relying heavily on direct manipulation of workspaces outside of algorithms, should not use `MantidSnapper` and instead call mantid functionality directly.

This updates two of the algorithms, `SmoothDataExcludingPeaksAlgo` and `DiffractionSpectrumWeightCalculator` to pass the peak lists through pointers instead of strings.  It also makes some needed change to enable that, like how pointer properties are handled between `GenericRecipe` and `MantidSnapper`.

## Explanation of work

Changed some string properties being used to pass a list of grouped detector peaks, to instead by pointer properties passing the same lists by their pointer.

Added a new method to inspect pointers that does not remove them from the cache, to use in `validateInputs` methods.

Fixed `MantidSnapper` to not cast pointers into pointers; that is, if a pointer property's value is already a pointer, just pass that pointer and don't make another layer of pointers.

### Dev testing

Run in CIS mode so that the toggle to remove background can be flipped.

Run both diffraction and normalization and ensure nothing changed.

### CIS testing

N/A

## Link to EWM item

[EWM#8086](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=8086)

### Verification

### Acceptance Criteria

**NOTE** the other work in EWM 8086 is handled in PR #505.  This only handles the updates to `SmoothDataExcludingPeaksAlgo` called for in EWM 8086.

- [x]  update and refactor SmoothDataExcludingPeaksAlgo to make use of pointers, validators, and other improvements

### DIFF
--- a/src/snapred/backend/recipe/GenerateFocussedVanadiumRecipe.py
+++ b/src/snapred/backend/recipe/GenerateFocussedVanadiumRecipe.py
@@ -5,7 +5,6 @@ from snapred.backend.log.logger import snapredLogger
 from snapred.backend.recipe.RebinFocussedGroupDataRecipe import RebinFocussedGroupDataRecipe
 from snapred.backend.recipe.Recipe import Recipe
 from snapred.meta.decorators.Singleton import Singleton
-from snapred.meta.redantic import list_to_raw
 
 logger = snapredLogger.getLogger(__name__)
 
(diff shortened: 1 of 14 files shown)


dlcaballero16 (approved):
Tested and was able to complete diffcal and normalization.

PR #526, opened 2025-01-08
Simple fix to Defect 8554

rboston628:
## Description of work

Within the Reduction workflow, if a calibration table is present and has a mask, that mask should be applied.

However, it wasn't, unless the user specified a pixel mask.

There were two issues:
1. mismatch of the name of the mask between the service and recipe
2. a spurious `if` in the service which ended up not loading the diffcal mask unless the user had specified another one to load.

## Explanation of work

Inside `ReductionRecipe`, just changed the name to match.

Inside `ReductionService`, had to fix the logic around which pixel masks are loaded.  This all takes place inside the `prepCombinedMask` function, making things a bit cleaner.  This will *always* load the diffcal mask, if there is one, and then load all user-specified pixel masks, and combine all of the loaded masks at the end.  The "combined" mask might be a combination of just one mask, and might have no masked pixels.

`ReductionService` will only pass this mask if it has masked pixels.  This requires a method to check the mask workspace.  This was added to `GroceryService`, as this seems the most obvious place for it.

Additional tests ensure:
- the mask workspace created in the service correctly sets the mask workspace inside the recipe (preventing the original issue)
- even if no pixel masks are sent, the diffcal mask is still loaded (preventing the original issue)
- prepCombinedMask pulls in all the masks it should and properly combines them
- fetchReductionGroceries only passes the combined mask workspace when it is non-empty
- the check on if a mask workspace is empty works correctly

### Dev testing

The added unit tests should ensure the defect behavior is fixed.

### CIS testing

Run reduction for a run with a diffcal mask and some available pixel masks.  Make sure they are properly combined.

## Link to EWM item

[EWM#8554](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=8554)

### Verification

### Acceptance Criteria

- [x] in Reduction workflow, if there is a diffcal mask present, it is loaded and applied to the workspace, even if the user does not specify pixel masks.

### DIFF
--- a/src/snapred/backend/dao/state/DetectorState.py
+++ b/src/snapred/backend/dao/state/DetectorState.py
@@ -1,6 +1,6 @@
 from enum import IntEnum
 from numbers import Number
-from typing import Literal, Tuple
+from typing import Dict, Literal, Tuple
 
 from pydantic import BaseModel, field_validator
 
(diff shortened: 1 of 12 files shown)


ekapadi (commented):
I went over the code changes, and mostly I'm fine with things!  I like the way that this streamlines the `ReductionRecipe` combine-mask loop.  

One thing I'd prefer, is that any _utility_ method used to create a mask, use a template workspace to get its instrument -- that's what we need to do in most places right now.  

We definitely _also_ need your new XML-instrument method, but that should be  a separate thing -- and, if you want, it could be triggered by an optional "flag" arg to the mask-from-array method (or something like `if instrumentDonor is None: ...` or `if templateWorkspace is None: ...`.   Somewhat ironically, I'll note that after my work on the effective-instrument stuff:  "instrument 2.0" is probably going to end up moving _away_ from XML instruments.

 When I create a mask for a test, I'm essentially _always_ concerned that its instrument will match that from another workspace that I'm using as input.  So this means the fact that an XML-instrument is created _for_ me, and I don't have a choice, wouldn't be a benefit in most cases.

Among small comments (and there will be more),  I'll note that I'd generally recommend  using `isinstance(workspace, MaskWorkspace)` rather than `workspace.id()` now.   (The "id" thing is _horribly_ non-pythonic -- and it ends up getting changed out everywhere it occurs.)  (Note also that "id" does not catch various instances of type erasure that I've seen, and a `MaskWorkspace` _is_ also a `MatrixWorkspace`, but usually "just" the latter is not what we want when we want a `MaskWorkspace`.)

I'll go over the code one more time, but before I do that I'll see if I can get it through testing.

walshmm (approved):

PR #528, opened 2025-01-09
Add Grocery keyset for increased safety calling Recipe

rboston628:
## Description of work

Groceries are passed as a grocery, where the key roughly corresponds to a mantid `WorkspaceProperty`, and where the value is the name of a workspace.

The `Recipe` ABC already had a method `mandatoryInputWorkspaces`, which was a set of workspaces which had to be given and had to exist in the ADS.

However, for optional workspaces, it was possible a user could pass an option workspace with an incorrect key, and it would not set that workspace.

This was a problem in Defect 8554, where the `ReductionService` was passing a mask workspace with the key `"combinedPixelMask"`, but `ReductionRecipe` was accessing the mask by the key `"combinedMask"`, meaning the mask was not being set.

This is fixed by also including a method `allGroceryKeys`, which ensures the groceries dictionary **only** has keys from that set.

## Explanation of work

Inside the `Recipe` ABC, there is a new method `allGroceryKeys` which returns the set of all possible keys that might be passed to this recipe.  If the groceries contain a key not in this set, then an error is produced.

Several workspaces inheriting from `Recipe` were refactored to implement this method.  Some tests also had to be changed to accommodate the new behavior with workspace keys.

The `ReductionRecipe` required additional work because of its `_applyRecipe` method.  This created a conflict, as the `self.groceries` dictionary was being passed, which included all of the workspaces needed for reduction, but which are not needed for this the sub-recipes.  This was reworked to not rely on the `self.groceries` dictionary, and instead requires all input workspaces to be used, be passed as keyword arguments.

@ekapadi noticed similar problems and also implemented some changes to this part of `ReductionRecipe`.  Hopefully our changes are compatible.

### Dev testing

Run all three workflows by hand to saving.

### CIS testing

N/A

## Link to EWM item

[EWM#8554](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=8554)

### Verification

### Acceptance Criteria

The simple fix to the defect was implemented in PR #526 .  This PR implements the enhanced validation mentioned in the story.

- [ ] the `Recipe` ABC implements better validation of groceries
- [ ] the `ReductionService` calls the `ReductionRecipe` with the correct key for the combined pixel mask


rboston628:
Merge AFTER PR #526

rboston628:
The recipe `CalculateDiffcalResidualRecipe` did not formerly have a unit test for it.  Creating one is outside the scope of this PR, but I did add some minimal number of tests.

tests/resources/application.yml
```diff
@@ -247,8 +247,8 @@ constants:
     highdSpacingCrop: 0.15
 
   RawVanadiumCorrection:
-    numberOfSlices: 10
-    numberOfAnnuli: 10
+    numberOfSlices: 1
```
walshmm, line 250: Why were these changed?

walshmm:
could you also add these AC and rationale to the story since this is more of a refactor?

tests/resources/application.yml
rboston628: In the tests, it's the difference between running in 1 second vs running in 1 minute

rboston628:
> could you also add these AC and rationale to the story since this is more of a refactor?

Defects can't have AC, but I added something to to description.

walshmm:
performed some basic calibration, normalization, and reduction and didnt run into any regression issues.
Reviewed the code, I approve these changes.

walshmm (approved):

PR #531, opened 2025-01-15
Ewm 5538 normalization regression tests

dlcaballero16:
## Description of work

This adds regression tests specifically for the normalization workflow. This tests a bunch of failure cases and makes sure normalization runs correctly.

## Explanation of work

This test is based on the normalization happy paths test. This test makes sure that intended failures have the correct warning message when failing, and that the final output of normalization is correct. TestSummary was also moved into its own class so that it can be reused within the normalization test.

### Dev testing

Run the test locally, or just make sure the integration test step passes.

### CIS testing

None, this for internal testing.

## Link to EWM item

[EWM#5538](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=5538)

### Verification

### Acceptance Criteria

- [ ] There's a lot of steps, just verify this test is handling every case listed in the EWM description


src/snapred/ui/widget/SmoothingSlider.py
```diff
@@ -79,7 +79,10 @@ def _updateSliderFromNumber(self):
             )
             return
         s = float(text)
-        v = math.log10(s)
+        if s == 0:
+            v = 0
```
rboston628, line 83: Are we sure we don't want `-inf` here?

tests/integration/test_normalization.py
```diff
@@ -0,0 +1,565 @@
...
+        for path_ in metadatas:
+            os.symlink(originalCalibrationHome / path_, tmpCalibrationHome / path_)
+        return tmpCalibrationHome
+
+    yield _calibration_home_from_mirror
+
+    # teardown => __exit__
+    _stack.close()
+    LocalDataService()._indexer.cache_clear()
+
```
rboston628, line 82: Is this fixture different from the implementation in the happy path panel tests?  If not, copy this fixture to `pytest_helpers`, delete the code from from tests, and instead import the fixture from the helper file.

tests/integration/test_normalization.py
```diff
@@ -0,0 +1,565 @@
...
+
+@pytest.mark.datarepo
+@pytest.mark.integration
+class TestGUIPanels:
```
rboston628, line 86: This might need a different name

tests/integration/test_normalization.py
```diff
@@ -0,0 +1,565 @@
...
+
+        if isinstance(self.testSummary, TestSummary):
+            if not self.testSummary.isComplete():
+                self.testSummary.FAILURE()
+            if self.testSummary.isFailure():
+                pytest.fail(f"Test Summary (-vv for full table): {self.testSummary}")
+        # # teardown...
+        # self._logWarningsMessageBox.stop()
+        self._actionPrompt.stop()
+        self.exitStack.close()
```
rboston628, line 121: Same with this fixture.  Actually, maybe the entire class and several of the methods should be moved to a central location and the integration tests inherit from it.

tests/integration/test_normalization.py
```diff
@@ -0,0 +1,565 @@
...
+                assert mp[1].call_count == 1
+
+            # Enter 58810 for run number and 58813 for background and click continue
+            msg = "Please select a sample"
+            mpCrit = MockQMessageBox().critical(msg)
+            with mpCrit[0]:
+                requestView.runNumberField.setText("58810")
+                requestView.backgroundRunNumberField.setText("58813")
+                qtbot.mouseClick(workflowNodeTabs.currentWidget().continueButton, Qt.MouseButton.LeftButton)
+                qtbot.wait(100)
```
rboston628, line 261: These lines can be moved outside of this context manager, to just after L254.  The comment at L263 should be moved to just above L255.

tests/integration/test_normalization.py
```diff
@@ -0,0 +1,565 @@
...
+
+            # Select Vanadium Cylinder as sample and click continue
+            msg = "Please select a grouping file"
+            mpCrit = MockQMessageBox().critical(msg)
+            with mpCrit[0]:
+                requestView.sampleDropdown.setCurrentIndex(3)
+                assert requestView.sampleDropdown.currentIndex() == 3
+                assert requestView.sampleDropdown.currentText().endswith("Silicon_NIST_640D_001.json")
+                qtbot.mouseClick(workflowNodeTabs.currentWidget().continueButton, Qt.MouseButton.LeftButton)
+                qtbot.wait(100)
```
rboston628, line 275: Similarly to above, move this to just under L267, then move comment at L277 to just above L268

tests/integration/test_normalization.py
```diff
@@ -0,0 +1,565 @@
...
+            self.testSummary.SUCCESS()
+
+            # On tweak peaks tab, test each text field(3 of these) by putting non numeric chars and verify error msg
+            tweakPeakView = workflowNodeTabs.currentWidget().view
+            mp[0].stop()
+
+            msg = "Smoothing parameter must be a numerical value"
+            mpWarn = MockQMessageBox().warning(msg)
+            with mpWarn[0]:
+                mpWarn[0].start()
```
rboston628, line 361: This looks like it's double patching

tests/integration/test_normalization.py
```diff
@@ -0,0 +1,565 @@
...
+            # Then set negative values and verify errors on recalculate
+            msg = "Smoothing parameter must be a nonnegative number"
+            mpWarn = MockQMessageBox().warning(msg)
+            with mpWarn[0]:
+                tweakPeakView.fieldXtalDMax.setText("1")
+                tweakPeakView.smoothingSlider.field.setValue("-1")
+                qtbot.wait(100)
+                mpWarn[0].stop()
+                assert len(exceptions) == 0
+                assert mpWarn[1].call_count == 1
```
rboston628, line 416: This looks like it's missing a `mouseClick`

tests/integration/test_normalization.py
```diff
@@ -0,0 +1,565 @@
...
+
+            # Set author to "Test"You must specify the author
+            msg = "You must specify the author"
+            mpCrit = MockQMessageBox().critical(msg)
+            with mpCrit[0]:
+                qtbot.mouseClick(workflowNodeTabs.currentWidget().continueButton, Qt.MouseButton.LeftButton)
+                qtbot.wait(100)
+                assert len(exceptions) == 0
+                assert mpCrit[1].call_count == 1
+                saveView.fieldAuthor.setText("Test")
```
rboston628, line 508: Move this line out of the context manager

tests/integration/test_normalization.py
```diff
@@ -0,0 +1,565 @@
...
+            # Set Comments to "This is a test"You must add comments
+            msg = "You must add comments"
+            mpCrit = MockQMessageBox().critical(msg)
+            with mpCrit[0]:
+                qtbot.mouseClick(workflowNodeTabs.currentWidget().continueButton, Qt.MouseButton.LeftButton)
+                qtbot.wait(100)
+                mpCrit[0].stop()
+                assert len(exceptions) == 0
+                assert mpCrit[1].call_count == 1
+                saveView.fieldComments.setText("This is a test")
```
rboston628, line 519: Move this line out of the context manager

tests/integration/test_normalization.py
dlcaballero16: I'll move the other one, but I'll keep this one here. This one is too modified compared to the happy path version

tests/integration/test_normalization.py
dlcaballero16: Nope, doesn't work. I moved those lines outside like you suggested and the warning box does not get mocked properly

tests/integration/test_normalization.py
dlcaballero16: Also does not work like above

rboston628: That's bizarre, as this is an error handler.  It should only be needed around the steps that create the error...

tests/integration/test_normalization.py
dlcaballero16: Not required here. When calling setValue() for the smoothing slider, it behaves like typing in a number and pressing enter, so no click needed.

src/snapred/ui/widget/SmoothingSlider.py
dlcaballero16: I'll set this to be -10. That way when it sets sliderValue a few lines down, it will be set to the minimum which is -1000. Minimum is defined in the init for this class

rboston628 (approved):
This looks good, and improves the code quality within integration tests

rboston628 (approved):

PR #535, opened 2025-01-23
fix write permission check

walshmm:
## Description of work

The existing perms checks provide a false negative in some cases on the analysis cluster, this new check seems to work.

This solution was tested with Malcolm via a script.

## To test

This may not be easy for a dev to test because the perms issue had to do with not being allowed to write to a specific ipts despite having write perms.  But here is a comparison script.

```

import os
import stat
from pathlib import Path

def _hasWritePermissionstoPathOld(filePath: Path) -> bool:
        # WARNING: `os.access` does not work correctly on the `/SNS` shared filesystem.

        # formerly:
        #   `return os.access(filePath, os.W_OK) if filePath.exists() else False`

        # alternative implementation (LINUX specific):
        hasPermissions = False
        if filePath.exists():
            stat_ = os.stat(filePath)

            # Get the path's permissions mode bits.
            mode = stat.S_IMODE(stat_.st_mode)

            # Get the path owner's user-id, and group-id.
            fuid = stat_.st_uid
            fgid = stat_.st_gid

            # Get the current user's user-id and group-ids
            uid = os.getuid()
            gids = os.getgroups()

            # Check for any overlap with the write permission's mode bits:
            #   checking the user-bits, if the current user is the owner;
            #   then checking the group-bits, if the user belongs to the file-owner's group;
            #   and finally checking the other-bits.
            hasPermissions = (
                (uid == fuid) and bool(mode & 0o200) or (fgid in gids) and bool(mode & 0o020) or bool(mode & 0o002)
            )
        return hasPermissions
     
  
def _hasWritePermissionstoPathNew(filePath: Path) -> bool :
    import tempfile
    if not filePath.is_dir():
        filePath = filePath.parent()
    tempfile.tempdir = str(filePath)
    try:
        with tempfile.TemporaryFile() as fp:
            fp.write(b'Hello world!')
    except Exception as e:
        return False
    return True
  
assert _hasWritePermissionstoPathOld(Path("/SNS/SNAP/IPTS-28343/")) is True
assert _hasWritePermissionstoPathOld(Path("/SNS/users/wqp/SNAP/IPTS-28343/shared/")) is False

assert _hasWritePermissionstoPathNew(Path("/SNS/SNAP/IPTS-28343/")) is True
assert _hasWritePermissionstoPathNew(Path("/SNS/users/wqp/SNAP/IPTS-28343/shared/")) is False
```
NOTE: you will need to change the path in the asserts to not point to my home.

## Link to EWM item

[EWM#8964](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=8964)

### Verification

### Acceptance Criteria

- [ ] reduction should allow users who have write permissions to write the output of reduction.

### DIFF
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -3,6 +3,7 @@ import glob
 import json
 import os
 import re
+import tempfile
 import time
 from errno import ENOENT as NOT_FOUND
 from functools import lru_cache
@@ -175,10 +176,8 @@ class LocalDataService:
 
     @staticmethod
     def _hasWritePermissionstoPath(filePath: Path) -> bool:
-        import tempfile
-
         if not filePath.is_dir():
-            filePath = filePath.parent()
+            filePath = filePath.parent
         tempfile.tempdir = str(filePath)
         try:
             with tempfile.TemporaryFile() as fp:
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -3,7 +3,6 @@ import glob
 import json
 import os
 import re
-import stat
 import time
 from errno import ENOENT as NOT_FOUND
 from functools import lru_cache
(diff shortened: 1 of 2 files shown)


src/snapred/backend/data/LocalDataService.py
```diff
@@ -176,35 +176,18 @@ def checkFileandPermission(filePath: Path) -> Tuple[bool, bool]:
...
+        if not filePath.is_dir():
+            filePath = filePath.parent
+        result = True
+        preTempDir = tempfile.tempdir
+        tempfile.tempdir = str(filePath)
+        try:
+            with tempfile.TemporaryFile() as fp:
+                fp.write(b"Hello world!")
+        except Exception:  # noqa: BLE001
+            result = False
```
rboston628, line 188: I'm worried in the case where open works but write doesn't, that the file won't be deleted

walshmm: propagating this script from slack
```
import tempfile
import os
tempfile.tempdir = "."
tempFileLocation = None
try:
    with tempfile.TemporaryFile() as f:
        f.write(b"Hello, world!")
        tempFileLocation = f.name
        assert os.path.exists(f.name), f"File {tempFileLocation} does not exist!"
        raise Exception("Arbitrary write error!")
except Exception as e:
    print(f"{e}")
    
print(f"Temp file location: {tempFileLocation}")

# is it still there?
assert not os.path.exists(tempFileLocation), f"File {tempFileLocation} still exists!"
```
I would need a replicator.

rboston628 (approved):
Code looks good, and script demonstrates continuity of behavior

PR #539, opened 2025-01-29
Transfer PV-logs during load of grouping and mask workspaces

ekapadi:
## Description of work

This is a supporting PR for the reduction live-data work.  The changes here allow all loaded workspaces, with the exception of the input data workspace itself, to be reused from reduction cycle to reduction cycle.

During development work, it was found that although Mantid's `LoadDiffCal` allows an instrument to be specified, it does not additionally transfer the relevant instrument PV-logs from the donor workspace.  The result of this is that any mask or grouping workspace loaded by SNAPRed would not have a defined stateID-SHA.  During the reduction process, pixel-mask compatibility is checked using the combination of the stateID-SHA, and the lite-mode state -- ADS-resident pixel-mask workspaces with no stateID-SHA will be skipped as incompatible.

Also found during development work, when the detector positions are compared between workspaces written using `SaveNexus` and read back using `LoadNexus`, they fail to match to the Mantid position and rotation target accuracy.  For positions, this accuracy is `1.0e-9`m, and for rotations, it is `1.0e-9`m at `1000.0`m, which corresponds to `1.0e-12` radians.  The reason for this readback mismatch has to do with the lack of a precision specification when writing the locations to the _legacy_ 'instrument_parameter_map' (in the NeXus file).  Unfortunately, this same legacy system is used to initialize the instrument parameters during reload.  This write-precision defect has been separately fixed in another Mantid PR. This current PR adds a temporary work around for this issue.

## Explanation of work

  * `LoadCalibrationWorkspaces` and `LoadGroupingDefinition` algorithms, both of which use Mantid's `LoadDiffCal` algorithm, now transfer the relevent instrument PV-logs to their output workspaces;

  * Any `FetchGroceriesAlgorithm` loader which goes through the `LoadNexus` pathway, now additionally calls `populateInstrumentParameters` for the output workspace.  This bypasses the inaccuracy of the legacy 'instrument_parameter_map', and directly utilizes the PV-log values, which are stored in their binary representation as double-precision values;

  * General utilities `transferInstrumentPVLogs` and `populateInstrumentParameters` are provided in `snapred.backend.data.util.PV_logs_util`.  The `populateInstrumentParameters` utility is a temporary work around, as `ExperimentInfo.populateInstrumentParameters` has now been exposed to the Python API in another Mantid PR.

## To test

Additional unit tests have been added to test the new `PV_logs_util`.
Current unit tests have also been adjusted to cover this work.  Mostly these adjustments involved adding a more-complete instrument to initialization to the workspaces used in the `GroceryService` tests.

### CIS testing

This is a SNAPRed _internal_ item.  It can be evaluated as part of  the reduction live-data work.

## Link to EWM item
Note that there is NO EWM item for this PR specifically.  These defects were fixed in support of the reduction live-data work.

[EWM#7437](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=7437)

### DIFF
--- a/src/snapred/backend/data/util/PV_logs_util.py
+++ b/src/snapred/backend/data/util/PV_logs_util.py
@@ -1,0 +1,42 @@
+"""
+Python `Mapping`-interface adapters and utility-methods relating to Mantid workspace logs (i.e. `Run`)
+  and process-variable(PV) logs.
+"""
+# Note: in the upcoming reduction live-data PR this file includes several additional `Mapping` adapters.
+
+from collections.abc import Iterable
+
+from mantid.api import Run
+from mantid.simpleapi import (
+    AddSampleLog,
+    mtd
+)
+
+from snapred.meta.Config import Config
+
+def transferInstrumentPVLogs(dest: Run, src: Run, keys: Iterable):
+    # Transfer instrument-specific PV-log values, between the `Run` attributes
+    #   of source and destination workspaces.
+    
+    # Placed here for use by various `FetchGroceriesAlgorithm` loaders.
+    for key in keys:
+        if src.hasProperty(key):
+            dest.addProperty(key, src.getProperty(key), True)
+    # REMINDER: the instrument-parameter update still needs to be explicitly triggered!
+
+def populateInstrumentParameters(wsName: str):
+    # This utility function is a "stand in" until Mantid PR #38684 can be merged.
+    # (see https://github.com/mantidproject/mantid/pull/38684)
+    # After that, `mtd[wsName].populateInstrumentParameters()` should be used instead.
+    
+    # Any PV-log key will do, so long as it is one that always exists in the logs.
+    pvLogKey = "run_title"
+    pvLogValue = mtd[wsName].run().getProperty(pvLogKey).value
+    
+    AddSampleLog(
+        Workspace=wsName,
+        LogName=pvLogKey,
+        logText=pvLogValue,
+        logType="String",
+        UpdateInstrumentParameters=True,
+    )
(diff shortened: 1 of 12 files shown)


src/snapred/backend/data/util/PV_logs_util.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+# Note: in the upcoming reduction live-data PR this file includes several additional `Mapping` adapters.
+
+from collections.abc import Iterable
+
+from mantid.api import Run
+from mantid.simpleapi import AddSampleLog, mtd
+
+
+def transferInstrumentPVLogs(dest: Run, src: Run, keys: Iterable):
```
rboston628, line 13: Should this require the `Run` objects, or would it be more idiomatic to pass the workspace string names?

rboston628: If `Run`, could some kind of parameter validation be added?

tests/unit/backend/data/util/test_PV_logs_util.py
```diff
@@ -0,0 +1,151 @@
...
+            BinWidth=0.001,
+            XUnit="dSpacing",
+            NumBanks=4,  # must produce same number of pixels as fake instrument
+            BankPixelWidth=2,  # each bank has 4 pixels, 4 banks, 16 total
+        )
+        LoadInstrument(
+            Workspace=wsName,
+            Filename=Resource.getPath("inputs/testInstrument/fakeSNAP_Definition.xml"),
+            RewriteSpectraMap=True,
+        )
```
rboston628, line 39: Isn't there a helper for this?

ekapadi:
This PR got "subsumed" into the "live-data stability PR", which has since been merged -- mostly due to the huge delay on that entire live-data EPIC.

**Thanks very much for your review comments!**

PR #541, opened 2025-02-03
reduction: live-data: complete

ekapadi:
## Description of work

This PR includes the complete SNAPRed reduction live-data implementation.  With respect to complexity, the required back-end changes probably came in within the original scope of this enabler.  In contrast, the required front-end changes, including the implementation of the non-linear workflow, were significantly more complicated than expected.

## Explanation of work

**LIVE-DATA back end:**

  * LIVE-DATA FALLBACK MODE: modifications to `GroceryService` `fetchNeutronDataSingleUse` and `fetchNeutronDataCached` methods so that loading from the live-data listener is used as a fallback.  This means that in addition to operating explicitly in live-data mode (described next), if the live-data run number is known, ALL WORKFLOWS can function with live data, simply by specifying this run number.

  * EXPLICIT LIVE-DATA MODE: this mode is triggered by using the new `liveData` `GroceryListBuilder` token.  This differs from the fallback mode, in that the data _duration_ can be specified.  Given an active run, either a subsection of its data can be loaded, specified using a time-duration prior to the present time, or all of the available data can be loaded.

  * A new `LiveMetadata` DAO, which encapsulates attributes of `mantid.api.Run` that are relevant to accessing the state of the live-data listener.  This object allows the convenient identification of significant live-data states, such as \<no run active\> and \<no beam active\>.

  * New `LocalDataService` live-data utility methods: `hasLiveDataConnection` and `getLiveMetadata`.  Note that due to limitations in the DAS / listener implementation, there is some minor code duplication here with the `FetchGroceriesAlgorithm` use of the `LoadLiveData` algorithm.  The primary objective of these `LocalDataService` methods is to detect whether a listener connection is available, and then if so, to load the _minimum_ possible in order to access the PV-logs associated with the live-data run.  (DAS-group itself needs an "enabler" to simply allow access to the PV-logs from the listener directly, but this issue has not been addressed yet.)

  * A new `LiveDataState` exception, which is another flow-control exception allowing changes in the live-data state to be detected and indicated.  For example, when a live-data run ends, the run number available from the listener changes to _zero_.  Or, when the beam goes down, the \<integrated proton charge\> goes to _zero_.

  * COARSE-GRAINED USER-CANCELLATION REQUEST: this uses a new `UserCancellation` flow-control exception in the `SNAPResponseHandler`, with an associated `ResponseCode` value.  This allows any workflow to be cancelled at the next "service point".  This is the only form of supported exit from the live-data reduction cycle.  Note that an implication of this is that now ALL WORKFLOWS support this type of user-cancellation request, and in general, the \<cancel button\> should always be enabled on the panel.

  * Modifications to `ReductionRecipe` so that all output workspaces, including the unfocused data, are updated "at once" at the appropriate processing stage.  This allows any open plots of these workspaces to display correctly, and not to show intermediate-calculation values.  All of this is important to the live-data standard usage scenario.

  * Any workspaces that are loaded in support of processing a `ReductionRequest` are retained between live-data cycles.  The input-data workspace itself is replaced at the start of each cycle.

  * Add calibration masks to `fetchCalibrations` workspaces list.  At present, the only use of this list is to add these workspaces to the retained-workspaces set. Both calibration table and mask workspaces should be retained across live-data cycles.

  * Move "lite_grouping_map" name to the `WorkspaceNameGenerator`.  This allows access to this name by the `ReductionWorkflow`, so that "lite_grouping_map" can be added to the workspaces retained across live-data cycles.

  * `ReductionWorkflow` now maintains a current `ReductionStatus` state set by using its `setStatus` method.  Retaining this state allows the live-data summary panel to show what part of the reduction process is presently active.

**LIVE-DATA front end:**

  * LIVE-DATA REDUCTION CYCLE: `ReductionWorkflow` now supports this new non-linear workflow.  During the first workflow pass, everything behaves _exactly_ like the current SNAPRed workflow, including the use of any required continue-anyway dialogs.  During future live-data cycles, the as specified `ReductionRequest` values are re-used.  Live-data cycling is canceled using the new `UserCancellation` system.  Once canceled, the entire sequence can be started again, possibly including the specification of new values.

  * LIVE-DATA STATE CHANGE: in addition to any `UserCancellation` exit, the live-data cycle will also be exited when a live-data state change is detected.  Normally, this would be the end of a live-data run, or rarely, the direct transition to another live-data run.

  * LIVE-DATA REQUEST PANEL: this panel shares most of its widget controls with the `ReductionRequestView`, which it is a variant of.  The key difference is that this panel supports an _active_ live-data run summary, using a combination of a steady state, or flashing `LEDIndicator` and a summary text display.  It also supports two slider controls, one for the data interval -- as a duration prior to the current time -- and one for the desired update interval.  Note that the normal `ReductionRequestView` variant replaces these controls with its run-number selection section.

  * TIME-OF-DAY UPDATE: this update occurs every second and is controlled by a separate `QTimer`.

  * METADATA UPDATE: this is a per-update interval update of the live-data panel's metadata, and is also controlled by a separate `QTimer`.  This update occurs _only_ when live-data reduction is not being processed, which will either be before the start of a live-data cycle, or between reduction-processing stages of an in-progress cycle.

  * LIVE-DATA REDUCTION, RESUBMISSION TIMING: this schedules the start of the next live-data cycle.  This uses the measured wallclock time of the previous reduction, in combination with the requested update interval.  This results in either a timed delay prior to the next submission, or possibly an immediate submission in case the wallclock time has exceeded the desired update interval.

**MISCELLANEOUS fixes:**

  * Allow SNAPRed to have its own `style.qss` or `workspace_style.qss` depending on how the application is loaded.  This then allows the centralization of SNAPRed's widget style properties.  (Previously, it did not have a separate style sheet at all when loaded under the workbench.)

  * Fix mix-up of static and non-static methods in `SNAPResponseHandler`.

  * Fix dialog modality, and parenting of the init-state message box, and its dialog.

  * Pass all former `groceries` arguments in the `ReductionRecipe` as explicit keyword arguments.  In general, the use of `groceries` in this recipe was completely abused.  Further, `ReductionRecipe._applyRecipe` is no longer called on any sub-recipe that doesn't actually need to be executed.  Removed redundant checking of keyword argument presence, as this is now handled by the `Recipe` base class itself.

  * An implicit (i.e. not specified in the naming template) 'hidden' token is now available for all `WorkspaceNameGenerator` workspace types.  When this token is set, the generated workspace name is prefixed with the Mantid \<hidden workspace\> double underscore.

  * Fix `StateValidationException` mixed-up code.  There is no reason at all that any end user would care whether or not they have permissions to write to the directory of the executing python file?!

  * `InitializeStateCheckView` dialog that potentially closes during its own initialization, now sends a queued signal (using `QMetaObject`), rather than calling `close` directly.  This fixes one type of GUI freeze behavior that has been observed.

  * \<cancel button\> "are you sure" dialog now displays "yes" "no", rather than "continue" "cancel" -- the previous behavior was really confusing!

  * \<cancel button\> is now active in all panels.  This is connected to the new `UserCancellation` flow-control exception.

  * Add complete instrument-parameter information to grouping and mask workspaces loaded by the `GroceryService`.  In addition work around the parameter-loading precision defect (a separate PR) in Mantid `SaveNexus` and `LoadNexus` by applying `populateInstrumentParameters` after all loading.  In combination, this allows grouping and mask workspaces to present the correct stateID-SHA, and also for all workspaces to meet Mantid's target accuracy for detector positions and rotations.

  * A new `PV_logs_util` module to allow presentation of `mantid.api.Run` and the NeXus logs as Python Mapping objects.  This centralizes the treatment of the DAS-log keys, and the initialization of the `DetectorState` objects, which are used during the computation of the stateID-SHA.  In addition, this module provides utility functions for transferring PV-logs between `mantid.api.Run` objects, and for triggering `ExperimentInfo.populateInstrumentParameters` (,prior to the merge of the Mantid PR that will expose that method to the Python api).

  * Add instrument-parameter PV-logs keys to `application.yml` -- this includes the alternative keys as well.

  * Return `(stateID-SHA, DetectorState)` from `LocalDataService.generateStateID`, as these are the values that are actually required for use.  This reduces the PV-logs parsing by a factor of two.

### Dev testing

I'll fill this in, as I go through each of these sections myself.

**Unit tests:**

  * I'm working to improve the code coverage right now.

**Mock live-data listener:**

It's important to start with these tests, so that one can get an expectation of how the system is _supposed_ to work.  Take a look at the new changes in the `liveData` section of `application.yml`.  Uncomment the appropriate values, as indicated.

  * Despite its failings, run "46680" is used as an example, although you can use any run that you'd prefer.  From both the live-data, and the non-live-data panels: all workflow variants, including with and without calibration and  normalization, should be examined. 
     For the live-data workflow: as mentioned above, the first-pass through the workflow sets the specifics used for future passes, which then continue automatically until either user-cancellation is requested, or the live-data state changes.  Mantid has memory-fragmentation issues.  This means that how many reduction cycles you can get through before you run out of memory really depends on the node you're running on.  I've managed to cycle continually through greater than thirty cycles (> 1/2 hour of total time for "46680").
  

**Tests with the real listener:**
For these tests, the `application.yml` `liveData` section should be reverted to its original state.

  * What behavior is observed here depends on the current active-run and beam state.  Hopefully I've dealt with all of the possible aspects of this state, but if not, any required additions will be spun off as new defects.  It will be useful here if there exists a calibration for the current run.  If there does, it will be located in the normal way.
  
**I'm expecting that we _will_ find defects, as these implementation changes are somewhat complicated.**  For any of these that don't have obvious quick fixes, we should spin off additional EWM items.
  

## Link to EWM item

[EWM#755](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=755)

### Verification

### Acceptance Criteria

- [ ] acceptance criterion 1
- [ ] acceptance criterion 2

### DIFF
--- a/src/snapred/backend/api/InterfaceController.py
+++ b/src/snapred/backend/api/InterfaceController.py
@@ -4,6 +4,7 @@ from snapred.backend.api.RequestScheduler import RequestScheduler
 from snapred.backend.dao import SNAPRequest, SNAPResponse
 from snapred.backend.dao.SNAPResponse import ResponseCode
 from snapred.backend.error.ContinueWarning import ContinueWarning
+from snapred.backend.error.LiveDataState import LiveDataState
 from snapred.backend.error.RecoverableException import RecoverableException
 from snapred.backend.log.logger import snapredLogger
 from snapred.backend.service.ServiceFactory import ServiceFactory
@@ -31,7 +32,7 @@ class InterfaceController:
         return "\n".join(snapredLogger.getWarnings())
 
     def executeRequest(self, request: SNAPRequest) -> SNAPResponse:
-        # execute the recipe
+        # execute the request
         # return the result
         try:
             self.logger.debug(f"Request Received: {request.json()}")
@@ -48,6 +49,12 @@ class InterfaceController:
         except RecoverableException as e:
             self.logger.error(f"Recoverable error occurred: {str(e)}")
             response = SNAPResponse(code=ResponseCode.RECOVERABLE, message=e.model.json())
+        except LiveDataState as e:
+            self.logger.error(
+                f"live-data state change: {e.model.transition}: "
+                + f"{e.model.endRunNumber} <- {e.model.startRunNumber}"
+            )
+            response = SNAPResponse(code=ResponseCode.LIVE_DATA_STATE, message=e.model.json())
         except ContinueWarning as e:
             self.logger.error(f"Continue warning occurred: {str(e)}")
             response = SNAPResponse(code=ResponseCode.CONTINUE_WARNING, message=e.model.json())
@@ -63,7 +70,7 @@ class InterfaceController:
         return response
 
     def executeBatchRequests(self, requests: List[SNAPRequest]) -> List[SNAPResponse]:
-        # verify all requests have same path
+        # verify that all requests have the same path
         for request in requests:
             if not requests[0].path == request.path:
                 self.logger.error("Mismatch of paths in list of requests")
--- a/src/snapred/backend/dao/ingredients/GroceryListItem.py
+++ b/src/snapred/backend/dao/ingredients/GroceryListItem.py
@@ -1,4 +1,5 @@
-from typing import Any, ClassVar, Literal, Optional, get_args
+import datetime
+from typing import Any, ClassVar, Literal, NamedTuple, Optional, get_args
 
 from pydantic import BaseModel, model_validator
 
@@ -11,6 +12,12 @@ logger = snapredLogger.getLogger(__name__)
 KnownUnits = Literal[wng.Units.TOF, wng.Units.DSP, wng.Units.DIAG]
 known_units = list(get_args(KnownUnits))
 
+LiveDataArgs = NamedTuple(
+    "LiveDataArgs",
+    [
+        ("duration", datetime.timedelta),
+    ],
+)
 
 GroceryTypes = Literal[
     "neutron",
@@ -33,10 +40,12 @@ class GroceryListItem(BaseModel):
     """
 
     # Reserved instrument-cache run-number values:
-    RESERVED_NATIVE_RUNNUMBER: ClassVar[str] = ReservedRunNumber.NATIVE  # unmodified _native_ instrument:
-    #   from 'SNAP_Definition.xml'
-    RESERVED_LITE_RUNNUMBER: ClassVar[str] = ReservedRunNumber.LITE  # unmodified _lite_ instrument  :
-    #   from 'SNAPLite.xml'
+
+    # unmodified _native_ instrument: from 'SNAP_Definition.xml'
+    RESERVED_NATIVE_RUNNUMBER: ClassVar[str] = ReservedRunNumber.NATIVE
+
+    # unmodified _lite_ instrument  : from 'SNAPLite.xml'
+    RESERVED_LITE_RUNNUMBER: ClassVar[str] = ReservedRunNumber.LITE
 
     workspaceType: GroceryTypes
     useLiteMode: bool  # indicates if data should be reduced to lite mode
@@ -44,7 +53,13 @@ class GroceryListItem(BaseModel):
     # optional loader:
     # -- "" tells FetchGroceries to choose the loader
     loader: Literal[
-        "", "LoadGroupingDefinition", "LoadCalibrationWorkspaces", "LoadNexus", "LoadEventNexus", "LoadNexusProcessed"
+        "",
+        "LoadCalibrationWorkspaces",
+        "LoadEventNexus",
+        "LoadGroupingDefinition",
+        "LoadLiveData",
+        "LoadNexus",
+        "LoadNexusProcessed",
     ] = ""
 
     # optional: Mantid-workspace number tag (only output if != 1)
@@ -67,6 +82,10 @@ class GroceryListItem(BaseModel):
     #   the automatic instrument-donor caching system.
     instrumentPropertySource: Optional[Literal["InstrumentName", "InstrumentFilename", "InstrumentDonor"]] = None
     instrumentSource: Optional[str] = None
+
+    # Set arguments for an explicit live-data request;
+    #   note that live-data is also the last fallback for the `fetchNeutronData` methods.
+    liveDataArgs: Optional[LiveDataArgs] = None
 
     # if set to False, neutron data will not be loaded in a clean, cached way
     # this is faster and uses less memory, if you know you only need one copy
@@ -103,8 +122,15 @@ class GroceryListItem(BaseModel):
                 case "neutron":
                     if v.get("runNumber") is None:
                         raise ValueError("Loading neutron data requires a run number")
+                    if v.get("liveDataArgs") is not None:
+                        duration = v.get("liveDataArgs").duration
+                        if duration < datetime.timedelta(seconds=0):
+                            raise ValueError(
+                                "live-data args 'duration' must be >= 0 seconds; "
+                                + "zero indicates to load all available data"
+                            )
                     if v.get("groupingScheme") is not None:
-                        del v["groupingScheme"]
+                        raise ValueError("Loading neutron data should not specify a grouping scheme")
                     if v.get("instrumentPropertySource") is not None:
                         raise ValueError("Loading neutron data should not specify an instrument")
                 case "grouping":
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -1,7 +1,7 @@
 # ruff: noqa: F811
 import json
-import os
 from collections.abc import Iterable
+from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
 
@@ -19,6 +19,7 @@ from snapred.backend.dao.ingredients import GroceryListItem
 from snapred.backend.dao.state import DetectorState
 from snapred.backend.dao.WorkspaceMetadata import UNSET, WorkspaceMetadata
 from snapred.backend.data.LocalDataService import LocalDataService
+from snapred.backend.error.LiveDataState import LiveDataState
 from snapred.backend.log.logger import snapredLogger
 from snapred.backend.recipe.algorithm.MantidSnapper import MantidSnapper
 from snapred.backend.recipe.FetchGroceriesRecipe import FetchGroceriesRecipe
@@ -65,6 +66,9 @@ class GroceryService:
         #        0: if it is loaded, but no copies have been made;
         #       >0: if any copies have been made, since loading
         self._loadedRuns: Dict[Tuple[str, bool], int] = {}
+
+        # _liveDataKeys: cache keys into `_loadedRuns` for any live data that has been loaded
+        self._liveDataKeys: List[Tuple[str, bool]] = []
 
         # Cache maps to workspace names, for various purposes
         self._loadedGroupings: Dict[Tuple[str, str, bool], str] = {}
@@ -199,11 +203,15 @@ class GroceryService:
         ipts = self.dataService.getIPTS(runNumber, instrumentName)
         return str(ipts)
 
-    def _createNeutronFilename(self, runNumber: str, useLiteMode: bool) -> str:
+    def _createNeutronFilePath(self, runNumber: str, useLiteMode: bool) -> Path:
+        # TODO: fully normalize this to pathlib.Path:
+        #   -- problems (among others): `GetIPTS` returns with an '/' at the end?
+
+        IPTS = self.getIPTS(runNumber)
         instr = "nexus.lite" if useLiteMode else "nexus.native"
         pre = instr + ".prefix"
         ext = instr + ".extension"
-        return self.getIPTS(runNumber) + Config[pre] + str(runNumber) + Config[ext]
+        return Path(IPTS + Config[pre] + str(runNumber) + Config[ext])
 
     @validate_call
     def _createGroupingFilename(self, runNumber: str, groupingScheme: str, useLiteMode: bool) -> str:
@@ -291,7 +299,7 @@ class GroceryService:
     def _createGroupingWorkspaceName(self, groupingScheme: str, runNumber: str, useLiteMode: bool) -> WorkspaceName:
         # TODO: use WNG here!
         if groupingScheme == "Lite":
-            return "lite_grouping_map"
+            return wng.liteDataMap().build()
         instr = "lite" if useLiteMode else "native"
         return f"{Config['grouping.workspacename.' + instr]}_{groupingScheme}_{runNumber}"
 
@@ -455,7 +463,7 @@ class GroceryService:
         :param copy: the name of the resulting cloned workspace
         :type copy: WorkspaceName
         :return: a pointer to the cloned workspace in the ADS
-        :rtype: a C++ shared pointer to a MatrixWorkspace
+        :rtype: a python wrapped, C++ shared pointer to a Workspace
         """
         from mantid.simpleapi import CloneWorkspace
 
@@ -546,7 +554,7 @@ class GroceryService:
         The grouping workspaces require an instrument definition, and do not always have their own instrument definition
         saved with them (such as XML groupings).  Therefore, when loading groupings, it is necessary to match them to
         the proper instrument definition *with the proper instrument state params*.
-        This uses the run number (and lite mode)to locate the proper state, and from that the proper instrument
+        This uses the run number (and lite mode) to locate the proper state, and from that the proper instrument
         definition with instrument params for that state.
 
         :param runNumber: a run number that was taken in the desired instrument state
@@ -747,18 +755,14 @@ class GroceryService:
                 raise RuntimeError(f"unable to load workspace {name} from {filePath}")
         return data
 
-    def fetchNeutronDataSingleUse(self, runNumber: str, useLiteMode: bool, loader: str = "") -> Dict[str, Any]:
+    def fetchNeutronDataSingleUse(self, item: GroceryListItem) -> Dict[str, Any]:
         """
         Fetch a neutron data file, without copy-protection.
         If the workspace is truly only needed once, this saves time and memory.
-        If the workspace needs to be loaded more than once, used cached method.
+        If the workspace needs to be loaded more than once, use the cached method.
 
-        :param runNumber: the neutron data run number
-        :type runNumber: str
-        :param useLiteMode: whether to use lite or native resolution
-        :type useLiteMode: bool
-        :param loader: the loader algorithm to use to load the data, optional
-        :type loader: str
+        :param item: the grocery-list item
+        :type item: GroceryListItem
         :return: a dictionary with the following keys
 
             - "result": true if everything ran correctly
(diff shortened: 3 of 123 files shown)


src/snapred/backend/api/InterfaceController.py
```diff
@@ -31,9 +35,16 @@ def getWarnings(self):
         return "\n".join(snapredLogger.getWarnings())
 
     def executeRequest(self, request: SNAPRequest) -> SNAPResponse:
-        # execute the recipe
+        # execute the request
         # return the result
         try:
+            # Coarse-granularity user-cancellation request:
+            #   this supports a possible cancellation, but only between each service call.
+            if QThread.currentThread().isInterruptionRequested():
```
walshmm, line 43: Could this go in [WorkflowImplementer.request](https://github.com/neutrons/SNAPRed/blob/9eb99a87d1fd77e511171e2feb4e2f810047f11a/src/snapred/ui/workflow/WorkflowImplementer.py#L130-L135)?  that should be the only way workflows interact with the api, and would keep pyqt on the frontend side of things.

src/snapred/backend/dao/ingredients/GroceryListItem.py
```diff
@@ -33,9 +40,12 @@ class GroceryListItem(BaseModel):
     """
 
     # Reserved instrument-cache run-number values:
-    RESERVED_NATIVE_RUNNUMBER: ClassVar[str] = ReservedRunNumber.NATIVE  # unmodified _native_ instrument:
+    RESERVED_NATIVE_RUNNUMBER: ClassVar[str] = ReservedRunNumber.NATIVE
+    # unmodified _native_ instrument:
```
walshmm, line 44: I think this got formatted and should be moved above line 43, the comment on 48 being moved to 46 as well

src/snapred/backend/data/GroceryService.py
```diff
@@ -768,100 +772,264 @@ def fetchNeutronDataSingleUse(self, runNumber: str, useLiteMode: bool, loader: s
...
+            success = True
+
+        if not success and not liveDataMode:
+            liteModeFilePath: Path = self._createNeutronFilePath(runNumber, True)
+            nativeModeFilePath: Path = self._createNeutronFilePath(runNumber, False)
+            cachedNativeMode = self._loadedRuns.get(self._key(runNumber, False)) is not None
+
+            match (useLiteMode, cachedNativeMode, liteModeFilePath.exists(), nativeModeFilePath.exists()):
+                # THERE ARE MANY SPECIAL CASES HERE, but basically a single-use fetch will use
+                #   cached (or on-disk lite versions) if available, but will not create any new cache entries if not.
```
walshmm, line 805: I had actually taken a look at this as part of a noncritical issue that had gotten shelved and did some work to break it up into its component parts in [this branch here](https://github.com/neutrons/SNAPRed/blob/ewm7732_normalize_by_direct_beam_monitor/src/snapred/backend/data/GroceryService.py#L750-L791), I think it simplifies the logic significantly.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -82,37 +89,37 @@ def _createFileNotFoundError(msg, filename):
...
+    def getInstrumentConfig(self):
+        return self._instrumentConfig
+
     def fileExists(self, path):
         return os.path.isfile(path)
 
     def _determineInstrConfigPaths(self) -> None:
         """This method locates the instrument configuration path and
-        sets the instance variable ``instrumentConfigPath``."""
+        sets the instance variable ``_instrumentConfigPath``."""
```
walshmm, line 111: too bad this stuff is going away in the historical SNAPInstPrm pr I'm working on!

src/snapred/backend/data/LocalDataService.py
```diff
@@ -1045,22 +1045,23 @@ def _prepareStateRoot(self, stateId: str):
         self._writeGroupingMap(stateId, groupingMap)
 
     def checkCalibrationFileExists(self, runId: str):
+        # TODO: run number format validation does not belong here!
```
walshmm, line 1048: :+1:

src/snapred/backend/data/LocalDataService.py
```diff
@@ -1258,3 +1259,97 @@ def writePixelMask(
...
+
+        facility, instrument = Config["liveData.facility.name"], Config["liveData.instrument.name"]
+        hostname = urlparse(
+            "http://" + ConfigService.getFacility(facility).instrument(instrument).liveDataAddress()
+        ).hostname
+        status = True
+        try:
+            socket.gethostbyaddr(hostname)
+        except Exception:  # noqa: BLE001
+            # specifically:
```
walshmm, line 1283: print stacktrace?  I dont know if this swallows the error message.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -1258,3 +1259,97 @@ def writePixelMask(
...
+        """Construct a 'LiveMetadata' instance from a 'mantid.api.Run' instance."""
+
+        logs = mappingFromRun(run)
+        metadata = None
+        try:
+            run_number: str = str(logs["run_number"])
+
+            # See comment at `snapred.backend.data.util.PV_logs_util.mappingFromRun` about this conversion.
+            start_time: datetime.datetime = datetime.datetime.replace(
+                np.datetime64(logs["start_time"], "us").astype(datetime.datetime), tzinfo=datetime.timezone.utc
```
walshmm, line 1298: could this logic be housed in a method with a helpful name?

src/snapred/backend/data/LocalDataService.py
```diff
@@ -1258,3 +1259,97 @@ def writePixelMask(
...
+            )
+        except (KeyError, RuntimeError, ValidationError) as e:
+            raise RuntimeError("unable to extract LiveMetadata from Run") from e
+        return metadata
+
+    def _readLiveData(self, ws: WorkspaceName, duration: int):
+        # 'StartTime=""' => read all of the available data
+
+        startTime = (
+            (datetime.datetime.utcnow() + datetime.timedelta(seconds=-duration)).isoformat() if duration != 0 else ""
```
walshmm, line 1326: Comment to clarify this is `duration` seconds in the passed, just so its easier to glance at.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -1258,3 +1259,97 @@ def writePixelMask(
...
+        )
+
+        # TODO: this call is partially duplicated at `FetchGroceriesAlgorithm`.
+        #   However, this separate method is required in order to specify a "fast load" for metadata purposes.
+        self.mantidSnapper.LoadLiveData(
```
walshmm, line 1331: if this is only used for the load of metadata, should we narrow the scope of arguments and change the method name as to not tempt misuse? I.e. `duration` is defined locally and the method is called `loadLiveRunForMetadata` or something of that nature?

src/snapred/backend/recipe/ReductionRecipe.py
```diff
@@ -272,8 +314,16 @@ def execute(self):
                 normalizationClone = self._getNormalizationWorkspaceName(groupingIndex)
 
             # 3. GenerateFocussedVanadiumRecipe
+
+            ##
+            ## TODO: =====> THIS NEXT SUB-RECIPE modifies its 'inputWorkspace': that behavior is not really OK! <======
```
walshmm, line 319: Will this require another ticket?

src/snapred/backend/recipe/algorithm/FetchGroceriesAlgorithm.py
```diff
@@ -88,65 +91,100 @@ def validateInputs(self) -> Dict[str, str]:
...
-                getattr(self.mantidSnapper, loaderType)(
-                    f"Loading data using {loaderType}",
-                    Filename=filename,
-                    OutputWorkspace=outWS,
-                )
+
+        # Allow live-data mode to replace an existing workspace
+        addOrReplace = loaderType == "LoadLiveData"
+
+        # TODO: the `if .. doesExist` should be centralized to cache treatment
```
walshmm, line 126: It probably also masks misuse via unnecessary load calls.  If it exists how did we get here?

src/snapred/backend/service/ReductionService.py
```diff
@@ -405,7 +414,12 @@ def fetchReductionGroceries(self, request: ReductionRequest) -> Dict[str, Any]:
             combinedPixelMask = None
 
         # gather the input workspace and the diffcal table
-        self.groceryClerk.name("inputWorkspace").neutron(request.runNumber).useLiteMode(request.useLiteMode).add()
+        if not request.liveDataMode:
+            self.groceryClerk.name("inputWorkspace").neutron(request.runNumber).useLiteMode(request.useLiteMode).add()
+        else:
+            self.groceryClerk.name("inputWorkspace").neutron(request.runNumber).useLiteMode(
+                request.useLiteMode
+            ).liveData(duration=request.liveDataDuration).add()
```
walshmm, line 422:
```suggestion
        self.groceryClerk.name("inputWorkspace").neutron(request.runNumber).useLiteMode(
                request.useLiteMode
            )
        if not request.liveDataMode:
            self.groceryClerk.add()
        else:
            self.groceryClerk.liveData(duration=request.liveDataDuration).add()
```
?

src/snapred/backend/service/ReductionService.py
```diff
@@ -518,9 +532,34 @@ def artificialNormalization(self, request: CreateArtificialNormalizationRequest)
         return artificialNormWorkspace
 
     def grabWorkspaceforArtificialNorm(self, request: ReductionRequest):
+        # TODO: REBASE NOTE:
+        #   This method actually seems to be a reduction sub-recipe: something like `PreprocessArtificialNormalizationRecipe`.
+        #   It should not get "special treatment" in comparison to any other `ReductionRecipe` sub-recipes!
+        #   It also isn't obvious here that incoming pixel masks should be ignored?!
```
walshmm, line 538: this method just generates a "preview" workspace that allows the user to tweak the art norm ingredients that ultimately get used in reduction.  I agree there shouldnt be any special treatment, though Im not totally following the example provided.

If its in regards to just seeking a "column" grouping, that's compromise as it would be unfeasible to view the effect on every group, though I guess the artnorm preview could be expanded to let the user select a grouping from the dropdown...

src/snapred/meta/mantid/WorkspaceNameGenerator.py
```diff
@@ -66,12 +24,16 @@ class WorkspaceType(str, Enum):
     DIFFCAL_MASK = "diffCalMask"
     DIFFCAL_METRIC = "diffCalMetric"
     DIFFCAL_TIMED_METRIC = "diffCalTimedMetric"
+
+    # TODO: REBASE NOTE: inconsistent nameing:
```
walshmm, line 28:
```suggestion
    # TODO: REBASE NOTE: inconsistent naming:
```

src/snapred/ui/threading/worker_pool.py
```diff
@@ -52,6 +64,17 @@ def run(self):
         else:
             raise ValueError("Worker target must return a SNAPResponse object.")
 
+    @Slot()
+    def requestCancellation(self):
+        # Request cancellation at next interruption point.
+
+        # IMPORTANT: this method should actually be executed on the MAIN thread.
```
walshmm, line 71: should this throw an exception in that case?

src/snapred/ui/view/reduction/ReductionRequestView.py
```diff
@@ -1,37 +1,89 @@
...
+    def getRunNumbers(self) -> List[str]:
+        pass
+
+    @abstractmethod
+    def getPixelMasks(self) -> List[str]:
+        pass
+
+
+class _RequestView(_RequestViewBase):
+    liveDataModeChange = Signal(bool)
```
walshmm, line 71: does this signal need to be here _and_ the actual livedataview class?

src/snapred/ui/view/reduction/ReductionRequestView.py
```diff
@@ -143,16 +238,522 @@ def verify(self):
...
+        text = f"duration ({self._formatDuration(seconds)} < t0)" if seconds > 0 else "duration: use all available data"
+        self.duration.setLabelText(text)
+
+    @Slot(int)
+    def _updateUpdateInterval(self, seconds: int):
+        self.updateInterval.setLabelText(f"update interval (t0 < {self._formatDuration(seconds)})")
+
+    @ExceptionToErrLog
+    @Slot(bool)
+    def _populatePixelMaskDropdown(self, useLiteMode: bool):
```
walshmm, line 575: This seems to be the same as the previously defined one?

src/snapred/ui/widget/LEDIndicator.py
```diff
@@ -0,0 +1,239 @@
...
+        self._flashDurations = (0.2, 1.80)
+
+        self._flashTimer = QTimer(self)
+        #   A "chain" update is used here, to prevent runaway-timer issues.
+        self._flashTimer.setSingleShot(True)
+        self._flashTimer.setTimerType(Qt.CoarseTimer)
+        self._flashTimer.timeout.connect(self._continueFlash)
+
+        # For now, we use a class-local style sheet.
+        #   TODO: add `LEDIndicator` properties to the global style sheet.
```
walshmm, line 64: move to stylesheet file??

src/snapred/ui/workflow/ReductionWorkflow.py
```diff
@@ -280,46 +528,67 @@ def onArtificialNormalizationValueChange(self, smoothingValue, lss, decreasePara
         )
 
         response = self.request(path="reduction/artificialNormalization", payload=request_)
+
+        #
+        # TODO: why isn't this checking that the request actually succeeded?
```
walshmm, line 533: I think the assumption is that if there was an error, the code would go through the error handling routine instead of just returning.  
There is some code duplication here, I dont see any reason one couldnt call the other to some degree.

tests/unit/backend/data/test_Indexer.py
```diff
@@ -535,10 +539,13 @@ def test_nextVersion(self):
         assert indexer.nextVersion() == here + 1
 
         record = self.record(here + 1)
-        # NOTE: Writing aribitrary records to disk is not allowed.
+        # NOTE: Writing arbitrary records to disk is not allowed.
         #       Our code should never produce an unindexed record.
         #       What value is there in writing a record that is
         #       inherently missing metadata supplied by the index?
+
+        # Which kind of suggests that that data should have been a part of the record, does it not?  :)
```
walshmm, line 547: :smiling_face_with_tear:

tests/util/InstaEats.py
```diff
@@ -107,61 +112,150 @@ def fetchWorkspace(self, filePath: str, name: WorkspaceName, loader: str = "") -
...
+                    self._loadedRuns[key] = 0
+                    self._liveDataKeys.append(key)
+                    success = True
 
-        # create a copy of the raw data for use
-        workspaceName = self._createCopyNeutronWorkspaceName(runNumber, useLiteMode, self._loadedRuns[key] + 1)
-        data["result"] = True
-        data["workspace"] = workspaceName
-        self._loadedRuns[key] += 1
+                case (True, True, _, _):
```
walshmm, line 167: Related to the simplification I linked waaaay above^

src/snapred/backend/api/InterfaceController.py
ekapadi: I moved this to `WorkflowImplementer`.  I agree that this location makes more sense.  I'm not certain that the user gets the "User cancellation" log message quite as soon as happened before, but that's a minor issue.  **They do get the feedback that their cancellation was accepted in that the "Cancellation" button itself is disabled after it's clicked.**

src/snapred/backend/dao/ingredients/GroceryListItem.py
ekapadi: Reformatted.

src/snapred/backend/data/LocalDataService.py
ekapadi: I agree that this is tricky enough that it requires its own method.  Since it most directly relates to the content of the PV-logs, I moved it to `PV_logs_util.datetimeFromLogTime`.

src/snapred/backend/recipe/ReductionRecipe.py
ekapadi: **Yes it should.**  Somehow this flow of control needs to indicate that these workspaces are _prepared_ (or grouped or whatever) to be used later in the flow.  This current implementation has that happening implicitly.

src/snapred/backend/recipe/algorithm/FetchGroceriesAlgorithm.py
ekapadi: **For live data mode, it probably _will_ exist in the ADS.**  One thing we need to check during testing is that at the end of the workflow, the actual {target: lite-mode} input workspace isn't deleted -- maybe the end user wants a running plot of that workspace also?  I don't know.  **(This is probably another minor defect for the current live-data implementation.)**

Right now, at the end of the workflow that workspace _is_ deleted, which is OK because then at the previous cache-check location (in `GroceryService`) it will trigger a new load.  **-- All of this is not yet working in an ideal manner.  :(  --**

src/snapred/backend/service/ReductionService.py
ekapadi: Strange.  I initially had something like that, and then I backtracked because I thought the `GroceryListBuilder` token methods didn't return the builder itself.  Anyway, it's changed back now.

src/snapred/meta/mantid/WorkspaceNameGenerator.py
ekapadi: done.

src/snapred/ui/threading/worker_pool.py
ekapadi: I clarified the comment.  The issue is only that if it's called in the _executing_ thread, it never completes in a timely manner.  It took me a while to figure out why the only thing my "cancellation request" was doing was giving me an additional log message at the end of the entire workflow.  :(

src/snapred/ui/view/reduction/ReductionRequestView.py
ekapadi: Apart from the typo in there (after "Error retrieving...").  **Yes, for the moment it's a duplicate.**  The issue is that for each of the stacked layouts "their own" signal needs to trigger "their own" `_populatePixelMaskDropdown` slot.  `StackedLayout` is not ideal as it doesn't help much with something like this (or with whole sets of duplicated controls that need to "mirror" each other).  **The signal behavior is complicated enough that I left this alone and didn't try to simplify it further.**

src/snapred/ui/view/reduction/ReductionRequestView.py
ekapadi: Same issue as discussed next (i.e. with "mirroring" the panels).  The signals (, and some of the slots) are mirrored in the `_RequestView` and `_LiveDataView`.  What I want is for the signal from the various views to propagate up to the `ReductionRequestView` itself.    I'm certain this can be cleaned up a bit, but it was complicated enough to figure out I wanted to leave it alone, for the moment.

src/snapred/ui/widget/LEDIndicator.py
ekapadi: **I'd like to leave that as a "TODO" for the moment,** until I get some feedback from Malcolm about the panel details.  These properties will all be custom: that is, they use the `@Property` decorator and the `qproperty-<name>` syntax.  It always takes me an unexpected amount of time to debug the "selectors" and to get them to work.  So, I didn't want to do that for something that's a preliminary version < of what we've been calling the "run preview" >.

src/snapred/backend/service/ReductionService.py
ekapadi: Mostly my concern was that it had a different `fetchReductionGroceries` than the rest of the reduction process.  It seems to work OK with the live-data stuff.  But, the fact that there were multiple code-paths where the input workspaces could be loaded bugged me!  :)

src/snapred/backend/data/LocalDataService.py
ekapadi: `_loadLiveData` is marked as a _private_ method, but I'm not sure that will protect us much.  **The main reason I left it like this as it is that  it will be very useful for cis-test scripts.**

I was also really frustrated with the live-data system in that we need to load a whole workspace just to figure out (e.g.) what the run number is.  So I'm hoping that eventually we can change this.   As you'll see during testing: that load "for metadata only" can take a _long_ time.  I'd much rather the DAQ group would give us some kind of read-only access to their _process-variable_ network.

walshmm:
## Manual Testing

In my manual testing I added a code change to my local to enable testing this on analysis.  I added a testRunNumber field to the application.yml and I threw a FileNotFound exception when this was encountered during the pvfile read in LocalDataService.  This then triggers the live data flow for testing.

In addition I setup a facade ipts to house the test run data, and gave the test run data file a specific name which I propagated back to the application.yml file.

**NOTE:** If you encounter an "algorithm doesnt exist" or some error complaining about the highestVersion of livedata not found, this is simply an exception masking another issue.  For me, it was live data not finding the test file I was pointing at.

### Regression
Initializing state still works with LiveData test workspace.
image

Calibration completes successfully with the LiveData test runnumber
Normalization completes successfully with the above as well.
Reduction completes with the LiveData test runnumber, Im not sure it should? should it automatically engage live data mode? should live data have its own run number field?
image

### Feature Testing
UI/UX bugs should probably be written up and de-prioritized in an effort to deliver the live data functionality. 

1. Fallback Mode: I can confirm as i needed to trigger said fallback to test live data
2. Explicit Live Data Mode: can confirm that the live data flow only is active when selected from the reduction tab.
3. Once a run is added to the list and live data is selected it will attempt to connect to the live data server
image
3.x.1. if the run is cleared and live data is selected it will retain the previous run number used, attempt to reconnect but toss an error 
image
3.2. if no run number is selected outright it will attempt to connect and use the test runnumber.  Maybe this was a misconception on my part, or maybe this is just the default?  What would it do if no test runnumber was specified?
3.x.3. if continue is clicked before connection is established, a developer oriented error message is displayed
image
4. Clicking continue proceeds with the live reduction 
image
4.1. **NOTE:** if all groups are masked due to the calibration used during reduction, there will be NO OUTPUT REDUCED DATA, this relates to another story so for now I think this is expected. 
image
4.2. If an artificial normalization is require it will prompt the user for parameters once and reuse those values going forward
image
4.3 recalc art norm still works
4.4. live reduced data is available in the typical output workspaces and only update with new fully reduced datas, hidden intermediate workspaces are indeed effective.
image
4.x.5. I ran into a bug where data was accidentally deleted?  ~~Maybe I forgot to pull latest..~~ Reran on latest, still possible to occur. 1. Initialize state 2. difcal 59039, LA11B6 sample, column, -> tweak right xtal 2.9, max chi 4000 -> save with applies to >= 46680 -> live reduction with artificial norm -> not sure if relevant but recalc art norm with different vals in each field-> continue and wait.
image
4.6 cancel results in new message and executes at next interface controller request 
image
4.x.7 attempting to live reduce, replying no to art norm prompt, and them attempting to reduce again and selecting yes results in error message. It may be stuck in this state due to something else?  Requires full program restart.
image
image

Some of these workflow bugs might need to be looked at, but we also want livedata overall to be incorporated into the snapred backend.  I may also ask you to help Malcolm set some of this up in his reduction script?

I need to re-review the staled files, I dont think this regresses the other workflows and this gets us most of the way with live data as is.

### Acceptance Criteria (STALE)
- [.] live data checkbox is operational
- [.] selected range correctly filters data coming into workspace
- [.] successfully creates a workspace containing the live data
- [.] successfully reduces the input data, creating output data equivalent to normal output of reduction

src/snapred/backend/service/ReductionService.py
walshmm: Oh I see.  We would need to subdivide `fetchReductionGroceries` or just accept a near full load for the art norm tab then.

walshmm:
I think we should move forward with this pr once all the checks pass.  In total, an improvement to the codebase.

walshmm (approved):

PR #546, opened 2025-02-18
ReductionRecipe: check fully-masked group using PGP.isMasked flags

ekapadi:
## Description of work
This fixes a critical defect due to the previous implementation of the `_isGroupFullyMasked` method.  In addition, the current implementation makes use of the already calculated `PixelGroupingParameters.isMasked` flags, which are set for any fully-masked subgroups in a grouping schema.

## Explanation of work

  * Repair the implementation of `ReductionRecipe._isGroupFullyMasked` so it actually evaluates whether or not every pixel within every subgroup of a grouping schema is masked.

  The previous implementation of this method was based on the assumption that each spectrum of a grouping workspace contained a list of the indices of all of the pixels participating in a subgroup of the schema, with the subgroup-id corresponding to spectrum index.  On the contrary, each spectrum of a grouping workspace contains a single value, which is the subgroup-id of the pixel with the same detector-index as that the spectrum-index.  The net effect if this confusion was that any mask workspace that had a small number of the lowest-index pixels masked, corresponding to the number of subgroups in the schema (+ 1), would result in the method determining that every subgroup in the schema was fully masked, which was clearly incorrect.

### Dev testing
Existing unit tests have been modified to cover these new changes.

### CIS testing
To duplicate the original defect, please see the EWM item.

## Link to EWM item

[EWM#9603](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=9603)

### DIFF
--- a/src/snapred/backend/recipe/ReductionRecipe.py
+++ b/src/snapred/backend/recipe/ReductionRecipe.py
@@ -231,20 +231,9 @@ class ReductionRecipe(Recipe[Ingredients]):
             )
         return sampleClone, normalizationClone
 
-    def _isGroupFullyMasked(self, groupingWorkspace: str) -> bool:
-        maskWorkspace = self.mantidSnapper.mtd[self.maskWs]
-        groupWorkspace = self.mantidSnapper.mtd[groupingWorkspace]
-
-        totalMaskedPixels = 0
-        totalGroupPixels = 0
-
-        for i in range(groupWorkspace.getNumberHistograms()):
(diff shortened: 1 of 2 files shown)


darshdinger (approved):
Code looks good. Pytests pass. Correct behavior seen within SNAPRed on analysis using a completely masked masking workspace.

PR #550, opened 2025-02-21
Ewm7732 normalize by direct beam monitor

walshmm:
## Description of work

This adds a toggle to tell snapred to normalize by events from the monitor rather than proton charge.

## Explanation of work

I refactored `GroceryService` to remove the two complex case statements, and instead model a more accurate fallback structure for loading native/lite/live data.

I then added the loading of monitor workspaces to the end, save the total events of the processed monitor workspace as a log, and use said log in normalization calculations down the line.

## To test

Run each workflow, ensure that they still work with the toggle off.
Record results as far as normalization by current is concerned( like the log 'NormalizationFactor')
Run them again, ensure that they work with the toggle on.
Observe the monitor workspace get loaded and then quickly deleted during load.
Compare the normalization factor this time around, confirm that they are different and that the toggle was effective.

Ensure the results of each workflow are as expected, no out of the ordinary graphs for the normal testing data.

## Link to EWM item

[EWM#7732](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=7732)

### Verification

### Acceptance Criteria

- [ ] There is a toggle in the application.yml for enabling this flow, and a config for which monitor to use (defaults false, and 0)
- [ ] Workspaces have additional logs indicating that this was their method of normalization
- [ ] The workspaces are in fact normalized this way (i.e. different results in the case of reduction, normalization, ect)
- [ ] The story mentions needing different 'width'/'L1'/'particleBouncs' for the RemovePromptPulse on the monitor workspace but did not provide a location for said width, however using the same ones provided normally seemed to yield decent results.

### DIFF
--- a/src/snapred/backend/dao/ingredients/GroceryListItem.py
+++ b/src/snapred/backend/dao/ingredients/GroceryListItem.py
@@ -55,6 +55,7 @@ class GroceryListItem(BaseModel):
     loader: Literal[
         "",
         "LoadCalibrationWorkspaces",
+        "LoadNexusMonitors",
         "LoadEventNexus",
         "LoadGroupingDefinition",
         "LoadLiveData",
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -1205,7 +1205,10 @@ class GroceryService:
         return maskWSName
 
     def _createMonitorWorkspaceName(self, runNumber: str, useLiteMode: bool) -> WorkspaceName:
-        return wng.monitor().runNumber(runNumber).lite(useLiteMode).build()
+        builder = wng.monitor().runNumber(runNumber)
+        if useLiteMode:
+            builder.lite(wng.Lite.TRUE)
+        return builder.build()
 
     def fetchMonitorWorkspace(self, item: GroceryListItem) -> WorkspaceName:
         runNumber, useLiteMode = item.runNumber, False
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -3,7 +3,7 @@ import json
 from collections.abc import Iterable
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import numpy as np
 from mantid.dataobjects import MaskWorkspace
@@ -395,7 +395,7 @@ class GroceryService:
         """
         return mtd.doesExist(name)
 
-    def renameWorkspace(self, oldName: WorkspaceName, newName: WorkspaceName):
+    def renameWorkspace(self, oldName: WorkspaceName, newName: WorkspaceName) -> WorkspaceName:
         """
         Renames a workspace in Mantid's ADS.
 
@@ -404,12 +404,14 @@ class GroceryService:
         :param newName: the name to replace the workspace name in the ADS
         :type newName: WorkspaceName
         """
-        self.mantidSnapper.RenameWorkspace(
-            f"Renaming {oldName} to {newName}",
-            InputWorkspace=oldName,
-            OutputWorkspace=newName,
-        )
-        self.mantidSnapper.executeQueue()
+        if oldName != newName:
+            self.mantidSnapper.RenameWorkspace(
+                f"Renaming {oldName} to {newName}",
+                InputWorkspace=oldName,
+                OutputWorkspace=newName,
+            )
+            self.mantidSnapper.executeQueue()
+        return newName
 
     def renameWorkspaces(self, oldNames: List[WorkspaceName], newNames: List[WorkspaceName]):
         """
@@ -458,13 +460,13 @@ class GroceryService:
         :return: a pointer to the cloned workspace in the ADS
         :rtype: a python wrapped, C++ shared pointer to a Workspace
         """
-        from mantid.simpleapi import CloneWorkspace  # noqa : TID251
-
-        if self.workspaceDoesExist(name):
-            ws = CloneWorkspace(InputWorkspace=name, OutputWorkspace=copy)
-        else:
-            ws = None
-        return ws
+        self.mantidSnapper.CloneWorkspace(
+            f"Cloning {name} to {copy}",
+            InputWorkspace=name,
+            OutputWorkspace=copy,
+        )
+        self.mantidSnapper.executeQueue()
+        return copy
 
     def writeWorkspaceMetadataAsTags(self, workspaceName: WorkspaceName, workspaceMetadata: WorkspaceMetadata):
         """
@@ -748,7 +750,7 @@ class GroceryService:
                 raise RuntimeError(f"unable to load workspace {name} from {filePath}")
         return data
 
-    def fetchNeutronDataSingleUse(self, item: GroceryListItem) -> Dict[str, Any]:
+    def _fetchNeutronDataSingleUse(self, item: GroceryListItem, missingDataHandler: Callable) -> Dict[str, Any]:
         """
         Fetch a neutron data file, without copy-protection.
         If the workspace is truly only needed once, this saves time and memory.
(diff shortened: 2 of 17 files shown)


tests/unit/backend/data/test_DataFactoryService.py
```diff
@@ -296,7 +296,7 @@ def test_getCloneOfWorkspace(self):
         mtd.add(wsname1, ws1)
         assert self.instance.workspaceDoesExist(wsname1)
         ws2 = self.instance.getCloneOfWorkspace(wsname1, wsname2)
-        assert ws1.getComment() == ws2.getComment()
+        assert mtd[str(ws1)].getComment() == mtd[str(ws2)].getComment()
```
rboston628, line 299: What change made this change necessary?

src/snapred/backend/recipe/algorithm/RawVanadiumCorrectionAlgorithm.py
```diff
@@ -126,13 +126,24 @@ def PyExec(self):
 
         pcV = self.mantidSnapper.mtd[self.outputVanadiumWS].run().getProtonCharge()
         pcB = self.mantidSnapper.mtd[self.outputBackgroundWS].run().getProtonCharge()
-        protonCharge = pcV / pcB
+
+        if Config["mantid.workspace.monitor.normalizeByBeam"]:
+            pcV = self.mantidSnapper.mtd.getSNAPRedLog(self.outputVanadiumWS, "normalizeByMonitorFactor")
+            pcB = self.mantidSnapper.mtd.getSNAPRedLog(self.outputBackgroundWS, "normalizeByMonitorFactor")
```
rboston628, line 132: Remove L127, 128, and put them here after an `else`?

src/snapred/backend/recipe/ReductionGroupProcessingRecipe.py
```diff
@@ -57,9 +57,7 @@ def queueAlgos(self):
         )
 
         self.mantidSnapper.NormalizeByCurrentButTheCorrectWay(
-            "Normalizing Current ... but the correct way!",
-            InputWorkspace=self.outputWS,
-            OutputWorkspace=self.outputWS,
+            "Normalizing Current ... but the correct way!", InputWorkspace=self.outputWS, OutputWorkspace=self.outputWS
```
rboston628, line 60: can you add a `,` at the end here, so it undoes this change

src/snapred/backend/data/GroceryService.py
```diff
@@ -402,12 +402,14 @@ def renameWorkspace(self, oldName: WorkspaceName, newName: WorkspaceName):
...
         :type newName: WorkspaceName
         """
-        self.mantidSnapper.RenameWorkspace(
-            f"Renaming {oldName} to {newName}",
-            InputWorkspace=oldName,
-            OutputWorkspace=newName,
-        )
-        self.mantidSnapper.executeQueue()
+        if oldName != newName:
+            self.mantidSnapper.RenameWorkspace(
```
rboston628, line 406: I think this is already a no-op if the names are the same

src/snapred/backend/data/GroceryService.py
```diff
@@ -456,13 +458,13 @@ def getCloneOfWorkspace(self, name: WorkspaceName, copy: WorkspaceName):
...
-        else:
-            ws = None
-        return ws
+        self.mantidSnapper.CloneWorkspace(
+            f"Cloning {name} to {copy}",
+            InputWorkspace=name,
+            OutputWorkspace=copy,
+        )
+        self.mantidSnapper.executeQueue()
+        return copy
```
rboston628, line 467: The old return was a `MatrixWorkspace`, whereas now it is a string.  Does that change any other behavior?

src/snapred/backend/recipe/algorithm/NormalizeByCurrentButTheCorrectWay.py
```diff
@@ -25,21 +29,42 @@ def PyInit(self):
...
+                CloneWorkspace(
+                    InputWorkspace=workspace,
+                    OutputWorkspace=self.getPropertyValue("OutputWorkspace"),
+                )
+
+                workspace = self.getPropertyValue("OutputWorkspace")
+
+                Scale(InputWorkspace=workspace, Factor=normVal, Operation="Multiply")
+                # add the normalization factor to the workspace log
+                self.mantidSnapper.mtd[workspace].mutableRun().addProperty(
```
rboston628, line 56: Why can't you use normal `mtd`?

rboston628 (commented):
I still need to look at the GroceryService changes

tests/unit/backend/data/test_DataFactoryService.py
walshmm: its generally discouraged to pass around workspace pointers, as Pete has warned, they can go out of scope randomly.

src/snapred/backend/data/GroceryService.py
walshmm: as far as unit tests, integration tests, and manual tests, it doesnt seem so.  As stated above, workspace names are safer.

src/snapred/backend/data/GroceryService.py
walshmm: I could double check but I think I added this couple days ago because of an exception it tossed.

src/snapred/backend/recipe/algorithm/RawVanadiumCorrectionAlgorithm.py
walshmm: done!

src/snapred/backend/recipe/ReductionGroupProcessingRecipe.py
walshmm: done!

src/snapred/backend/data/GroceryService.py
walshmm: This tosses an exception if the names are the same
image

rboston628: Well, that's just dumb

ekapadi:
One question about how you recommend to _compare_ the normalization factors?  A step-by-step might be useful.  -- Thanks!

ekapadi:
**Monitor `WNG` workspace-name format is coming up as `monitor_false...` which is inconsistent with how the "lite-mode" token is formafted for `run`** -- I think the solution is to initialize your `WNG.monitor` method in exactly the same way as the `run` method.  (And possibly, _not_ to fix anything beyond that for this PR...  :)  )

ekapadi:
Without using monitor:  `NormalizationFactor = 2777.79`

With monitor: `NormalizationFactor = 1.465e-7`  ??? **This seems way to large of a change to me.**

walshmm:
> Without using monitor: `NormalizationFactor = 2777.79`
> 
> With monitor: `NormalizationFactor = 1.465e-7` ??? **This seems way to0 large of a change to me.** Do you have a simple way to explain why this is what I should be expecting? `NormalizationFactor` * `..normalizeByMonitorFactor` ~= 1.0

I could be recording the wrong value?  I may have mistook NormalizeByCurrent as saving the actual scaling factor used as opposed to proton charge(?)  I'll take a look.

walshmm:
Though this is a little weird as far as the background workspace is concerned, as we are not doing a typical normalization?  its some ratio of background to vanadium run monitor counts.  This would be difficult to record in a number close to 2777 I think?

walshmm:
The purpose of saving it to NormalizationFactor is so that later down the line it is not normalized again...

ekapadi (commented):
I added a few comments as "one offs" so I could do the review a little bit at a time.  So far the only issues are with the WNG-formatted _name_ of the monitor workspace, and with the extreme difference between the `NormalizationFactor` in the two modes.  All of the workflows work as expected (including the live-data workflow).  **I'm looking at the grocery-service changes next...**

src/snapred/backend/dao/ingredients/GroceryListItem.py
```diff
@@ -56,6 +56,7 @@ class GroceryListItem(BaseModel):
         "",
         "LoadCalibrationWorkspaces",
         "LoadEventNexus",
+        "LoadNexusMonitors",
```
ekapadi, line 59: Suggest: in alphabetical order -- that's what I had been doing here.  :)

src/snapred/backend/data/GroceryService.py
```diff
@@ -763,120 +765,137 @@ def fetchNeutronDataSingleUse(self, item: GroceryListItem) -> Dict[str, Any]:
...
 
-        workspaceName: str = self._createNeutronWorkspaceName(runNumber, useLiteMode)
+        # This checks to make sure existing cache is valid
+        self._updateNeutronCacheFromADS(runNumber, useLiteMode)
 
-        success = False
-        convertToLiteMode = False
+        # 1. check cache
+        # 2. check if file exists
+        # e. cannot find data.
```
ekapadi, line 780: 'e.' or '3.'?

src/snapred/backend/data/GroceryService.py
```diff
@@ -763,120 +765,137 @@ def fetchNeutronDataSingleUse(self, item: GroceryListItem) -> Dict[str, Any]:
...
+        elif filepath.exists():
+            workspaceName = self._createNeutronWorkspaceName(runNumber, useLiteMode)
+            data = self.grocer.executeRecipe(str(filepath), workspaceName, loader)
+        else:
+            data = missingDataHandler()
+            workspaceName = data["workspace"]
 
-        self._updateNeutronCacheFromADS(runNumber, useLiteMode)
+        data["workspace"] = workspaceName
+        # NOTE: When would it ever be false?  If the workspace is not found, it should raise an error.
```
ekapadi, line 794: We are not _necessarily_ consistent about this.  Sometimes the return value is used, and sometimes it's not.  This is not a good thing!

src/snapred/backend/data/GroceryService.py
```diff
@@ -763,120 +765,137 @@ def fetchNeutronDataSingleUse(self, item: GroceryListItem) -> Dict[str, Any]:
...
+        else:
+            raise ConnectionError("no live-data connection is available")
+        return data
 
-                case (False, _, _, True):
-                    # native mode and native exists on disk
-                    data = self.grocer.executeRecipe(str(nativeModeFilePath), workspaceName, loader)
-                    success = True
+    def _fetchNeutronDataNative(self, item: GroceryListItem) -> Dict[str, Any]:
+        item.useLiteMode = False
```
ekapadi, line 836: Changing the value on the _incoming_ `item` isn't a good idea.  Definitely you need to make a copy here!

src/snapred/backend/data/GroceryService.py
```diff
@@ -763,120 +765,137 @@ def fetchNeutronDataSingleUse(self, item: GroceryListItem) -> Dict[str, Any]:
...
-        if success and convertToLiteMode:
-            # fetch single-use does not export converted to lite data
-            self.convertToLiteMode(workspaceName, export=False)
-            data["workspace"] = workspaceName
-            self._processNeutronDataCopy(runNumber, workspaceName)
-
+        return self._fetchNeutronDataSingleUse(item, tryLiveData)
+
+    def _fetchNeutronDataLite(self, item: GroceryListItem, export=False) -> Dict[str, Any]:
+        item.useLiteMode = True
```
ekapadi, line 852: Same comment as previous.  It makes it _really_ difficult to trace what is happening.  So definitely make a copy of the `item` before modifying it.

src/snapred/backend/data/GroceryService.py
```diff
@@ -763,120 +765,137 @@ def fetchNeutronDataSingleUse(self, item: GroceryListItem) -> Dict[str, Any]:
...
+
+        def tryNative():
+            data = self._fetchNeutronDataNative(item)
+            data["fromNative"] = data["workspace"]
+            item.useLiteMode = True
```
ekapadi, line 857: And again...

src/snapred/backend/data/GroceryService.py
```diff
@@ -763,120 +765,137 @@ def fetchNeutronDataSingleUse(self, item: GroceryListItem) -> Dict[str, Any]:
...
-            data["workspace"] = workspaceName
-            self._processNeutronDataCopy(runNumber, workspaceName)
-
+        return self._fetchNeutronDataSingleUse(item, tryLiveData)
+
+    def _fetchNeutronDataLite(self, item: GroceryListItem, export=False) -> Dict[str, Any]:
+        item.useLiteMode = True
+
+        def tryNative():
+            data = self._fetchNeutronDataNative(item)
```
ekapadi, line 855: I spent quite a bit of time trying to figure out _where_ you are loading the lite data _from disk_ if there is a converted copy already available _on disk._  Would you please "walk me through" that scenario, (or e.g. point me to those lines of code). -- Thanks!

src/snapred/backend/data/GroceryService.py
```diff
@@ -1276,6 +1201,22 @@ def fetchCompatiblePixelMask(self, maskWSName: WorkspaceName, runNumber: str, us
...
 
+    def _createMonitorWorkspaceName(self, runNumber: str, useLiteMode: bool) -> WorkspaceName:
+        return wng.monitor().runNumber(runNumber).lite(useLiteMode).build()
+
+    def fetchMonitorWorkspace(self, item: GroceryListItem) -> WorkspaceName:
+        runNumber, useLiteMode = item.runNumber, False
+        workspaceName = self._createMonitorWorkspaceName(runNumber, useLiteMode)
+
+        # TODO: Cache if necessary
+        if self.workspaceDoesExist(workspaceName):
```
ekapadi, line 1212: Do we do this type of caching for any other type of workspace?  If not, I think we need to be consistent about how we cache.  That is, you probably need a `_loadedMonitors` dict.

src/snapred/backend/data/GroceryService.py
```diff
@@ -1490,7 +1432,7 @@ def getResidentWorkspaces(self, excludeCache: bool):
             workspaces = workspaces.difference(self.getCachedWorkspaces())
         return list(workspaces)
 
-    def _processNeutronDataCopy(self, runNumber, workspaceName):
+    def _filterEvents(self, runNumber, workspaceName):
```
ekapadi, line 1435: Are we moving the `removePromptPulse` here as well, or is that part of _another_ story?

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -34,6 +34,9 @@ def unique_name(self, n=5, prefix="", suffix=""):
     def unique_hidden_name(self):
         return mtd.unique_hidden_name()
 
+    def getSNAPRedLog(self, wsName, logname):
+        return self[wsName].getRun().getLogData(f"{Config['metadata.tagPrefix']}{logname}").value
```
ekapadi, line 38: I really, _really_ want this to throw a `KeyError` if the log isn't there.  If I'm remembering correcty, that's _not_ what `mantid.api.Run` does.  Can we please wrap this in  a `try`?

src/snapred/backend/recipe/algorithm/NormalizeByCurrentButTheCorrectWay.py
```diff
@@ -1,13 +1,17 @@
 from mantid.api import AlgorithmFactory, MatrixWorkspaceProperty, PropertyMode, PythonAlgorithm
 from mantid.kernel import Direction
-from mantid.simpleapi import CloneWorkspace, NormaliseByCurrent, mtd
+from mantid.simpleapi import CloneWorkspace, NormaliseByCurrent, Scale
```
ekapadi, line 3: Why are we using `MantidSnapper` inconsistently here?

src/snapred/resources/application.yml
```diff
@@ -157,6 +157,9 @@ reduction:
 
 mantid:
   workspace:
+    monitor:
+      normalizeByBeam: false
+      normID: 0
```
ekapadi, line 162: **I suggest expanding these names slightly (e.g.) `normalizeByBeamMonitor` and `normalizationMonitorID` (**or just `monitorId`).  Also we should check whether it's "Id" or "ID", to be consistent.

ekapadi (commented):
**I really like the new flow-control changes in `GroceryService` -- it _tremendously_ simplifies things!**  I left a few comments about remaining questions that I have, but once those are dealt with I am fine with all of these changes.  (Provided we set "liveData.enabled" to `False` for  the moment -- I don't think the mutex error is caused by anything in this PR.)

src/snapred/backend/data/GroceryService.py
walshmm: prompt pulse has already been moved here, that change is in next.

src/snapred/backend/data/GroceryService.py
walshmm: Updated to make a deep copy instead.

src/snapred/backend/data/GroceryService.py
walshmm: I removed these lines, loading monitors even for larger runs is ~1-2 seconds.  Adding the same caching scheme would complicate this, especially since I'm just deleting the workspace immediately

ekapadi (approved):
I ran through all of the workflows, including enabling the live data and running that against the "fake" listener.  (This is the live-data case that works on "next" right now.)  All of the workflows work as they should, including the "direct to live data" and the "fallback cases".  So I'm happy to approve this!

PR #551, opened 2025-02-21
Added dictionary to MantidSnapper to keep track of workspaces 

dlcaballero16:
## Description of work

MantidSnapper now has a dictionary that keeps track of workspaces Snapper has created

## Explanation of work

Dictionary is keyed on workspace names, and value is a tuple of the workspace itself plus the algorithm that created the workspace. Error checking is done to make sure the workspace actually exists in the dictionary before deleting. A getWorkspace function was added to safely retrieve worksapces.

### Dev testing

Open workbench on analysis and run this simple script:

``` python
from snapred.backend.dao.reduction.ReductionRecord import ReductionRecord
from snapred.backend.recipe.algorithm.MantidSnapper import MantidSnapper
from snapred.backend.recipe.algorithm.Utensils import Utensils
import h5py
from snapred.backend.data.NexusHDF5Metadata import NexusHDF5Metadata as n5m
from snapred.meta.redantic import parse_file_as
import time

inputFile = "/SNS/users/wqp/SNAP/IPTS-24641/shared/SNAPRed/04bd2c53f6bf6754/lite/46680/2024-11-26T121144/reduced_046680_2024-11-26T121144.nxs.h5"
outputFile = "/SNS/users/wqp/tmp/test_segfault_append.nxs"
record = parse_file_as(ReductionRecord, "/SNS/users/wqp/SNAP/IPTS-24641/shared/SNAPRed/04bd2c53f6bf6754/lite/46680/2024-11-26T121144/ReductionRecord.json")

untensils = Utensils()
untensils.PyInit()
mantidSnapper = untensils.mantidSnapper

ws = mantidSnapper.LoadNexus("..", Filename=inputFile, OutputWorkspace="ws")
mantidSnapper.executeQueue()
#This will pass
mantidSnapper.getWorkspace("ws")
#This should fail since "wds" does not exist
mantidSnapper.getWorkspace("wds")
```

### CIS testing

None, this only used internally.

## Link to EWM item

[EWM#9389](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=9389)

### Verification

### Acceptance Criteria

- [x] make sure workspaces are correctly added to dictionary

### DIFF
--- a/src/snapred/backend/recipe/algorithm/MantidSnapper.py
+++ b/src/snapred/backend/recipe/algorithm/MantidSnapper.py
@@ -47,6 +47,8 @@ class MantidSnapper:
 
     typeTranslationTable = {"string": str, "number": float, "dbl list": list, "boolean": bool}
     _mtd = _CustomMtd()
+
+    _infoMutex = Lock()
     _workspacesInfo: Dict[str, Dict[str, Any]] = {}
 
     def __init__(self, parentAlgorithm, name):
@@ -265,11 +267,13 @@ class MantidSnapper:
         self._algorithmQueue = []
 
     @classmethod
-    def _addWorkspaceInfo(cls, workspaceName: str, algorithmName: str, workspaceProperty: str):
+    def _addWorkspaceInfo(cls, workspaceName: str, algorithmName: str, propertyName: str):
+        cls._infoMutex.acquire()
         cls._workspacesInfo[workspaceName] = {
-            "propertyName": workspaceProperty,
+            "propertyName": propertyName,
             "algorithm": algorithmName,
         }
+        cls._infoMutex.release()
 
     @classmethod
     def getWorkspacesInfo(cls) -> Dict[str, Dict[str, Any]]:
@@ -279,4 +283,5 @@ class MantidSnapper:
         :return: dict by workspace-name key of information about the output workspaces
         :rtype: Dict[str, Dict[str, Any]]
         """
+        # WARNING: for access that isn't READ-ONLY, synchronization will be required!
         return cls._workspacesInfo
--- a/src/snapred/backend/recipe/algorithm/MantidSnapper.py
+++ b/src/snapred/backend/recipe/algorithm/MantidSnapper.py
@@ -262,11 +262,10 @@ class MantidSnapper:
         self._algorithmQueue = []
 
     @classmethod
-    def _addWorkspaceInfo(cls, name, workspaceProperty: Property):
+    def _addWorkspaceInfo(cls, algorithmName: str, workspaceProperty: Property):
         cls._workspacesInfo[workspaceProperty.valueAsStr] = {
-            "name": workspaceProperty.valueAsStr,
-            "algorithm": name,
-            "Direction": workspaceProperty.direction,
+            "propertyName": workspaceProperty.name,
+            "algorithm": algorithmName,
         }
 
     @classmethod
@@ -274,7 +273,7 @@ class MantidSnapper:
         """
         Grab dictionary of workspaces created by MantidSnapper
 
-        :return: dict of workspace names containing dicts of the workspace name and which algo created it
+        :return: dict by workspace-name key of information about the output workspaces
         :rtype: Dict[str, Dict[str, Any]]
         """
         return cls._workspacesInfo
--- a/src/snapred/backend/recipe/algorithm/MantidSnapper.py
+++ b/src/snapred/backend/recipe/algorithm/MantidSnapper.py
@@ -199,7 +199,8 @@ class MantidSnapper:
                     returnVal = access_pointer(returnVal)
                 val.update(returnVal)
                 if "Workspace" in str(algorithm.getProperty(prop)):
-                    MantidSnapper._addWorkspaceInfo(returnVal, name)
+                    if returnVal != "":
+                        MantidSnapper._addWorkspaceInfo(returnVal, name)
         except (RuntimeError, TypeError) as e:
             logger.error(f"Algorithm {name} failed for the following arguments: \n {kwargs}")
             self.cleanup()
--- a/src/snapred/backend/recipe/algorithm/MantidSnapper.py
+++ b/src/snapred/backend/recipe/algorithm/MantidSnapper.py
@@ -1,10 +1,10 @@
 import time
 from collections import namedtuple
 from threading import Lock
-from typing import Dict
+from typing import Any, Dict
 
 from mantid.api import AlgorithmManager, IWorkspaceProperty, Progress, mtd
-from mantid.kernel import Direction
+from mantid.kernel import Direction, Property
 from mantid.kernel import ULongLongPropertyWithValue as PointerProperty
 
 from snapred.backend.error.AlgorithmException import AlgorithmException
@@ -47,7 +47,7 @@ class MantidSnapper:
 
     typeTranslationTable = {"string": str, "number": float, "dbl list": list, "boolean": bool}
     _mtd = _CustomMtd()
-    _workspaceInfo: Dict[str, Dict[str, str]] = {}
+    _workspacesInfo: Dict[str, Dict[str, Any]] = {}
 
     def __init__(self, parentAlgorithm, name):
         """
@@ -198,11 +198,10 @@ class MantidSnapper:
                 if isinstance(algorithm.getProperty(prop), PointerProperty):
                     returnVal = access_pointer(returnVal)
                 val.update(returnVal)
-                if (
-                    isinstance(algorithm.getProperty(prop), IWorkspaceProperty)
-                    and not algorithm.getProperty(prop).isDefault
+                if isinstance(algorithm.getProperty(prop), IWorkspaceProperty) and not (
+                    algorithm.getProperty(prop).isDefault and algorithm.getProperty(prop).isOptional()
                 ):
-                    MantidSnapper._addWorkspaceInfo(returnVal, name)
+                    MantidSnapper._addWorkspaceInfo(name, algorithm.getProperty(prop))
         except (RuntimeError, TypeError) as e:
             logger.error(f"Algorithm {name} failed for the following arguments: \n {kwargs}")
             self.cleanup()
@@ -263,15 +262,19 @@ class MantidSnapper:
         self._algorithmQueue = []
 
     @classmethod
-    def _addWorkspaceInfo(cls, val, name):
-        cls._workspaceInfo[val] = {"name": val, "algorithm": name}
+    def _addWorkspaceInfo(cls, name, workspaceProperty: Property):
+        cls._workspacesInfo[workspaceProperty.valueAsStr] = {
+            "name": workspaceProperty.valueAsStr,
+            "algorithm": name,
+            "Direction": workspaceProperty.direction,
+        }
 
     @classmethod
-    def getWorkspaceInfo(cls) -> Dict[str, Dict[str, str]]:
+    def getWorkspacesInfo(cls) -> Dict[str, Dict[str, Any]]:
         """
         Grab dictionary of workspaces created by MantidSnapper
 
         :return: dict of workspace names containing dicts of the workspace name and which algo created it
-        :rtype: Dict[str, Dict[str, str]]
+        :rtype: Dict[str, Dict[str, Any]]
         """
-        return cls._workspaceInfo
+        return cls._workspacesInfo
--- a/src/snapred/backend/recipe/algorithm/MantidSnapper.py
+++ b/src/snapred/backend/recipe/algorithm/MantidSnapper.py
@@ -1,6 +1,7 @@
 import time
 from collections import namedtuple
 from threading import Lock
+from typing import Dict
 
 from mantid.api import AlgorithmManager, Progress, mtd
 from mantid.kernel import Direction
@@ -46,6 +47,7 @@ class MantidSnapper:
 
     typeTranslationTable = {"string": str, "number": float, "dbl list": list, "boolean": bool}
     _mtd = _CustomMtd()
+    workspaceDict: Dict[str, Dict[str, str]] = {}
 
     def __init__(self, parentAlgorithm, name):
         """
@@ -84,7 +86,6 @@ class MantidSnapper:
         self._export = False
         self.timeout = 60  # seconds
         self.checkInterval = 0.05  # 50 ms
-        self.workspaceDict = {}
         if self._export:
             self._cleanOldExport()
 
@@ -197,7 +198,7 @@ class MantidSnapper:
                 if isinstance(algorithm.getProperty(prop), PointerProperty):
                     returnVal = access_pointer(returnVal)
                 val.update(returnVal)
-                self._addWorkspaceToDict(prop, val, name)
+                self._addWorkspaceToDict(prop, returnVal, name)
         except (RuntimeError, TypeError) as e:
             logger.error(f"Algorithm {name} failed for the following arguments: \n {kwargs}")
             self.cleanup()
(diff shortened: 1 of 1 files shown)


src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -254,3 +256,18 @@ def cleanup(self):
             self._prog_reporter.report(self._endrange, "Done")
         self._progressCounter = 0
         self._algorithmQueue = []
+
+    def _addWorkspaceToDict(self, prop, val, name):
+        if prop == "OutputWorkspace":
+            if self.mtd.doesExist(val):
+                self.workspaceDict[val] = (self.mtd[val], name)
```
peterfpeterson, line 263: You are asking for trouble here. The workspace references have a habit of getting invalidated because ADS (called `mtd` here) doesn't know to update viewers of workspaces. There is a framework for that, but it would take some time to find it.

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -254,3 +256,18 @@ def cleanup(self):
...
+            elif val in self.workspaceDict:
+                del self.workspaceDict[val]
+
+    def getWorkspace(self, workspace: str):
```
peterfpeterson, line 267: document that this is returning a tuple of `(wksp, name)` which is a strange thing since I imagine all you want is the workspaces.

dlcaballero16: Kort had suggested that we could also keep track of which algorithm created the workspace. I'll add documentation

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -254,3 +257,28 @@ def cleanup(self):
             self._prog_reporter.report(self._endrange, "Done")
         self._progressCounter = 0
         self._algorithmQueue = []
+
+    def _addWorkspaceToDict(self, prop, val, name):
+        if prop == "OutputWorkspace":
```
ekapadi, line 262: Here you probably need to check the \<option type\>, rather than the _name_ of the property.  For example if it's `InOut` workspace it's unlikely that it will have the name `OutputWorkspace`.  **Also: please add the \<option type\> to the dict.**

ekapadi: Also: is this called at the point the the workspace will definitely already exist?

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -254,3 +257,28 @@ def cleanup(self):
...
+        :type workspace: str
+        :return: dictionary containing the workspace name and which algo created it
+        :rtype: Dict[str, str]
+        """
+        if self.mtd.doesExist(workspace):
+            return self.workspaceDict[workspace]
+        elif workspace in self.workspaceDict:
+            del self.workspaceDict[workspace]
+        else:
+            raise ValueError(f"Workspace name {workspace} does not exist in MantidSnapper")
```
ekapadi, line 284: `workspaceDict` should be `_workspaceDict` -- it's a private attribute.

**This probably should be a `RuntimeError`, as it indicates that something happened with respect to our _programming_ that we didn't expect.**  I'd also suggest that the exception be raised unless the workspace name is in _both_ places.   (And, I'm thinking that any `del` should probably be an explicit method -- see next.)

It looks like you need `@classmethod clearWorkspaceInfo` and `@classmethod delWorkspaceInfo(name: str)` as well.

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -254,3 +257,28 @@ def cleanup(self):
...
         self._algorithmQueue = []
+
+    def _addWorkspaceToDict(self, prop, val, name):
+        if prop == "OutputWorkspace":
+            if self.mtd.doesExist(val):
+                self.workspaceDict[val] = {"name": val, "algorithm": name}
+            elif val in self.workspaceDict:
+                del self.workspaceDict[val]
+
+    def getWorkspaceInfo(self, workspace: str) -> Dict[str, str]:
```
ekapadi, line 268: `@classmethod def getWorkspaceInfo(cls, ...)`

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -254,3 +257,28 @@ def cleanup(self):
             self._prog_reporter.report(self._endrange, "Done")
         self._progressCounter = 0
         self._algorithmQueue = []
+
+    def _addWorkspaceToDict(self, prop, val, name):
```
ekapadi, line 261: `@classmethod def _addWorkspaceInfo` (note the name change:  in general mentioning the "type" of something in its name is going to be _redundant_.)

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -196,6 +198,7 @@ def executeAlgorithm(self, name, outputs, **kwargs):
                 if isinstance(algorithm.getProperty(prop), PointerProperty):
                     returnVal = access_pointer(returnVal)
                 val.update(returnVal)
+                self._addWorkspaceToDict(prop, returnVal, name)
```
ekapadi, line 201: `self.__class__._addWorkspaceInfo(...)` or (better) `MantidSnapper._addWorkspaceInfo(...)`.

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -46,6 +47,7 @@ class MantidSnapper:
 
     typeTranslationTable = {"string": str, "number": float, "dbl list": list, "boolean": bool}
     _mtd = _CustomMtd()
+    workspaceDict: Dict[str, Dict[str, str]] = {}
```
ekapadi, line 50: Same comment as below:  I'd call this `_workspaceInfo` (the 'dict' is kind-of obvious  :)  ).

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -254,3 +257,28 @@ def cleanup(self):
             self._prog_reporter.report(self._endrange, "Done")
         self._progressCounter = 0
         self._algorithmQueue = []
+
+    def _addWorkspaceToDict(self, prop, val, name):
+        if prop == "OutputWorkspace":
+            if self.mtd.doesExist(val):
+                self.workspaceDict[val] = {"name": val, "algorithm": name}
```
ekapadi, line 264: This _does_ add the value to the class dict, but it's _really_ ambiguous what's going on.  The best solution is to make this a `@classmethod`, and then it will be explicit.

src/snapred/backend/recipe/algorithm/MantidSnapper.py
dlcaballero16: Yes, this is called after the algorithm is executed, so the workspace will exist by then

src/snapred/backend/recipe/algorithm/MantidSnapper.py
ekapadi: **Adding this from our "slack" thread:**
"""
I'm suggesting that you make one more pass and (1) move most of the methods to @classmethod, and (2) just return the "dict" via a getWorkspacesInfo method -- then we don't need to figure out right now, all of the issues associated with how this is going to be used.  That is MantidSnapper adds to the workspace info, but we can then do whatever we want to with the resulting dict.  (For example, I'm not sure if it's even an error if the workspace is not in the ADS.  I don't think we know for certain yet.  If you return the dict and it throws a KeyError when we ask for a workspace that's not there -- that's just fine.)
And finally, (3) This is the most important thing -- you need to look at the type of the argument:  e.g. Output Input InputOutput etc. -- otherwise you miss many output workspaces that may have different names besides OutputWorkspace.  If you save the argument type as one of the fields in the workspaceInfo , then you should probably also be recording InOut workspaces as well -- that is, anything that has passed through MantidSnapper as an output workspace, regardless of whether MantidSnapper explicitly created it.  If you record all of this, and then later I only need the output workspaces I can do
```
[k for k, v in MantidSnapper.getWorkspacesInfo().items() if v["argType"] == Output]  # not sure the "type" of the argType, 
```
but you can look it up...
"""

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -46,6 +47,7 @@ class MantidSnapper:
 
     typeTranslationTable = {"string": str, "number": float, "dbl list": list, "boolean": bool}
     _mtd = _CustomMtd()
+    _workspaceInfo: Dict[str, Dict[str, str]] = {}
```
ekapadi, line 50: `Dict[str, Dict[str, Any]]` is probably better -- then we can save things like `Direction` and `PropertyMode` as _themselves_ and not as "str".

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -196,6 +198,11 @@ def executeAlgorithm(self, name, outputs, **kwargs):
                 if isinstance(algorithm.getProperty(prop), PointerProperty):
                     returnVal = access_pointer(returnVal)
                 val.update(returnVal)
+                if (
+                    isinstance(algorithm.getProperty(prop), IWorkspaceProperty)
+                    and not algorithm.getProperty(prop).isDefault
```
ekapadi, line 203: **A slight change with this** -- I realized yesterday evening that (to identify the "empty" case) it is _probably_, more correctly: `<property>.isDefault and <property>.mode == PropertyMode.Optional`.  Possibly nitpicking, but I can imagine a case where we have a non-optional output workspace that has a _default_ name.  (Note:  I'm not sure if the \<property\> field name is `.mode`...)

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -254,3 +261,17 @@ def cleanup(self):
             self._prog_reporter.report(self._endrange, "Done")
         self._progressCounter = 0
         self._algorithmQueue = []
+
+    @classmethod
+    def _addWorkspaceInfo(cls, val, name):
+        cls._workspaceInfo[val] = {"name": val, "algorithm": name}
+
+    @classmethod
+    def getWorkspaceInfo(cls) -> Dict[str, Dict[str, str]]:
```
ekapadi, line 270: I suggest `getWorkspacesInfo` (i.e. as a plural), with the same comment as prev. RE the return type.

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -254,3 +261,17 @@ def cleanup(self):
             self._prog_reporter.report(self._endrange, "Done")
         self._progressCounter = 0
         self._algorithmQueue = []
+
+    @classmethod
+    def _addWorkspaceInfo(cls, val, name):
+        cls._workspaceInfo[val] = {"name": val, "algorithm": name}
```
ekapadi, line 267: Also please add `direction: <the actual direction value, not the str...>`.  I think the best way to do this is to just pass in the `<property>`.  That is: `def _addWorkspaceInfo(cls, workspaceProperty: Property) ` and then you have access to whatever you need.  We _do_ however, want to split out the property into its fields:  I'm not sure how _persistent_ the actual property objects will be if we don't do this.

ekapadi (commented):
Only a few minor issues.  Things are looking good!

src/snapred/backend/recipe/algorithm/MantidSnapper.py
dlcaballero16: I couldn't use only the workspace property as it does not contain the algorithm name it comes from, or least not in an obvious way

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -254,3 +260,21 @@ def cleanup(self):
...
+            "algorithm": name,
+            "Direction": workspaceProperty.direction,
+        }
+
+    @classmethod
+    def getWorkspacesInfo(cls) -> Dict[str, Dict[str, Any]]:
+        """
+        Grab dictionary of workspaces created by MantidSnapper
+
+        :return: dict of workspace names containing dicts of the workspace name and which algo created it
```
ekapadi, line 277: This should read "dict by workspace-name key of information about the output workspaces" -- it _should_ be obvious from what the field names (in the "info" dict) what the _values_ mean.   

As per my recent slack message, before slack went down (for me, at least):  {"propertyName": /<the property name/>, "algorithm": /< the algorithm name />} -- and then anything else we should be able to look up from those values.  

That is, despite what I said before, we don't need "direction" or anything else, but without the "propertyName", we don't really know where the workspace came from (for example, it could have a _random_ hidden name).

src/snapred/backend/recipe/algorithm/MantidSnapper.py
ekapadi: No longer relevant.

src/snapred/backend/recipe/algorithm/MantidSnapper.py
ekapadi: no longer relevant

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -254,3 +260,21 @@ def cleanup(self):
...
         self._progressCounter = 0
         self._algorithmQueue = []
+
+    @classmethod
+    def _addWorkspaceInfo(cls, name, workspaceProperty: Property):
+        cls._workspacesInfo[workspaceProperty.valueAsStr] = {
+            "name": workspaceProperty.valueAsStr,
+            "algorithm": name,
+            "Direction": workspaceProperty.direction,
+        }
```
ekapadi, line 270: Please add the "propertyName" and _remove_ the `<workspace name>` -- we already have that as the _key_.

src/snapred/backend/recipe/algorithm/MantidSnapper.py
ekapadi: That's just fine.  I checked what you did in the code and it is OK.

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -254,3 +260,21 @@ def cleanup(self):
             self._prog_reporter.report(self._endrange, "Done")
         self._progressCounter = 0
         self._algorithmQueue = []
+
+    @classmethod
+    def _addWorkspaceInfo(cls, name, workspaceProperty: Property):
+        cls._workspacesInfo[workspaceProperty.valueAsStr] = {
```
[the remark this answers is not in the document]
ekapadi: And also then `"propertyName": workspaceProperty.name`

src/snapred/backend/recipe/algorithm/MantidSnapper.py
ekapadi: That's a good point.  Just so we're not passing in a whole lot of stuff that we don't need to.

ekapadi (approved):
**In working through the PR and code review of this story,  it became evident that quite a bit depends on how this feature will actually be used.**
For this reason, I recommended that we do _not_ evict entries from the `workspacesInfo` dict based on whether or not the workspace is actually present in the ADS.
If this is something that turns out to be required, that "one line" change can be added back in at that time.

For example, if we wanted to see which workspaces _had_ been created by a workflow, which is a quite reasonable usage scenario, only looking at workspaces that still continue to exist wouldn't be part of the requirement.

I have tested, and have gone over all of these changes in detail, and I am satisfied with the code.  I approve these changes.

PR #557, opened 2025-03-06
Revert "Added dictionary to MantidSnapper to keep track of workspaces "

walshmm:
Reverts neutrons/SNAPRed#551


walshmm:
this revert segfaulted, might not be this.

PR #566, opened 2025-03-16
Stability fixups for live-data mode

ekapadi:
## Description of work

The majority of the changes implementing SNAPRed's live-data capability were merged at the PR for EWM#7436 (SNAPRed PR#541). This commit deals with additional required changes discovered during testing with the actual live-data listener.

## Explanation of work

**This commit includes the following changes:**

  * During testing it was found that almost always the instrument-parameter related PV-logs of the live-data input workspaces have been incompletely initialized.  The utility method `PV_logs_util.mappingFromRun` has been modified to optionally use a list of alternative keys in order to more reliably obtain several of the instrument-parameter values.  These keys are specified in a new nested-list structure in `application.yml` stored at the compound key `instrument.PVLogs.instrumentKeys`;

  * Log a warning when the live-data fallback has been triggered -- this warning is only logged during the fallback case.

  * Allow for the possibility that no IPTS directory exists at all -- this seems to be a common live-data case when diagnostic runs are being made with the instrument;  (Note that Mantid `GetIPTS` sometimes locks up when this case occurs.  This will be reported as a separate DEFECT.)

  * Modifications to `LocalDataService.getIPTS` to allow full caching of both IPTS-directories, and also the case when "no IPTS directory exists";

  * Full caching of both `LocalDataService.generateStateId`, and `LocalDataService.readDetectorState` methods, allowing far fewer calles to `GetIPTS`.  (Primarily this is a work around to the `GetIPTS` lock-up issue.)

  * Allow `LocalDataService.readDetectorState` to return `None` when no run is active.

  * Reduce logging output during live-data mode as much as possible.  This includes both silencing the `LoadLiveData` and `GetIPTS` algorithm by declaring them as "quiet algorithms` (in `MantidSnapper`), but also modifying some traceback logging so that it only occurs at the `debug` logging level;  (A major issue when running live-data mode when `debug` logging is enabled has been that Mantid's logging system locks up, due to excessive output to the "messages" panel.  In addition, the `LoadLiveData` algorithm does not correctly initialize the logging level of it's live-data listener.  Both of these issues will also be written up as separate DEFECTs.)

  * Log a `WARNING` if the "messages" panel logging level is set to `DEBUG` in MantidWorkbench when SNAPRed is initialized.  Although SNAPRed's live data mode will run correctly in this case, when the user exits SNAPRed, and the "messages" panel is reset to this `DEBUG` level, the log messages from the live-data listener will lock-up Mantid's logging system;

  * Change all of the reduction-workflow "exclude" workspaces lists to `set`.  This is more consistent with how this information is actually used within the workflow;

  * More robust treatment of `LiveMetadata` load, adding reasonable default values when some values are not present in the logs.  Primarily this allows the "PROTON CHARGE IS ZERO" case to be identified more successfully;

  * In all cases, substitute "PROTON CHARGE IS ZERO" for "BEAM IS DOWN".  Although in this case the beam may indeed be down, only the assertion about proton charge can be made accurately.

**Defects that have been fixed as part of this commit:**

  * The live-data toggle is no longer re-enabled after a normal reduction workflow, when the `application.yml` has `liveData.enabled: false`;

  * In live-data mode, in case of any error, reduction output workspaces are no longer deleted if they have valid content from a previous reduction cycle that _has_ completed successfully;

  * Several required fixes to what happens during an end-of-run transition.  This transition should now be working correctly, and displays a `warning` messagebox;

  * In the diffraction-calibration assessment panel: fix a defect that prevented the "load" function from working correctly.  This involved enabling the 'populateInstrumentParameters' method to successfully take a  workspace group as input;

  * In `GroceryService`: do not assume the diffraction-calibration mask is loaded just because the table is loaded. This caused errors in any successive reduction workflow, due to the _next_ defect;

  * When multiple workflows are being run _successfully_ add the diffraction-calibration mask to the "_keeps" list.  It was being added before, but was not using its correct workspace name;

  * Send the correct <start time> values to `LoadLiveData`.  These values are <EPOCH ZERO>, and <EPOCH + 1 second> which correspond to <now> and <all available data for the run> respectively.  The correct  behavior is that a _zero_ duration setting in the duration slider corresponds to <load all available data>, and due to this issue, this case was not being treated correctly;

**New DEFECTS found that have NOT been fixed:**

  * Race condition in Mantid `GetIPTS`;

  * Race conditions in Mantid's logging system;

  * The live-data listener does not shut down correctly when its calling algorithm has exited.  This is actually a high-priority defect as it has been associated with Mantidworkbench shut-down issues.  In this regard I've added all of the possible work arounds that I could think of.  However, it would be better to actually solve this problem within the Mantid code.  Almost certainly, this might require slight modifications to the live-listener interface, which at present is woefully inadequate for the requirements of the present usage;

  * The possibility of running `LoadLiveData` as a stay-resident algorithm was explored.  It was decided that although this functionality might be necessary in live-data "Accumulate" mode, it would actually generate erroneous output from `LoadLiveData` when run in `Replace` mode.  As a result of this, this stay-resident mode is not used and `Replace` mode is now the only mode that is supported.  (There's a note to this effect in `application.yml`, but there isn't an actual error check at initialization.  Solving the previously mentioned defect with the live listener should fix this issue.)

## To test
 
For "by hand" testing with the live listener, one needs to be patient, as there seem to be long periods where only very short runs occur, or only runs separated by long "PAUSED" intervals.  I have quite a few suggestions here, but I am in progress of drafting this section.

**In order to save time use `monitor.sns.gov` to determine the current \< _run_ _state_ \> for the SNAP instrument.  Also, it's important to modify your \< instrument parameters: appliesTo \> values so that the current run-numbers are included.**

When you run live-data sometimes there seem to be "drop outs", that is input-data workspaces that don't have their run number set correctly.  In this case SNAPRed will mistakenly identify this as an end-of-run condition, so you still need to keep track of what's actually going on at [monitor.sns.gov](http://monitor.sns.gov/).  In this case, just hit \<cancel\> then \<continue\> again, and things usually fix themselves.

### Dev testing
New unit tests have been implemented to cover most of the substantive changes.

## Link to EWM item

[EWM#9504](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=9504)

### DIFF
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -17,7 +17,7 @@ import h5py
 from mantid.api import Run
 from mantid.dataobjects import MaskWorkspace
 from mantid.kernel import ConfigService, PhysicalConstants
-from pydantic import ValidationError
+from pydantic import ValidationError, validate_call
 
 from snapred.backend.dao import (
     GSASParameters,
@@ -70,7 +70,7 @@ from snapred.meta.mantid.WorkspaceNameGenerator import (
 from snapred.meta.mantid.WorkspaceNameGenerator import (
     WorkspaceType as wngt,
 )
-from snapred.meta.redantic import parse_file_as, validate_call, write_model_pretty
+from snapred.meta.redantic import parse_file_as, write_model_pretty
 
 logger = snapredLogger.getLogger(__name__)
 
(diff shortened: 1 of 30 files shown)


src/snapred/backend/data/LocalDataService.py
```diff
@@ -1328,8 +1367,9 @@ def _readLiveData(self, ws: WorkspaceName, duration: int):
             "load live-data chunk",
             OutputWorkspace=ws,
             Instrument=Config["liveData.instrument.name"],
-            AccumulationMethod=Config["liveData.accumulationMethod"],
+            AccumulationMethod="Replace",
             StartTime=startTime,
+            PreserveEvents=False,
```
walshmm, line 1372: Should this be a parameter on the `_readLiveData` method itself?  I imagine this was done with reading metadata in mind.

walshmm (dismissed):

walshmm (approved):

PR #569, opened 2025-03-31
Flip east and west banks in side by side view

glass-ships:
For EWM Story [9534](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=9534)

This PR just flips the East and West detector panels in side-by-side view, following a meeting with Malcolm

### DIFF
--- a/src/snapred/resources/SNAPLite_Definition.xml
+++ b/src/snapred/resources/SNAPLite_Definition.xml
@@ -113,60 +113,60 @@
 
   <!-- East -->
   <type name="Column1">
-    <component type="panel"   idstart="2048" idfillbyfirst="y" idstepbyrow="32" >
-      <side-by-side-view-location x="-0.44" y="0.17"/>
+    <component type="panel"   idstart="131072" idfillbyfirst="y" idstepbyrow="256" >
+      <side-by-side-view-location x="0.1" y="0.17"/>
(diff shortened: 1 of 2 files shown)


darshdinger:
These changes look good. Before approving, I've got one question & one comment:

Question: Are there any testing parameters for this PR?

Comment: Insure that integration tests pass. You may need to re-run them if they fail due to this ongoing segmentation issue that Kort is still working on fixing. Re-running the test on the runners allows them to pass sometimes. If they continue to fail, please reach out to Daniel to see about adding a skip as I believe he may have had to do that in order to merge one of this previous PRs.

`codecov/project` will always fail if the `CI / integration-tests` fail as well. But both of these checks need to be met otherwise, we cannot merge unfortunately. :)

glass-ships:
> Are there any testing parameters for this PR?

None that I can think of, unless there's a way to text that mantid's instrument viewer successfully arranged the panels in a particular layout (I've not done any work with Mantid before so I genuinely don't know) 

> codecov/project will always fail if the CI / integration-tests fail as well. 

Only slightly confusing - I do see where codecov/project failed, but everything else looks like it passed (?)

darshdinger (approved):

PR #570, opened 2025-04-03
fix for the application of cropping factors to pgp's

walshmm:
## Description of work

**NOTE:** There is a potential defect/deviation between the workflows involving masking.  In `SousChef#L132`, set maskWorkspace to none for the sake of this test.  This should be covered in future work.  Otherwise you will likely observe differences due to masking.

This is a fix that enables snapred to uniformly apply its cropping factor and subsequently record the actual pgp values in each workflow.

Previously this was just applied in PixelGroup, which for some reason did not uniformly apply this to both the reduction workflow and normalization workflow.  Incidentally, because it was applied through a method call and wasnt flattened into the data it was not recorded on disk either.

## Explanation of work

I just moved where the cropping factor was applied.

## To test

1. Run reduction on a given run (e.g. 59039)
2. Run Normalization on the same run, with some other bg run (58810) and the column grouping, but stop at the tweak peak view
3. Run the cist test script included in the pr, comparing the log output and workspaces for each spectra in the referenced workspaces 

a favorable result should look something like this:
image

## Link to EWM item

[EWM#10466](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=10466)

### Verification

### Acceptance Criteria

- [ ] x values match for each spectra (cis test script)
- [ ] binnings match for each workspace (cis test script)

### DIFF
--- a/src/snapred/backend/dao/state/PixelGroup.py
+++ b/src/snapred/backend/dao/state/PixelGroup.py
@@ -82,7 +82,7 @@ class PixelGroup(BaseModel):
         #   different Mantid algorithms use either "0.0" or "NaN" to indicate an unspecified value,
         #   so these values may need to be filtered.
         return [
-            self.pixelGroupingParameters[gid].dResolution.maximum - Config["constants.CropFactors.highdSpacingCrop"]
+            self.pixelGroupingParameters[gid].dResolution.maximum
             if not self.pixelGroupingParameters[gid].isMasked
             else default
             for gid in self.groupIDs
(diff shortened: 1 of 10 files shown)


src/snapred/backend/dao/state/PixelGroup.py
darshdinger, line 1: See that you undid my changes. I guess I thought defining this within the PixelGroup object would allow it to be a global change and it would handle this but I guess that ended up not being true. (Observation comment, no changes needed.)

src/snapred/backend/recipe/algorithm/PixelGroupingParametersCalculationAlgorithm.py
darshdinger, line 1: This makes more sense.

darshdinger (commented):
Code looks good. Will test now.

darshdinger (approved):

PR #574, opened 2025-04-08
Fix integration tests and re-activate skipped tests.

ekapadi:
## Description of work

**This is the SNAPRed PR, corresponding to the `SaveNexusProcessed` SEGFAULT PR: [Mantid PR#39094](https://github.com/mantidproject/mantid/pull/39094).** 
In addition to reactivating skipped tests, this PR fixes several issues associated with the integration-tests and the test runners.

**Issues that must be addressed before this PR can be merged:**
  * the associated `mantid` PR needs to be merged, and a release needs to be available which includes it;
  * the proper `mantid` release version needs to be specified in `environment.yml`;
  * the tests which are marked `pytest.mark.skip` due to the `SaveReductionData` SEGFAULT need to be re-activated.
 
## Explanation of work

  * Minor fixup: Name the diffraction-calibration mask workspace so that the run number it uses is the same as that used by the corresponding table workspace.  (I.e. This will be the run number of the calibration run, not that of the input-data workspace.)  Initially it was thought that this naming was part of what was causing the integration tests to fail, but regardless, it is now fixed.

  * Remove 'Singleton' decorator from all 'Recipe'.  This caused problems in the integration-test runs because when a test failed, it triggered other test failures as the recipe's `MantidSnapper` instance would retain items in its algorithm queue.

  * Fix failing integration tests.  Some of these were not actually being run by the github runners.

  * Fix `.github/actions.yml` so that it properly _excludes_ tests for the "unit test step" and _includes_ all integration tests for the "integration test" step.

## To test
Issues addressed by this PR relate to the fixing of tests that were failing, but were not being run by the test runners.

## Link to EWM item

[EWM#10343](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=10343)

### Verification

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -9,7 +9,7 @@ dependencies:
 - python=3.10
 - pip
 - pydantic>=2.7.3,<3
-- mantidworkbench=6.12.0.2rc2 
+- mantidworkbench=6.12.0.2rc2
 - qtpy
 - pre-commit
 - pytest
--- a/environment.yml
+++ b/environment.yml
@@ -9,7 +9,7 @@ dependencies:
 - python=3.10
 - pip
 - pydantic>=2.7.3,<3
-- mantidworkbench=6.11.0.5
+- mantidworkbench=6.12.0.2rc2 
 - qtpy
 - pre-commit
 - pytest
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -36,7 +36,7 @@ jobs:
         /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16
     - name: Test with pytest
       run: |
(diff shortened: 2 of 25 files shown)


environment.yml
```diff
@@ -9,7 +9,7 @@ dependencies:
 - python=3.10
 - pip
 - pydantic>=2.7.3,<3
-- mantidworkbench=6.11.0.5
+- mantidworkbench=6.12.0.1
```
walshmm, line 12: should this be updated in `pyproject.toml#L7` as well?

ekapadi: Yes, but we still don't have the _actual_ tag yet.  This was just the tag I needed to get the tests to work "one before" the tag that we need to actually remove the `pytest.mark.skip` (on the tests that have a `writeReductionData` in them).

walshmm (approved):
Integration tests seem to be passing, and the new mantid seems to have actually built proper so I think I can approve this.

PR #575, opened 2025-04-10
remove sample and bg workspaces after RawVanadiumCorrection, do not r…

walshmm:
## Description of work

This is more work regarding the data management of sample data.
Largely repeated workflows with the same sample data dont seem to be a prime use case, and thus it is more beneficial to not cache them. 
This pr removes the sample data and background sample data cache for the normalization workflow.
In addition it also fixes the unnecessary recalculating of the RawVanadiumCorrectionAlgo and skips this on continue since it is already done for the preview.

## To test

Run the normalization workflow and observe the workspaces present at each stage as well as the algorithms executed with each action.
Confirm the AC.

## Link to EWM item

[EWM#10276](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=10276)

### Verification

### Acceptance Criteria

- [x] There are no `raw` and `copy` pairs of initial sample data workspaces
- [x] After the first step, there should only be one unfocussed even workspace `raw_van_corr`
- [x] RawVanadiumCorrection algo is only ever executed once in the entire workflow, even recalcs.
- [x] The final output is not to be converted to histograms at this time due to logistical difficulties in managing native vs lite cases

### DIFF
--- a/src/snapred/backend/service/NormalizationService.py
+++ b/src/snapred/backend/service/NormalizationService.py
@@ -121,6 +121,11 @@ class NormalizationService(Service):
             self.groceryClerk.name("backgroundWorkspace").neutron(request.backgroundRunNumber).useLiteMode(
                 request.useLiteMode
             ).dirty().add()
+        else:
+            # check that the corrected vanadium workspaces exist already
+            if not self.groceryService.workspaceDoesExist(correctedVanadium):
+                raise RuntimeError(f"Workspace {correctedVanadium} does not exist. Renew flag incorrectly applied.")
+
         self.groceryClerk.name("groupingWorkspace").fromRun(request.runNumber).grouping(groupingScheme).useLiteMode(
             request.useLiteMode
         ).add()
--- a/src/snapred/backend/service/NormalizationService.py
+++ b/src/snapred/backend/service/NormalizationService.py
@@ -114,7 +114,7 @@ class NormalizationService(Service):
             ).dict()
 
         # gather needed groceries and ingredients
-        if not request.renew:
+        if request.correctedVanadiumWs is None:
             self.groceryClerk.name("inputWorkspace").neutron(request.runNumber).useLiteMode(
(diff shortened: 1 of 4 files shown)


src/snapred/backend/service/NormalizationService.py
```diff
@@ -114,10 +114,13 @@ def normalization(self, request: NormalizationRequest):
             ).dict()
 
         # gather needed groceries and ingredients
-        self.groceryClerk.name("inputWorkspace").neutron(request.runNumber).useLiteMode(request.useLiteMode).add()
-        self.groceryClerk.name("backgroundWorkspace").neutron(request.backgroundRunNumber).useLiteMode(
-            request.useLiteMode
-        ).add()
+        if not request.renew:
```
ekapadi, line 117: This is a tricky issue:  if you aren't going to load them, since you're building a grocery list, I think at minimum you need to check and make sure these workspaces are already present in the ADS.  (I can imagine lots of "developer usage error" cases where this "if renew" bypass goes wrong.)

ekapadi (commented):
I had just the one comment so far.  I still need to check what's going on with the "don't recalc" stuff.  (And  then test everything.)

ekapadi:
Not meaning to "nit pick" about this PR:  it occurs to me that the sense of `NormalizationRequest.renew` is _inverted_.  The "use existing" is a _special_ _optimization_, and it probably should _not_ be the default behavior.  Is there any reason this couldn't default to `True`?  (Or possibly _rename_ it to something that means "don't recalculate"?)

ekapadi (approved):
I ran through the workflow and I can see that this is a major improvement with respect to how things function.  I approve these changes!

**One very minor issue:**  please add the "acceptance criteria" to the pane at the EWM item. -- Thanks!

PR #577, opened 2025-04-15
Fully-masked subgroups: quick fix

ekapadi:
## Description of work

There are several distinct ways of dealing with masking in Mantid. For example: (1) the mask flags may be set on the `DetectorInfo` (e.g. `MaskDetectorFlags`), (2) explicit mask workspaces may be used, (3) fully-masked spectra may be set to zero (e.g. `MaskDetectors`), and finally (4) fully-masked spectra may be omitted from a workspace.  In general SNAPRed uses a combination of (1) and (2), but does not either zero the spectra or omit fully-masked spectra.  The problem that the current PR solves is caused by the fact that `DiffractionFocussing`, in its default mode, excludes fully-masked spectra from its output workspace.

## Explanation of work

This PR includes the following changes:

  *  For the "quick fix", the `PixelGroupingParameters` calculation result is modified to omit PGP corresponding to fully-masked subgroups.  Since `PixelGroup` was already implemented as a `Mapping` using the subgroup ids as keys, this is only a minor modification.

  * As `PixelGroupingParameters` are already present in existing `CalibrationRecord` on SNS, to provide backwards compatibility the `isMasked` field is not removed, but is marked as `deprecated`.

  * `PixelGroup` is modified to exclude fully-masked `PixelGroupingParameters`.  With respect to its I/O, fully-masked subgroups are skipped at read, and are excluded at write.  (This does not affect the reading of any _existing_ `PixelGroup` json, but only its _re-writing_.)

  * In addition, the `PixelGroup` associated with a fully-masked grouping schema is modified so that its `timeOfFlight.binWidth` evaluates to `NaN`.  This allows the <detector peaks> peak-selection and purging algorithms to correctly return \<empty\> lists in this case.

  * With respect to \<detector peaks\> selection and purging:

    -- The masked `PixelGroup` is now used where appropriate to the calculation.  This allows the number of entries in the `List[GroupPeakList]` to correspond to the number of PGP in the `PixelGroup`.

    -- It's unfortunate that although `GroupPeakList` _essentially_ corresponds to a `dict` key-value pair, and `List[GroupPeakList]` corresponds to a `Dict[<subgroup id>, PeakList]`, they weren't implemented this way.  Changing them at this point would involve quite a bit of refactoring.  (Mostly due to backwards-compatibility issues.)

    -- `List[List[GroupPeakList]]` then corresponds to a `Dict[<focus group>, Dict[<subgroup id>, PeakList]]`, and has the same unfortunate refactoring requirement.

  * At an appropriate point in the reduction process, it is verified that the current order of the subgroup-IDs as presented by `PixelGroup`, corresponds to the order as presented by the workspace returned by `DiffractionFocussing`.  In this regard, it's noted that `DiffractionFocussing` sets the output-workspace spectrum numbers to correspond to the subgroup-IDs.

  * New warning messages have been added for both the fully-masked grouping schema case, and the case where there are fully-masked subgroups.  These two cases now work as expected with respect to the workflows.  The former skips any processing for a fully-masked grouping schema, and the latter simply provides a warning about the masked subgroups but continues processing.

  * `PixelGroupingParametersCalculationAlgorithm` has been renamed to `PixelGroupingParametersCalculation`.

  * The `EffectiveInstrumentRecipe` has been modified to work with the new `PixelGroup` format.  For the moment, spectra corresponding to fully-masked subgroups are excluded from the output workspace.  However, it seems likely that if this recipe is ever used to help build GSAS input, these spectra might need to be set to _zero_ instead.

  * A new cis-test script is implemented that allows easily verifying the new workflow behavior with both fully-masked subgroups, and with any completely masked grouping schema.

### Dev testing

By and large, existing unit tests cover all of these changes.  Where necessary, new _golden data_ was generated.

### CIS testing
A new cis-test script has been provided.

## Link to EWM item

[EWM#10364](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=10364)
**Please also see:** [EWM#8088](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=8088) for more of a description.

### Verification

- [x ] the author has read the EWM story and acceptance critera

### Acceptance Criteria

1. SNAPRed will be able to deal with fully-masked subgroups within any grouping schema.
2. SNAPRed warns the user (in the log) when any subgroup is fully masked.
3. SNAPRed will be able to deal with grouping schema in which all of the subgroups are fully masked.
4. SNAPRed warns the user (in the log) when any grouping schema only includes fully-masked subgroups.

- [x] acceptance criterion 1
- [x] acceptance criterion 2
- [x] acceptance criterion 3
- [x] acceptance criterion 4


walshmm:
# Additional Testing

Malcolm had generated a mask by hand `"/SNS/SNAP/shared/temp/Col1Col6Masked.h5"` which I then also used. Though this seems to be effectively the same as the cis test script, (Probably a good thing)

## Fully Masked Subgroups
image
it seems to work both in the example provided in pr and a real world example

## Fully Masked Groupings
I then modified the cis test script to generate a mask that completely masked everything for every grouping,
Reduction Completes with warning and no output "successfully" 
image
image
Should this be an exception?

Otherwise it seems to handle fully masked groupings by skipping and warning which is to be expected.

walshmm:
I managed to trigger a stacktrace using the cis test script.
first I run it masking Columns 1-5
```
    maskComponentByName(maskWs, "Column1")
    maskComponentByName(maskWs, "Column2")
    maskComponentByName(maskWs, "Column3")
    maskComponentByName(maskWs, "Column4")
    maskComponentByName(maskWs, "Column5")
```
Then I comment each of those out and run it again without clearing the workspace list
```
 Delta: Must specify for each spetra (2!=1)
 XMax: Must specify max for each spectra (2!=1)
 XMin: Must specify min for each spectra (2!=1)
  at line 135 in '/SNS/users/wqp/git/SNAPRed/src/snapred/backend/recipe/algorithm/FocusSpectraAlgorithm.py'
  caused by line 284 in '/SNS/users/wqp/git/SNAPRed/src/snapred/backend/recipe/algorithm/MantidSnapper.py'
  caused by line 254 in '/SNS/users/wqp/git/SNAPRed/src/snapred/backend/recipe/algorithm/MantidSnapper.py'
```
clearing the workspace list fixes this

walshmm:
The latest changes seemed to have fixed the previous error and the segfault I had experienced with the vanilla cis_test script.
I'll rereview the code, and approve after integration tests pass.

PR #578, opened 2025-04-21
Ewm10277 event managment in reduction

walshmm:
## Description of work

This is a collection of improvements identified by going over the reduction workflow with Malcolm.
Please refer to the acceptance criteria for a complete list.

## To test

The best way to test this is to run reduction on your favorite run in a fully calibrated and normalized state.
This way you can observe all the intended changes.
Some workspaces are now hidden and may require you use workbench to view the actual workspace list with mtd.
Others may require you to observe the code and execution (reduction in convertunits calls, or rebin calls)
You will also need one additional run that has a different vanadium associated with it to confirm that the cache is properly busted only when a new vanadium is needed.

## Link to EWM item

[EWM#10277](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=10277)

### Verification

### Acceptance Criteria

- [x] 1. modify groceryservice to check if rawvanadium is still in memory (and valid) use if it is
- [x] 2. grocery service should not clone sample, this isn't needed
- [x] 3. ApplyDifcal should be moved to groceryservice and out of preprocessingrecipe
- [x] 4. preprocessing on vanadium _should_ be applied to a clone.
- [x] 5. in reductionRecipe, loop over PGS, let diffraction focusing generate new output (keeping unfocused in ram, so no need to clone)
- [x] 6. possibly can reduce number of ConvertUnits if pre-processed converts to d (confirm this OK)
- [x] 7. At end of workflow don't delete original unmasked rawvanadium workspace (it can be used in next workflow)
- [x] 8. Make umasked rawvanadium hidden workspace
- [x] 9. Remove extra rebin from ApplyNormalization (data are already binned correctly)


src/snapred/backend/service/NormalizationService.py
```diff
@@ -137,16 +139,10 @@ def normalization(self, request: NormalizationRequest):
...
-            request.runNumber, request.useLiteMode
-        )
         calRunNumber = self.dataFactoryService.getCalibrationRecord(
             request.runNumber, request.useLiteMode, calVersion
         ).runNumber
 
-        self.groceryClerk.name("diffcalWorkspace").diffcal_table(request.runNumber, calVersion).useLiteMode(
-            request.useLiteMode
-        ).add()
         self.groceryClerk.name("maskWorkspace").diffcal_mask(request.runNumber, calVersion).useLiteMode(
```
ekapadi, line 146: This  makes me nervous!  I don't think that the mask should be treated separately like this.  If you load the diffraction-calibration table, you would also want to load the mask corresponding to that.  I don't believe that there are any exceptions to this.

Also, if this stays in, it should probably be using `calRunNumber`.  (If I remember correctly, it becomes `calRunNumber` in the end anyway.)   Unfortunately, there's still way too much "implicit lookup" of versions going on.

src/snapred/backend/service/ReductionService.py
```diff
@@ -428,21 +428,19 @@ def fetchReductionGroceries(self, request: ReductionRequest) -> Dict[str, Any]:
...
-        if calVersion is not None:
-            self.groceryClerk.name("diffcalWorkspace").diffcal_table(
-                request.runNumber, calVersion, alternativeState=request.alternativeState
-            ).useLiteMode(request.useLiteMode).add()
-
         if normVersion is not None:  # WARNING: version may be _zero_!
             self.groceryClerk.name("normalizationWorkspace").normalization(request.runNumber, normVersion).useLiteMode(
                 request.useLiteMode
-            ).add()
+            ).diffcalVersion(calVersion).dirty().add()
```
ekapadi, line 443: Somewhere around here you could add `.hidden()` (if it were a "builder" token).  :)

src/snapred/ui/workflow/ReductionWorkflow.py
```diff
@@ -480,6 +480,7 @@ def _triggerReduction(self, workflowPresenter: WorkflowPresenter):
         # Add loaded calibrations, calibration-masks, and normalizations to the list of workspaces to retain.
         self._keeps.update(loadedCalibrations)
         self._keeps.update(loadedNormalizations)
+        self.outputs.update(loadedNormalizations)
```
ekapadi, line 483: Would you please add a comment to this?  Something like: "normalizations are so expensive to load that we need to retain them as if they are 'outputs'"...   Otherwise, this is quite confusing!

src/snapred/ui/workflow/ReductionWorkflow.py
```diff
@@ -517,7 +518,7 @@ def _triggerReduction(self, workflowPresenter: WorkflowPresenter):
                 if response.code == ResponseCode.OK:
                     self._finalizeReduction(response.data.record, response.data.unfocusedData)
 
-                # after each run, clean workspaces except groupings, calibrations, normalizations, and outputs
+                # after each run, clean workspaces except groupings, calibrations, and outputs
```
ekapadi, line 521: I think I would have left the comment the way it was.  You _are_ retaining the normalizations, so the previous comment was correct.  AND, the normalizations are _NOT_ actually "outputs".

tests/unit/backend/recipe/test_ReductionGroupProcessingRecipe.py
```diff
@@ -71,6 +65,12 @@ def test_queueAlgos(self):
     def test_cook(self):
         untensils = Utensils()
         mockSnapper = unittest.mock.Mock()
+
+        mockSnapper.mtd = unittest.mock.MagicMock()
+        mockSnapper.mtd.__getitem__ = unittest.mock.Mock()
+        mockWorkspace = mockSnapper.mtd["input"]
+        mockWorkspace.getAxis().getUnit().unitID = unittest.mock.Mock(return_value="dSpacing")
```
ekapadi, line 72: Just one more reason to have a `_validateWSUnits` in that algorithm.

tests/unit/backend/recipe/test_ReductionGroupProcessingRecipe.py
```diff
@@ -95,6 +95,11 @@ def test_cook(self):
     def test_cater(self):
         untensils = Utensils()
         mockSnapper = unittest.mock.Mock()
+        mockSnapper.mtd = unittest.mock.MagicMock()
+        mockSnapper.mtd.__getitem__ = unittest.mock.Mock()
+        mockWorkspace = mockSnapper.mtd["input"]
+        mockWorkspace.getAxis().getUnit().unitID = unittest.mock.Mock(return_value="dSpacing")
```
ekapadi, line 101: ...  and again ...

tests/unit/backend/service/test_ReductionService.py
```diff
@@ -905,12 +899,11 @@ def mock_compatible_mask(wsname, runNumber, useLiteMode):  # noqa ARG001
...
-                runNumber, version, useLiteMode = item.runNumber, item.version, item.useLiteMode
+                runNumber, useLiteMode = item.runNumber, item.useLiteMode
+                version = item.diffcalVersion if item.diffcalVersion else item.normcalVersion
                 workspaceName = f"{runNumber}_{useLiteMode}_v{version}"
-                hasher = hashlib.shake_256()
-                hasher.update(json.dumps(item.__dict__).encode("utf-8"))
-                x = int.from_bytes(hasher.digest(1), "big")
-                mask = [int(x) for x in list("{0:0b}".format(x))]
+                x = 8
+                mask = [int(x) for x in range(x)]
```
ekapadi, line 906: **Sorry, but this is now even worse than it was before!**  I'll take a moment tomorrow and see if I can figure out what was supposed to be going on with this.  You have `maskFromArray([0, 1, 2, 3, 4...], workspaceName)` but it should be something like`maskFromArray([0, 0, 1, 0, 1, 1, 0, 0...], workspaceName)`.

tests/unit/backend/service/test_ReductionService.py
```diff
@@ -984,13 +977,12 @@ def trackFetchGroceryDict(*args, **kwargs):
 
         # construct the expected grocery dictionaries
         groceryClerk = self.service.groceryClerk
-        groceryClerk.name("inputWorkspace").neutron(request.runNumber).useLiteMode(request.useLiteMode).add()
-        groceryClerk.name("diffcalWorkspace").diffcal_table(
-            request.runNumber, request.versions.calibration
-        ).useLiteMode(request.useLiteMode).add()
+        groceryClerk.name("inputWorkspace").neutron(request.runNumber).useLiteMode(request.useLiteMode).diffcalVersion(
+            1
```
ekapadi, line 981: Shouldn't this be `request.versions.calibration`?

src/snapred/backend/data/GroceryService.py
```diff
@@ -1259,6 +1270,43 @@ def fetchMonitorWorkspace(self, item: GroceryListItem) -> WorkspaceName:
...
+
+        tableWorkspaceName, maskWorkspaceName = self._lookupDiffCalWorkspaceNames(
+            diffcalItem.runNumber,
+            diffcalItem.useLiteMode,
+            diffcalItem.diffcalVersion,
+            alternativeState=alternativeState,
+        )
+        if mask:
+            if maskWorkspaceName is not None:
+                result = self.fetchCalibrationWorkspaces(item)
```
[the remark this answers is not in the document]
walshmm: Thanks, I missed this one

src/snapred/backend/data/GroceryService.py
```diff
@@ -1133,8 +1135,8 @@ def fetchNormalizationWorkspace(self, item: GroceryListItem) -> Dict[str, Any]:
         :rtype: Dict[str, Any]
         """
 
-        runNumber, useLiteMode, version = item.runNumber, item.useLiteMode, item.version
-        workspaceName = self._createNormalizationWorkspaceName(runNumber, useLiteMode, version)
+        runNumber, useLiteMode, version = item.runNumber, item.useLiteMode, item.normcalVersion
+        workspaceName = self._createNormalizationWorkspaceName(runNumber, useLiteMode, version, hidden=True)
```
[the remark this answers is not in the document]
walshmm: > the runNumber used in the normalization workspace name is that associated with the normalization
That is correct, I can give this the same `normCalItem` treatment as diffCal and section off the record runnumber lookup

src/snapred/backend/data/GroceryService.py
```diff
@@ -1299,40 +1347,23 @@ def fetchGroceryList(self, groceryList: Iterable[GroceryListItem]) -> List[Works
...
-                    # NOTE: fetchCalibrationWorkspaces will set the workspace name
-                    # to that of the table workspace.  Because of possible confusion with
-                    # the behavior of the mask workspace, the workspace name is overridden here.
-                    tableWorkspaceName, _ = self._lookupDiffCalWorkspaceNames(
-                        item.runNumber, item.useLiteMode, item.version, alternativeState=alternativeState
-                    )
-                    result = self.fetchCalibrationWorkspaces(item)
-                    result["workspace"] = tableWorkspaceName
+                    result = self.fetchDiffcalForSample(item, mask=False)
                 case "diffcal_mask":
```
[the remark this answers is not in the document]
walshmm: I can update the name, no problem.

src/snapred/backend/data/GroceryService.py
```diff
@@ -1533,3 +1566,16 @@ def _processNeutronDataCopy(self, item: GroceryListItem, workspaceName):
             self.writeWorkspaceMetadataAsTags(workspaceName, metadata)
 
             self.deleteWorkspaceUnconditional(monitorWs)
+
+        if item.diffcalVersion is not None:
+            # then load a diffcal table and apply it.
+            # NOTE: This can result in a different diffcal being applied to normalization vs sample
+            #       This is the expected and desired behavior.
+
+            diffcalWorkspace = self.fetchDiffcalForSample(item, mask=False)["workspace"]
```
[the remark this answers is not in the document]
walshmm: I think that makes sense, I'll update it accordingly.

src/snapred/ui/workflow/ReductionWorkflow.py
walshmm: I think I had misread the comment

tests/unit/backend/recipe/test_ReductionGroupProcessingRecipe.py
walshmm: I see, probably easier to mock this way.

src/snapred/backend/service/NormalizationService.py
ekapadi: 25.04.2025:  The problem here is that I don't want the calibration run number and version to be established in two separate places.  Right now, you figure out `calRunNumber`, but then don't use that for the mask load.  (If I'm remembering correctly _within_ the mask load it will do exactly the same lookup, but that's irrelevant, as almost certainly one of those lookups should be removed.)  **This is why using the `calRunNumber` here for the mask is important.**

src/snapred/backend/recipe/algorithm/FocusSpectraAlgorithm.py
```diff
@@ -112,17 +117,17 @@ def PyExec(self):
...
         # convert to d-spacing and diffraction focus and rebin ragged
-        self.mantidSnapper.ConvertUnits(
-            "Converting to Units of dSpacing...",
-            InputWorkspace=self.inputWSName,
-            Emode="Elastic",
-            Target="dSpacing",
-            OutputWorkspace=self.outputWSName,
-            ConvertFromPointData=True,
-        )
+        # self.mantidSnapper.ConvertUnits(
```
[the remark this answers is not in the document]
ekapadi: 25.04.2025:  Still needs a comment...

src/snapred/backend/recipe/GenericRecipe.py
```diff
@@ -99,6 +99,10 @@ class FocusSpectraRecipe(GenericRecipe[FocusSpectraAlgorithm]):
     pass
 
 
+class ConvertUnitsRecipe(GenericRecipe[ConvertUnits]):
+    pass
```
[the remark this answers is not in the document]
walshmm: In order avoid cloning before `DiffractionFocussing`, ConvertUnits needed to be moved out of the FocusSpectraAlgorithm.

The recipe allows service layer classes the ability to more  cognizantly prep its data before submitting it to other recipes instead of blindly converting and potentially generating a clone because of it.

src/snapred/backend/recipe/ReductionRecipe.py
```diff
@@ -253,34 +249,29 @@ def execute(self):
...
-            )
-            self._cloneIntermediateWorkspace(self.normalizationWs, "normalization_preprocessed")
 
         if bool(self.maskWs) and all(
             (self._isGroupFullyMasked(groupingIndex) for groupingIndex in range(len(self.groupingWorkspaces)))
         ):
             raise RuntimeError(
                 "There are no unmasked pixels in any of the groupings.  Please check your mask workspace!"
             )
+        self.mantidSnapper.ConvertUnits(
```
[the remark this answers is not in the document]
walshmm: Im not sure what is being asked, but the recipe wrapper is to keep the architecture pattern in line.
Since this is a recipe itself, its free to use the actual algorithm.

src/snapred/backend/service/NormalizationService.py
walshmm: `calRunNumber` is not used to look up the appropriate calibration file though. It is a key of `(sampleRunNumber, version` which both the sample and mask `GroceryItem` share.  I forget where calRunNumber is used in the `NormalizationWorkflow`  

This is difficult because either we explicitly ask for a mask (as above)
Or we implicitly load the mask when the load the appropriate diffcal, and it just ends up in the groceries list?

walshmm: a `calRunNumber` could be `59039` which when search on the index may find a calibration not applicable to the actual sample of say `58810`

src/snapred/backend/recipe/ReductionGroupProcessingRecipe.py
```diff
@@ -32,28 +32,41 @@ def chopIngredients(self, ingredients):
         logger.debug(f"dMax: {self.pixelGroup.dMax()}")
         logger.debug(f"dBin: {self.pixelGroup.dBin()}")
 
+    def _validateWSUnits(self, key, ws):
+        if key == "inputWorkspace":
+            # assert that the input workspace is in TOF
```
ekapadi, line 37: This comment should say "dspacing" not "TOF".

src/snapred/backend/recipe/algorithm/FocusSpectraAlgorithm.py
```diff
@@ -81,6 +82,10 @@ def validateInputs(self) -> Dict[str, str]:
         errors = {}
         # make sure the input workspace can be reduced by this grouping workspace
         inWS = self.mantidSnapper.mtd[self.getPropertyValue("InputWorkspace")]
+
+        if not FocusSpectraAlgorithm.unitsMatch(inWS, "dSpacing"):
```
ekapadi, line 86: I'm confused:   doesn't the `validator=WorkspaceUnitValidator("dSpacing")` in the arg-property definitions already check this?

src/snapred/backend/recipe/algorithm/MantidSnapper.py
```diff
@@ -22,7 +22,9 @@ class _CustomMtd:
     def __getitem__(self, key):
         if str(key.__class__) == str(callback(int).__class__):
             key = key.get()
-        return mtd[key]
+        if self.doesExist(key):
+            return mtd[key]
+        raise KeyError(f"Workspace {key} not found in mtd: {self.getObjectNames()}")
```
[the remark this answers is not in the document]
ekapadi: I'm fine with that, but personally  would switch it based on the logging level.  However, _you_ don't need to worry about that for this PR -- there's a "reduce logging spam" enabler coming up!  :)

src/snapred/backend/service/NormalizationService.py
ekapadi: The implicit load of the mask doesn't _automatically_ end up in the groceries list.  It would only be in the ADS, and only added to the grocery list if someone did that _explicitly_ .  For example:  what you're doing in this method:  the `buildGroceryList` will load the calibration-table first, then when it gets to the mask, it will find it in the ADS _and_ add it to the list.   (Note:  I still don't like the fact that there's an additional `getLatestApplicableVersion` fallback in the `GroceryService` methods.)

ekapadi (commented):
Only one very minor comment about "dSpacing" vs. "TOF" in a comment, and possibly redundant units checking (when the algorithm had a `WorkspaceUnitsValidator` attached)?

I fully checked all of the cached normalization cases, including the case where there are two runs from the same state, but which have different normalizations (as per the `<indexer>.appliesTo`).  Everything works as it should.  

Take a look at my comment, and I'm ready to approve these changes.  I hope we can add `PreprocessReductionRecipe` back in, though.

src/snapred/backend/recipe/algorithm/FocusSpectraAlgorithm.py
walshmm: I think that was a mistep in a rebase I had refactored according to a comment.  It is definitely redundant!

ekapadi (approved):
I checked over it one more time, and everything still looks OK.  I approve these changes.

PR #582, opened 2025-04-29
Ewm2990 reload application yml

walshmm:
## Description of work

The goal of this pr is to enable snapred's Config to be re-loadable at runtime.

This is mostly accomplished in this pr with a couple minor exceptions.

## Explanation of work

Essentially I had to replace each flattened call to Config (i.e. saved as a prop, default method param, default member value) to a generator so that the latest value is always pulled.

This ranged from adding Field default_generators, to creating a decorator that inspects args and replaces defaults.

## To test

Refer to the cis test script, also use the new button in the main gui.
You can actually turn off the button in the cis_mode props, reload, close and reopen the main gui to make that button disappear.

## Link to EWM item

[EWM#2990](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=2990)

### Verification

### Acceptance Criteria

- [x] add a cis_mode flag to enable a `Reload Configuration` button
- [x] the reload button provides a positive message informing the user of the env they reloaded.
- [x] Reloading the configuration is effective across the codebase

### DIFF
--- a/src/snapred/backend/dao/indexing/Versioning.py
+++ b/src/snapred/backend/dao/indexing/Versioning.py
@@ -1,16 +1,15 @@
-from enum import Enum
 import numpy
 from pydantic import BaseModel, ConfigDict, field_validator
 
 from snapred.meta.Config import Config
 from snapred.meta.Enum import StrEnum
 
-# TODO: Wrap in class so that default values may be overridden  
+# NOTE: This should probably not be reconfigurable at runtime.
+#       This would be liable to only cause indexing issues.
 VERSION_START = Config["version.start"]
 
 
 class VersionState(StrEnum):
-    START = Config["version.start"]
     DEFAULT = Config["version.friendlyName.default"]
     LATEST = "latest"
     NEXT = "next"
--- a/src/snapred/backend/dao/indexing/Versioning.py
+++ b/src/snapred/backend/dao/indexing/Versioning.py
@@ -1,14 +1,16 @@
+from enum import Enum
 import numpy
 from pydantic import BaseModel, ConfigDict, field_validator
 
 from snapred.meta.Config import Config
 from snapred.meta.Enum import StrEnum
 
+# TODO: Wrap in class so that default values may be overridden  
 VERSION_START = Config["version.start"]
-VERSION_NONE_NAME = Config["version.friendlyName.error"]
 
 
 class VersionState(StrEnum):
+    START = Config["version.start"]
     DEFAULT = Config["version.friendlyName.default"]
     LATEST = "latest"
     NEXT = "next"
--- a/src/snapred/backend/dao/indexing/Versioning.py
+++ b/src/snapred/backend/dao/indexing/Versioning.py
@@ -46,8 +46,8 @@ class VersionedObject(BaseModel):
             # Conversion from HDF5 metadata.
             value = int(value)
 
-        if value < VERSION_START:
-            raise ValueError(f"Version must be greater than {VERSION_START}")
+        if value < VERSION_START():
+            raise ValueError(f"Version must be greater than {VERSION_START()}")
 
         return value
 
--- a/src/snapred/backend/dao/indexing/Versioning.py
+++ b/src/snapred/backend/dao/indexing/Versioning.py
@@ -4,9 +4,11 @@ from pydantic import BaseModel, ConfigDict, field_validator
 from snapred.meta.Config import Config
 from snapred.meta.Enum import StrEnum
 
+
 # NOTE: This should probably not be reconfigurable at runtime.
 #       This would be liable to only cause indexing issues.
-VERSION_START = Config["version.start"]
+def VERSION_START():
+    return Config["version.start"]
 
 
 class VersionState(StrEnum):
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -22,6 +22,7 @@ from snapred.backend.recipe.algorithm.MantidSnapper import MantidSnapper
 from snapred.backend.recipe.FetchGroceriesRecipe import FetchGroceriesRecipe
 from snapred.backend.service.WorkspaceMetadataService import WorkspaceMetadataService
 from snapred.meta.Config import Config
+from snapred.meta.decorators.ConfigDefault import ConfigDefault, ConfigValue
 from snapred.meta.decorators.Singleton import Singleton
 from snapred.meta.InternalConstants import ReservedRunNumber
 from snapred.meta.mantid.WorkspaceNameGenerator import (
@@ -185,8 +186,8 @@ class GroceryService:
                 del self._loadedInstruments[key]
 
     ## FILENAME METHODS
-
-    def getIPTS(self, runNumber: str, instrumentName: str = Config["instrument.name"]) -> str:
+    @ConfigDefault
+    def getIPTS(self, runNumber: str, instrumentName: str = ConfigValue("instrument.name")) -> str:
         """
         Find the approprate IPTS folder for a run number.
 
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -359,7 +359,7 @@ class GroceryService:
         NOTE: This method will IGNORE runNumber if the provided version is VersionState.DEFAULT
         """
         wsName = wng.diffCalTable().runNumber(runNumber).version(version).build()
-        if version in [VersionState.DEFAULT, VERSION_START]:
+        if version in [VersionState.DEFAULT, VERSION_START()]:
             wsName = wng.diffCalTable().runNumber("default").version(VersionState.DEFAULT).build()
         return wsName
 
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -92,6 +92,7 @@ class LocalDataService:
     # conversion factor from microsecond/Angstrom to meters
     # (TODO: FIX THIS COMMENT! Obviously `m2cm` doesn't convert from 1.0 / Angstrom to 1.0 / meters.)
     CONVERSION_FACTOR = Config["constants.m2cm"] * PhysicalConstants.h / PhysicalConstants.NeutronMass
+    _verifyPaths = None
 
     def __init__(self) -> None:
         self.mantidSnapper = MantidSnapper(None, "Utensils")
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -55,6 +55,7 @@ from snapred.backend.error.StateValidationException import StateValidationExcept
 from snapred.backend.log.logger import snapredLogger
 from snapred.backend.recipe.algorithm.MantidSnapper import MantidSnapper
 from snapred.meta.Config import Config
+from snapred.meta.decorators.ConfigDefault import ConfigDefault, ConfigValue
 from snapred.meta.decorators.ExceptionHandler import ExceptionHandler
 from snapred.meta.decorators.Singleton import Singleton
 from snapred.meta.InternalConstants import ReservedRunNumber, ReservedStateId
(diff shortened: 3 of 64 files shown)


src/snapred/backend/dao/indexing/Versioning.py
```diff
@@ -4,8 +4,9 @@
 from snapred.meta.Config import Config
 from snapred.meta.Enum import StrEnum
 
+# NOTE: This should probably not be reconfigurable at runtime.
+#       This would be liable to only cause indexing issues.
```
ekapadi, line 8: I'm wondering here if you should instead _allow_ them to change it, but give a really _nasty_ log warning about what a bad idea it is?!

src/snapred/backend/data/GroceryService.py
```diff
@@ -185,8 +186,8 @@ def _updateInstrumentCacheFromADS(self, runNumber: str, useLiteMode: bool):
                 del self._loadedInstruments[key]
 
     ## FILENAME METHODS
-
-    def getIPTS(self, runNumber: str, instrumentName: str = Config["instrument.name"]) -> str:
+    @ConfigDefault
+    def getIPTS(self, runNumber: str, instrumentName: str = ConfigValue("instrument.name")) -> str:
```
ekapadi, line 190: You will see more on this later (in my comment at the decorator).  I think it's _amazing_ that the decorator works.   However, I'm really wondering if it wouldn't have been more robust to do something instead like just move these to `classproperty`?  :)

src/snapred/backend/data/LocalDataService.py
```diff
@@ -189,7 +194,8 @@ def getUniqueTimestamp() -> float:
         return nextTimestamp
 
     @lru_cache
-    def _getIPTS(self, runNumber: str, instrumentName: str = Config["instrument.name"]) -> str | None:
+    @ConfigDefault
+    def _getIPTS(self, runNumber: str, instrumentName: str = ConfigValue("instrument.name")) -> str | None:
```
ekapadi, line 198: I'm wondering whether or not this messes with `lru_cache`?  Not in terms of the cached values, but  I guess if its the "inner" decorator there's no issue?  Let me know what you think!

src/snapred/backend/recipe/algorithm/WashDishes.py
```diff
@@ -20,8 +20,19 @@ def PyInit(self):
         self.declareProperty(StringArrayProperty(name="WorkspaceList", values=[], direction=Direction.Input))
         self.setRethrows(True)
 
-        self.cis_enabled: bool = Config["cis_mode.enabled"]
-        self.cis_preserve: bool = Config["cis_mode.preserveDiagnosticWorkspaces"]
+    @property
+    def cis_enabled(self) -> bool:
+        """
+        Check if CIS mode is enabled
+        """
```
ekapadi, line 27: It seemed to me that quite possibly, _all_ of these `@property` should actually be `@classproperty` (which you only introduce later).  Take a look over these, and see what you think?

src/snapred/backend/recipe/algorithm/WashDishes.py
```diff
@@ -20,8 +20,19 @@ def PyInit(self):
...
+        return Config["cis_mode.enabled"]
+
+    @property
+    def cis_preserve(self) -> bool:
+        """
+        Check if CIS mode is enabled
```
ekapadi, line 33: Comment should read "Check if diagnostic workspaces should be retained" or something like that.

src/snapred/backend/service/CrystallographicInfoService.py
```diff
@@ -22,7 +30,9 @@ def name():
         return "ingestion"
 
     @FromString
-    def ingest(self, cifPath: str, crystalDMin: float = D_MIN, crystalDMax: float = D_MAX) -> Dict[Any, Any]:
+    def ingest(
+        self, cifPath: str, crystalDMin: float = _Properties.D_MIN, crystalDMax: float = _Properties.D_MAX
```
ekapadi, line 34: **AND, here is an example of NOT using `@ConfigDefault` + `@ConfigValue`.**  The issue that I have is that that decorator is _REALLY_ complicated!  :(

src/snapred/backend/service/Service.py
```diff
@@ -31,12 +31,14 @@ def reg(func):
...
 class Service(ABC):
-    _pathDelimiter = Config["orchestration.path.delimiter"]
-
     def __init__(self):
         self._paths: Dict[str, Any] = self._getInstancePaths()
         self._lambdas: Dict[str, List[GroupingLambda]] = {}
 
+    @property
+    def _pathDelimiter(self):
+        return Config["orchestration.path.delimiter"]
```
ekapadi, line 40: `_pathDelimiter()` ends up being declared in three different places?

src/snapred/backend/service/ServiceDirectory.py
```diff
@@ -8,7 +8,10 @@
 @Singleton
 class ServiceDirectory:
     _services: Dict[str, Any] = {}
-    _pathDelimiter = Config["orchestration.path.delimiter"]
+
+    @property
```
ekapadi, line 12: ... declared again _here_ ...

src/snapred/backend/service/ServiceFactory.py
```diff
@@ -38,6 +37,10 @@ def __init__(self):
         self.serviceDirectory.registerService(WorkspaceService)
         self.serviceDirectory.registerService(WorkspaceMetadataService)
 
+    @property
+    def _pathDelimiter(self):
+        return Config["orchestration.path.delimiter"]
```
ekapadi, line 42: ... and again _here_ ...

src/snapred/meta/decorators/ConfigDefault.py
```diff
@@ -0,0 +1,52 @@
...
+class ConfigValue:
+    def __init__(self, value: Any):
+        self.value = value
+
+    def get(self):
+        return Config[self.value]
+
+
+def ConfigDefault(func: Callable[..., Any]):
+    @functools.wraps(func)
```
ekapadi, line 17: Comments on this decorator are as mentioned previously.  This is a really cool implementation!  **However, is it too complex?**  I'm also concerned that we're doing this _every_ time the functions are called.

src/snapred/meta/decorators/classproperty.py
```diff
@@ -0,0 +1,6 @@
+# https://stackoverflow.com/questions/1697501/staticmethod-with-property
+
+
+class classproperty(property):
```
ekapadi, line 4: _Possibly_ should be used for most of the `@property` conversions as well.   (Note: I don't care one way or another, but if we pass the `self` and don't use it in the method, it's likely to be a `@staticmethod`  (e.g. or almost equivalently, an `@classmethod`).

src/snapred/ui/main.py
```diff
@@ -90,6 +86,21 @@ def __init__(self, parent=None, window_flags=None, translucentBackground=False):
 
         splitter.addWidget(AlgorithmProgressWidget())
 
+        if Config["cis_mode.reloadConfigButton"]:
+            self.reloadConfigButton = QPushButton("Reload Config")
+
+            def reloadAndInform():
+                Config.reload()
+                # Inform the user that the configuration has been reloaded
```
ekapadi, line 94: **And what happens if the reload _fails_?**  I'm wondering if when SNAPRed loads, it should save a copy of a successfully-loaded `application-environment.yml` to use for recovery?

src/snapred/ui/view/DiffCalTweakPeakView.py
```diff
@@ -35,15 +35,16 @@ class DiffCalTweakPeakView(BackendRequestView):
 
     """
 
+    FIGURE_MARGIN = 0.5  # top + bottom: inches
+
+    # Defaults, NOTE: the gui element will need to be closed for new defaults to take effect
```
ekapadi, line 40: Why aren't these split out to `@property` like in `NormalizationTweakPeakView`?  What's the difference?  This also suggests to me that there needs to be a `QMessageBox` pop up at the reload that includes the text **"You may need to close and re-open the SNAPRed calibration panel for some of these changes to take effect."**

ekapadi (commented):
So far, I've made a pass over everything _except_ the new test code.  I have some minor concerns that the new decorator is a little-bit too complex, but then I'm not overly worried about that!

There seems some inconsistency with the use of `@property` and your new `@classproperty` -- it looks to me as if the latter should be used almost everywhere.  Also the same is true of the `_Properties` class treatment vs. use of the new decorator -- this seems a good example of an _alternative_ way to do this, but why didn't you just use the decorator there as well?

I'll run this through its tests tomorrow, and also make a pass over the tests code.

walshmm:
> this seems a good example of an _alternative_ way to do this, but why didn't you just use the decorator there as well?

This was actually a mistake, looking at it I dont think reload would have actually worked for that.  It would probably evaluate the property immediately and cache it.

src/snapred/backend/data/GroceryService.py
walshmm: As we discussed after this comment was made, and just duplicating it here for records sake:  it would require the developer to handle defaults manually in each of these cases.

src/snapred/backend/data/LocalDataService.py
walshmm: During execution they are executed from top to bottom, so the `lru_cache` should sidestep `ConfigDefault`, saving some work.  Either way, the cache should still save us the getIPTS execution time.

src/snapred/ui/view/DiffCalTweakPeakView.py
walshmm: the difference was that these are not constants.  Though thinking about this again a week later I can probably move these into the init and have those refer to some default method?

src/snapred/ui/main.py
```diff
@@ -90,13 +91,19 @@ def __init__(self, parent=None, window_flags=None, translucentBackground=False):
...
+                    # Inform the user that the configuration has been reloaded
+                    QMessageBox.information(
+                        self,
+                        "Configuration Reloaded",
+                        f"Env {Config.getCurrentEnv()} configuration has been successfully reloaded.",
+                    )
+                except Exception:  # noqa: BLE001
+                    # Additionally, show error message as popup
+                    msg = "Error encountered while reloading the configuration.\n"
+                    msg = msg + "Please ensure the `.yml` file is properly formatted and available\n"
```
ekapadi, line 105: **I still think we might need to _write_ their previous-good configuration to a safe place (and then mention it in this message)?**  The case I'm thinking about is if they completely "hose" their `application.yml` and then can't figure out how to fix it.  The fact that SNAPRed is _running_ with an OK configuration isn't going to help them much in that case.

One way of dealing with this would be in `Config`, after a successful load, just save a copy of the file to `~/.SNAPRed/application_yml.bak`.  (This is kind-of what Mantid does with `mantid-user.properties`.

tests/unit/backend/dao/test_VersionedObject.py
```diff
@@ -197,10 +197,10 @@ def test_write_version_index_entry():
...
         vo.model_dump_json()
 
-    vo.version = VERSION_START
-    assert f'"version":{VERSION_START}' in vo.model_dump_json()
-    assert f'"version":"{VERSION_START}"' not in vo.model_dump_json()
-    assert vo.dict()["version"] == VERSION_START
+    vo.version = VERSION_START()
+    assert f'"version":{VERSION_START()}' in vo.model_dump_json()
+    assert f'"version":"{VERSION_START()}"' not in vo.model_dump_json()
+    assert vo.dict()["version"] == VERSION_START()
```
ekapadi, line 203: deprecated `dict`

tests/unit/backend/dao/test_VersionedObject.py
```diff
@@ -274,8 +274,8 @@ def test_write_version_record():
...
         vo.model_dump_json()
-    vo.version = VERSION_START
+    vo.version = VERSION_START()
 
-    assert f'"version":{VERSION_START}' in vo.model_dump_json()
-    assert f'"version":"{VERSION_START}"' not in vo.model_dump_json()
-    assert vo.dict()["version"] == VERSION_START
+    assert f'"version":{VERSION_START()}' in vo.model_dump_json()
+    assert f'"version":"{VERSION_START()}"' not in vo.model_dump_json()
+    assert vo.dict()["version"] == VERSION_START()
```
ekapadi, line 281: ... deprecated `dict` ...

ekapadi (approved):
I'm noting RE "acceptance criteria" that instead of adding an entire new panel, we just get the "config reload" button.  However, I agree that that is what we need to do at this point!
I ran this branch through the workflows, and evertying works as expected.  

I attempted Config reloads with various "legal yml with bad values" and "illegal yml".  The former is still accepted because we don't have a typing-schema for the `application.yml` --  but that isn't part of this story.  

The latter clearly fails the reload, with a warning messagebox, but does not crash SNAPRed.  In all cases of a successful Config load, the "known good" file is now backed up (and the backup location is indicated in the logs).  This makes recovering even from the former case straightforward.    **I approve these changes.**

PR #595, opened 2025-05-19
Ewm11197 calibration lookup by state

walshmm:
## Description of work

This is intended to fix the way snapred reads calibration/normalization workspaces such that users do not require read perms to anything besides the reduction run and the calibration root.

## Explanation of work

Right now snapred requires read permissions to the original files used to generate the Calibration and Normalization files.
This is because it uses their logs to generate the state id we use for lookup, but this is redundant as we can generate that from the target sample run.
In addition, we already supply "version" as part of Calibration and Normalization lookup, which means at Service level we already have access to this info.

Thus, this pr refactors the calibration and normalization lookups to remove the optional `alternativeState`, and instead require `state`.

## To test

This requires a regression test of each workflow, and most of all reduction.
It may require that you inspect/insert logs to ensure no run data is required to load calibration/normalizations.

## Link to EWM item

[EWM#11197](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=11197)

### Verification

### Acceptance Criteria

- [ ] snapred no longer requires the inspection of calibration/normalization sample workspaces (i.e. the original ones stored in the ipts that may not have read perms for everyone) as part of reduction,

### DIFF
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -236,7 +236,7 @@ class GroceryService:
     def _createDiffCalOutputWorkspaceFilename(self, item: GroceryListItem) -> str:
         ext = Config["calibration.diffraction.output.extension"]
         return str(
-            self._getCalibrationDataPath(item.runNumber, item.useLiteMode, item.diffCalVersion)
+            self._getCalibrationDataPath(item.useLiteMode, item.diffCalVersion, item.state)
             / (self._createDiffCalOutputWorkspaceName(item) + ext)
         )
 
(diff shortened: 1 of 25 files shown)


src/snapred/backend/data/GroceryService.py
```diff
@@ -393,14 +420,25 @@ def _createDiffCalMaskWorkspaceName(
     ) -> WorkspaceName:
         return wng.diffCalMask().runNumber(runNumber).version(version).build()
 
+    def _lookupNormcalRunNumber(self, sampleRunNumber: str, useLiteMode: bool, version: int, state: str):
+        # runNumber = self._lookupNormcalRunNumber(<run number>, <use lite mode> [, ...])
```
dlcaballero16, line 424: remove commented line

tests/unit/backend/service/test_ReductionService.py
```diff
@@ -1021,6 +1038,8 @@ def test_fetchReductionGroceries_load(self):
         However, it was already here, for simplicity of review I am not moving it.
         """
 
+        # self.service.dataFactoryService.constructStateId = mock.Mock(return_value=(self.stateId1, None))
```
dlcaballero16, line 1041: remove commented line

dlcaballero16 (approved):
Tested and verified that calibration and normalization are now both looked up by state

PR #610, opened 2025-06-23
configure deploy build based on version

walshmm:
### CIS testing

``` python
```

## Link to EWM item

### Verification

### Acceptance Criteria

- [ ] acceptance criterion 1
- [ ] acceptance criterion 2

### DIFF
--- a/src/snapred/meta/Config.py
+++ b/src/snapred/meta/Config.py
@@ -14,6 +14,13 @@ import yaml
 
 from snapred import __version__ as snapredVersion
 from snapred.meta.decorators.Singleton import Singleton
+from snapred.meta.Enum import StrEnum
+
+
+class DeployEnvEnum(StrEnum):
+    NEXT = "snapred_next"
+    QA = "snapred_qa"
+    PROD = "snapred_prod"
 
 
 def isTestEnv() -> bool:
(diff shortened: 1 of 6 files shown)


src/snapred/meta/Config.py
```diff
@@ -136,6 +143,26 @@ def expandhome(direc: str) -> str:
...
+
+    def mergeAndExport(self, envName: str) -> None:
+        """
+        Merge the current configuration with the specified environment configuration
+        and export it to the application.yml file.
+        """
+        self._logger.debug(f"Merging/exporting configuration with environment: {envName}")
+        self.refresh(envName, False)
+        # Export the merged configuration to application.yml
+        with Resource.open("application.yml", "w") as file:
```
ekapadi, line 163: The thing that really bugs me about this is that it's going to strip out any useful comments that are present in the \<source\> YML.  I think it might be a better idea to `{replaceTag}` sections from a \<source YML\> using string formatting.  **Hopefully it's clear what I mean here!**

src/snapred/meta/Config.py
```diff
@@ -136,6 +143,26 @@ def expandhome(direc: str) -> str:
         if "samples" in self._config and "home" in self._config["samples"]:
             self._config["samples"]["home"] = expandhome(self._config["samples"]["home"])
 
+    def configureForDeploy(self) -> None:
```
ekapadi, line 146: I'm wondering if these should be in a _script_, which is located somewhere else (e.g. `utils/deploy/generate_application_yml.py`)?

ekapadi (commented):
Two issues:

1. A huge advantage of YAML (over the JSON variant of YAML), is that it can include comments.  Regenerating the yaml using `yaml.dump` is definitely something that's needed, as it _verifies_ that the produced YAML file is valid, however I think that you need to find a way to _preserve_ the incoming comments from the \<source\> file.  I'm suggesting actually using string-format operations, with `{replaceTag}`-style entries in the source-file strings.  (And then as a separate step:  verify that the new file is actually valid, but then output the file that still has the _comments_ in it!)
2. Please add a "step by step" about how to build the conda package locally?  I've never managed to figure out how to do that.

src/snapred/meta/Config.py
walshmm: Id rather keep this stuff in one place and avoid one off scripts unless the action is really one off or outside snapred's normal jurisdiction

ekapadi (approved):
Ran "pixi run build-conda" and examined the as-produced `application.yml`, which is as it should be.

Apart from the fact that our "deploy build chain" is not "out of source" -- it modifies files in the repositories working directories, I'm satisfied that that is a _separate_ issue, and outside of the scope of this work item.  

All of the changes behave as expected, and I approve these changes.

PR #612, opened 2025-06-25
EWM 11979 CIS Hooks 

walshmm:
## Description of work

This pr adds an implementation that enables external api users to insert hooks into the established snapred workflow to help facilitate prototyping of future features.

## To test

There is a cis test included that outlines how to use the new hooks.
NOTE: this cis test also includes an extra hook to trigger validation checks.

## Link to EWM item

[EWM#11979](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=11979)

### Verification

### Acceptance Criteria

- [ ] acceptance criterion 1
- [ ] acceptance criterion 2

### DIFF
--- a/src/snapred/backend/api/HookManager.py
+++ b/src/snapred/backend/api/HookManager.py
@@ -10,7 +10,7 @@ logger = snapredLogger.getLogger(__name__)
 @Singleton
 class HookManager:
     """
-    Registers, manages, and executes Callables though a given request.
+    Registers, manages, and executes Callables through a given request.
     """
 
     def __init__(self):
--- a/src/snapred/backend/api/HookManager.py
+++ b/src/snapred/backend/api/HookManager.py
@@ -61,7 +61,7 @@ class HookManager:
                     raise TypeError(
                         f"Hook '{hookName}' must be a callable with one argument"
                         + " (the instance in which the callable is executed) plus"
-                        + " additional arguments from HookArgs."
+                        + " additional keyword arguments from HookArgs."
                     )
             self.hooks[hookName] = hookList
 
--- a/src/snapred/backend/api/HookManager.py
+++ b/src/snapred/backend/api/HookManager.py
@@ -1,4 +1,10 @@
+from typing import Dict, List
+
+from snapred.backend.dao.Hook import Hook
+from snapred.backend.log.logger import snapredLogger
 from snapred.meta.decorators.Singleton import Singleton
+
+logger = snapredLogger.getLogger(__name__)
 
 
 @Singleton
@@ -14,13 +20,28 @@ class HookManager:
         """
         Reset the HookManager, clearing all registered hooks.
         """
-        self.hooks = {}
-        self.executed_hooks = set()
+        self.hooks: Dict[str, List[Hook]] = {}
+        self.executedHooks: Dict[str, int] = {}
 
     def allHooksExecuted(self) -> bool:
-        return len(self.hooks) == len(self.executed_hooks)
+        return all(self.executedHooks.get(hookName, 0) == len(hookList) for hookName, hookList in self.hooks.items())
 
-    def register(self, hooks: dict[str, callable]):
+    def validateAllHooksExecuted(self):
+        """
+        Validate that all hooks have been executed.
+
+        :raises ValueError: If not all hooks have been executed.
+        """
+        logger.debug(f"Validating all hooks executed: {self.executedHooks}")
+
+        if not self.allHooksExecuted():
+            raise ValueError(
+                "Not all hooks were executed. "
+                + f"Executed: {self.executedHooks}, "
+                + f"Registered: {self.hooks.keys()}"
+            )
+
+    def register(self, hooks: Dict[str, List[Hook]]):
         """
         Register a set of hooks.
 
@@ -29,17 +50,20 @@ class HookManager:
         if hooks is None:
             return  # if no hooks are provided, just return
 
-        for hookName, hook in hooks.items():
-            # assert that the hook is a callable with one argument
-            if not callable(hook):
-                raise TypeError(f"Hook '{hookName}' must be a callable.")
-            numArgs = hook.__code__.co_argcount
-            if numArgs != 1:
-                raise TypeError(
-                    f"Hook '{hookName}' must be a callable with one argument,"
-                    + " the instance in which the callable is executed."
-                )
-            self.hooks[hookName] = hook
+        for hookName, hookList in hooks.items():
+            if len(hookList) == 0:
+                raise ValueError(f"Hook '{hookName}' must have at least one callable registered.")
+            for hook in hookList:
+                func, funcKwargs = hook.func, hook.kwargs  # unpack the hook from the package
+
+                numArgs = func.__code__.co_argcount
+                if numArgs != 1 + len(funcKwargs):
+                    raise TypeError(
+                        f"Hook '{hookName}' must be a callable with one argument"
+                        + " (the instance in which the callable is executed) plus"
+                        + " additional arguments from HookArgs."
+                    )
+            self.hooks[hookName] = hookList
 
     def execute(self, hookName: str, instance: object):
         """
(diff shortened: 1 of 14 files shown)


walshmm:
this needs to pack the args such that they can be persisted later
hooks should also allow the specification of a stack of hooks.

src/snapred/backend/api/HookManager.py
```diff
@@ -0,0 +1,86 @@
...
+from snapred.backend.log.logger import snapredLogger
+from snapred.meta.decorators.Singleton import Singleton
+
+logger = snapredLogger.getLogger(__name__)
+
+
+@Singleton
+class HookManager:
+    """
+    Registers, manages, and executes Callables though a given request.
```
ekapadi, line 13: typo: "through" <- "though"

src/snapred/backend/dao/Hook.py
```diff
@@ -0,0 +1,20 @@
+from typing import Any, Callable
+
+from pydantic import BaseModel, field_serializer
+from pydantic.json_schema import SkipJsonSchema
+
+
```
ekapadi, line 6: **Please explain how and why this would ever be serialized.**  When it's reloaded it's not going to have the correct `Callable` filled in.  Does this really need to be a `BaseModel` at all? -- Thanks!

src/snapred/backend/dao/indexing/Record.py
```diff
@@ -30,6 +31,8 @@ class Record(VersionedObject, extra="allow"):
     # TODO (complicated): add a validator to enforce this.
     calculationParameters: CalculationParameters
 
+    hooks: Dict[str, List[Hook]] | None = None
+
```
ekapadi, line 35: OK, here's where it would be serialized.  Should this actually be `hooks: Dict[str, List[str]]`.  That is, I think we _know_ it's going to only be an `str`, but we're pretending that that isn't the case.   :(

src/snapred/backend/dao/reduction/ReductionRecord.py
```diff
@@ -34,6 +35,7 @@ class ReductionRecord(BaseModel):
     workspaceNames: List[WorkspaceName]
 
     alternativeCalibrationFilePath: Optional[Path] = None
+    hooks: Dict[str, List[Hook]] | None = None
```
ekapadi, line 38: See prev comment...  And "ditto" for all `Record`s.

src/snapred/backend/api/HookManager.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def reset(self):
+        """
+        Reset the HookManager, clearing all registered hooks.
+        """
+        self.hooks: Dict[str, List[Hook]] = {}
+        self.executedHooks: Dict[str, int] = {}
+
+    def allHooksExecuted(self) -> bool:
+        return all(self.executedHooks.get(hookName, 0) == len(hookList) for hookName, hookList in self.hooks.items())
```
ekapadi, line 27: I assume this method is provided to facilitate testing.  If so, could you add a comment to that effect?

src/snapred/backend/api/HookManager.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def validateAllHooksExecuted(self):
```
ekapadi, line 29: "ditto": prev comment

tests/cis_tests/reduction_with_cheese_hooks.py
```diff
@@ -0,0 +1,67 @@
...
+useLiteMode=True
+
+
+reductionService = ReductionService()
+groups = reductionService.loadAllGroupings(runId, useLiteMode)
+# find column group
+columnGroup = next((group for group in groups["focusGroups"] if "column" in group.name.lower()), None)
+
+# NOTE: Be careful how you define hook methods.  They should not be class methods.
+def cheeseHook(self, cheeseMasks: list[str]):
```
ekapadi, line 23: Is this actually checked anywhere, for example:
```
from types import MethodType

@field_validator("func", mode="before")
@classmethod
def _validate_func(cls, v: Any) -> Any:
    if not isinstance(v, MethodType) or isinstance(v, classmethod):
        raise ValueError("A function used by a hook must be an Instance method, not a classmethod, or any other type.")
    return v

```

src/snapred/backend/api/HookManager.py
walshmm: I wouldnt say this is purely for testing, I just thought it would be cleaner to give this list comprehension a name.

src/snapred/backend/recipe/Recipe.py
```diff
@@ -160,8 +162,12 @@ def cook(self, ingredients: Ingredients, groceries: Dict[str, str]) -> bool:
         Main interface method for the recipe.
         Given the ingredients and groceries, it prepares, executes and returns the final workspace.
         """
+        recipeName = self.__class__.__name__
         self.prep(ingredients, groceries)
-        return self.execute()
+        self.hookManager.execute(f"Pre{recipeName}", self)
+        result = self.execute()
+        self.hookManager.execute(f"Post{recipeName}", self)
+        return result
```
ekapadi, line 170: If this is the _second_ time this recipe is called within the service method.  How does this work?  (Repeat of question from previous comment.)

src/snapred/backend/api/HookManager.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        Execute a registered hook with the given instance.
+
+        :param hookName: The name of the hook to execute.
+        :param instance: The instance to pass to the hook.
+        :raises KeyError: If the hook is not registered.
+        """
+
+        if hookName in self.hooks:
+            if self.executedHooks.get(hookName, 0) >= len(self.hooks[hookName]):
```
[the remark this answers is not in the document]
ekapadi: RE "2."  **I think I see how this case works**, you need to provide _two_ hooks in your hooks list at the `hookName` key.

src/snapred/backend/recipe/Recipe.py
ekapadi: Never mind, I see how it works!

tests/cis_tests/reduction_with_cheese_hooks.py
```diff
@@ -0,0 +1,71 @@
...
+from snapred.backend.dao.request.ReductionExportRequest import ReductionExportRequest
+from snapred.backend.dao.request.ReductionRequest import ReductionRequest
+from snapred.backend.service.ReductionService import ReductionService
+
+import time
+
+runId="59039"
+useLiteMode=True
+
+
```
ekapadi, line 16: I want to run this as part of the PR verification.  Do you have any input data?

src/snapred/backend/api/HookManager.py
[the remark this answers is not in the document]
walshmm: 1. This occurs to inform whoever is prototyping their hooks that they've set things up wrong and how to fix it.  Suppose you dont know the total number of hooks off the top of your head, or you include one too few hooks.
2. hooks operate as a queue, they would simply be additional hooks in the queue.

src/snapred/backend/dao/Hook.py
```diff
@@ -0,0 +1,47 @@
...
+from typing import Any
+
+from pydantic import BaseModel, Field, field_serializer
+from pydantic.json_schema import SkipJsonSchema
+
+
+class Hook(BaseModel, arbitrary_types_allowed=True):
+    p_func: SkipJsonSchema[FunctionType | str] = Field(
+        ...,
+        alias="func",
```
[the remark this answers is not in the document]
walshmm: it does not skip validation

src/snapred/backend/dao/indexing/Record.py
walshmm: https://github.com/neutrons/SNAPRed/pull/612/files#diff-eb1c9569656e402bd5981768461e959c033d79d6473407b43f2a4b240f650094R20-R30

If you access it as `func` when it is a `str` it tosses an exception, but you still have the available `funcStr` in case you want to reference it for some record keeping job.  This allows you to explicitly ask for what you want without breaking reading from disk.  These are saved under the record, so we would not want to raise when reading those.

ekapadi (approved):
I tested this with a modified version of the provided cis-test script:

- Verified that the use-as-expected hooks defined as `@staticmethod` are called at the correct point in the reduction process;
- Verified that the written `ReductionRecord` could be read back in successfully;
- Verified that if anyone attempts to use the `hooks` from the read-back record, they will see a useful exception message: "TypeError: This Hook instance is just a record, func no longer available." -- this is just fine;
- Verified that if a `@classmethod` is passed as a hook, an exception will be raised;
- Verified that if an \<instance method\> is passed as a hook, an exception will be raised.

Everything works as it should.  **I approve these changes.**

PR #620, opened 2025-07-27
no duplicate run numbers in reduction panel

ekapadi:
## Description of work

Do not allow duplicate run numbers to be entered in the reduction-request panel.

## Explanation of work

  * Changed the `List[str]` to a `Set[str]`;
  * Fixed several of the integration tests.  The `MockQMessageBox` seems to be somewhat _unstable_ with respect to where the mock is actually attached.  At  present, I don't understand why these changes were actually required to get the tests to work!

### Dev testing

1. Run "reduction" on next -- it should allow you to enter duplicate run numbers.  In which case it will actually reduce the run more than once.
2. Try the same thing on this branch.  Note that at present, there's no error message when a duplicate run is entered, but I thought it was sufficient to not accept the entry.
 
### CIS testing

As for "Dev testing".

## Link to EWM item

[EWM#12295](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20%28Change%20Management%29#action=com.ibm.team.workitem.viewWorkItem&id=12295)

### DIFF
--- a/src/snapred/backend/dao/calibration/CalibrationDefaultRecord.py
+++ b/src/snapred/backend/dao/calibration/CalibrationDefaultRecord.py
@@ -1,13 +1,13 @@
 from typing import Annotated, Dict, List
 
-from pydantic import Field
+from pydantic import ConfigDict, Field
 
 from snapred.backend.dao.calibration.Calibration import Calibration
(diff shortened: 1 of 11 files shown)


walshmm (approved):
works as expected :+1:

PR #623, opened 2025-07-30
ContinueWarning for DEFAULT_DIFFRACTION_CALIBRATION

ekapadi:
## Description of work
Previously the `ReductionWorkflow` only recognized the distinction between a _missing_ diffraction calibration, and any existing diffraction calibration including the default.  In reality, the workflow was never entered with an _actual_ missing diffraction calibration, as it always required the state to be initialized.  From the beginning, It had been intended that the default diffraction calibration be identified with the missing diffraction calibration case, but these issues were confused during the implementation.  With the changes of this PR, there is now only `MISSING_DIFFRACTION_CALIBRATION`, both in `ContinueWarning.Type`, and in the workspace-metadata markers.  Both the reduction and normalization workflows now warn the user that the diffraction calibration is _missing_, when only the default diffraction calibration exists.

## Explanation of work

**This PR includes the following changes:**
    
  * Equate `ContinueWarning.Type.MISSING_DIFFRACTION_CALIBRATION` with `ContinueWarning.Type.DEFAULT_DIFFRACTION_CALIBRATION`.  Remove alternate code paths, and metadata markers.  From this point on, only `MISSING_DIFFRACTION_CALIBRATION` should be used.
    
   * Add new exception checks: when called in any workflow, `getLatestApplicableCalibrationVersion` should never return `None`.  At present, it is required that the state be initialized, so if `None` is returned it indicates a developer error.  Since returning `None` actually has a valid meaning -- no diffraction calibration exists, even the default -- these checks are placed in the workflow validators, and not in the method itself.
    
  * Rework `ReductionService.validateReduction` so that it gives the appropriate warning when the calibration is missing -- it now warns the user that a default calibration will be used.
    
  * In `NormalizationService._validateDiffractionCalibration`, give the same warning.
    
  * For both workflows: update the metadata marking accordingly.
    
  * Make the continue-warning messages in normalization workflow and reduction workflow as similar as possible.
    
  * Cleanup the logic flow in `SNAPResponseHandler._handleContinueWarning` to use standard Qt practices.  Specifically, `QMessageBox.exec()` returns the _default_ codes, but for non-default buttons, the recommended practice is to test against the \<button instance\>, rather than the \<button text\>.  It would be preferable if somehow this case were moved to the `ReductionWorkflow`, but I did not attempt to do that.

**Misc. related defects fixed:**
    
  * 'MockQMessageBox': remove 'No valid FocusGroups were specified for mode: 'lite'' hack; rework to use regular expressions for expected 'text' and 'detailedText'.
    
  * snapred-data: add lite-mode groupings to the grouping maps -- for the moment, these are just the native-mode groupings, but with changed names.  If they are ever used, we will need to adjust the contents accordingly.
    
  * Fix intermittent failure in 'CreateArtificialNormalizationAlgo.py' (vanishing 'weakPtr' reference to workspace): 'Variable invalidated, data has been deleted.'.

### Dev testing

**[Compare the behavior on "next" and on this branch]:**

1. Move any existing calibration _state_ directory out of the way.
2. Initialize a state so that only the _default_ calibration is available.
3. Run a reduction using a run in this state.
4. In both the normalizaton and the reduction workflows:  while observing the sequence of `ContinueWarning` messages, step by step, continue until both a non-default calibration, and a _real_ normalization are available.  (Remember to set your "appliesTo" correctly when you save the calibrations!)

### CIS testing

As for "Dev testing".

## Link to EWM item

[EWM#12331](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=12331)

### DIFF
--- a/src/snapred/ui/workflow/ReductionWorkflow.py
+++ b/src/snapred/ui/workflow/ReductionWorkflow.py
@@ -499,6 +499,7 @@ class ReductionWorkflow(WorkflowImplementer):
                     "Currently, Artificial Normalization can only be performed on a "
                     "single run at a time.  Please clear your run list and try again."
                 )
+
             for runNumber in self.runNumbers:
                 # Set status to indicate that the artificial normalization is being calculated.
                 self.setStatus(ReductionStatus.CALCULATING_NORM)
--- a/pixi.lock
+++ b/pixi.lock
@@ -169,14 +169,14 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libarchive-3.8.1-gpl_h98cc613_100.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libasprintf-0.25.1-h3f43e3d_1.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libasprintf-devel-0.25.1-h3f43e3d_1.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/libblas-3.9.0-33_h59b9bed_openblas.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/libblas-3.9.0-34_h59b9bed_openblas.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libboost-1.88.0-h6c02f8c_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libboost-python-1.88.0-py311h5b7b71f_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libbrotlicommon-1.1.0-hb9d3cd8_3.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libbrotlidec-1.1.0-hb9d3cd8_3.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/libbrotlienc-1.1.0-hb9d3cd8_3.conda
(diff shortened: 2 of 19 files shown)


walshmm:
# Testing
## Reduction with only default diffcal
image
continue anyway message informs me I only have a default diffcal. Successfully completes.
## Normalization with only default diffcal
image
Norm workflow also warns that I am missing a diffcal and only default diffcal is available. Successfully completes.
## Reduction with normcal but no diffcal
image
Same as the norm workflow. Successfully completes.
## Normalization with real diffcal
Executes without ContinueAnyway 
## Reduction with Normalization and Calibration
Executes without ContinueAnway

# Result
This fulfills the story AC :+1: I will approve

walshmm (dismissed):

src/snapred/ui/workflow/ReductionWorkflow.py
```diff
@@ -494,7 +494,13 @@ def _triggerReduction(self, workflowPresenter: WorkflowPresenter):
         if ContinueWarning.Type.MISSING_NORMALIZATION in self.continueAnywayFlags and (
             ContinueWarning.Type.CONTINUE_WITHOUT_NORMALIZATION not in self.continueAnywayFlags
         ):
+            # *** DEBUG ***
+            print(f" runNumbers: {self.runNumbers}, flags: {self.continueAnywayFlags}")
+
             if len(self.runNumbers) > 1:
+                # *** DEBUG ***
+                print("*** HERE <***")
```
walshmm, line 502: remove?

PR #625, opened 2025-08-02
Normalization-calibration workflow: apply sample overrides

ekapadi:
## Description of work
In the normalization-calibration workflow: apply any sample overrides from the calibrant sample json.

## Explanation of work
There was an existing treatment of the application of sample overrides in the diffraction-calibration workflow.  The changes of this PR simply apply the analogous treatment within the normalization workflow.

### Dev testing

Compare with the behavior on "next":
  * In the normalization workflow, load a sample that has overrides, and note that they are actually applied (as the values shown in the tweak-peaks panel).

### CIS testing

As for "Dev testing".

## Link to EWM item

[EWM#12333](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=<12333>)


walshmm:
the dmin and dmax fields, while they do contain the override values now, seem to have been disabled.  I dont think this is correct.

walshmm (approved):
Looks good, overrides work and the forms properly reset

PR #626, opened 2025-08-04
populate instrument parameters: workspace groups

ekapadi:
## Description of work
Recently `ExperimentInfo.populateInstrumentParameters` was exposed in Mantid's python api.  The effect of calling this method is _essential_ when loading any workspace which includes an instrument.  By default, when loading a workspace Mantid initializes the instrument-parameter values by reading from a single \<parameter map\> string, which _historically_ has not been written at full floating-point precision.  On calling `populateInstrumentParameters` the instrument parameter values are recalculated from the PV-logs, and reapplied to the detectors, thereby making the detector _positions_ and _rotations_ consistent.

In SNAPRed, for workspaces loaded from nexus format, `populateInstrumentParameters` is called within `FetchGroceriesAlgorithm`.  However, for grouping workspaces and mask workspaces, the corresponding call occurs within `LoadCalibrationWorkspaces`.  Since `FetchGroceriesAlgorithm` is also used to load diagnostic workspace groups, it's important to _recursively_ call `populateInstrumentParameters` for each workspace in a group (that has a type for which the call applies).

Formerly, there was a fallback implementation accomplishing the same effect as `populateInstrumentParameters` in `PV_logs_util`.  When the method was then exposed in Mantid's Python API, the translation between the two implementations wasn't performed correctly, and `populateInstrumentParameters` ended up being called _directly_ on `WorkspaceGroup`, which has no such method.

## Explanation of work
When `FetchGroceriesAlgorithm` uses `LoadNexusProcessed` to load a \<workspace group\>, all of the workspaces within that group are checked, and `populateInstrumentParameters` is called as appropriate.  

### Dev testing

The analogous unit tests from `PV_logs_util` have been updated and applied to test the current changes.  Note that the fallback utility method has not been removed from the codebase -- it should still be useful for diagnostic applications. In order to verify that this defect has been fixed, it's sufficient to compare the behavior between "next" and this branch.  **On "next", at the calibration assessment panel, this defect is triggered when one attempts to load a diagnostic `WorkspaceGroup` for comparison.**

## Link to EWM item

[EWM#12569](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=12569)

### DIFF
--- a/src/snapred/backend/data/GroceryService.py
+++ b/src/snapred/backend/data/GroceryService.py
@@ -615,9 +615,9 @@ class GroceryService:
 
     def _fetchInstrumentDonor(self, runNumber: str, useLiteMode: bool) -> WorkspaceName:
         """
-        The grouping workspaces require an instrument definition, and do not have their own instrument definition
-        saved with them.  Therefore, when loading groupings, it is necessary to match them to
-        the proper instrument definition *with the proper instrument state params*.
+        Grouping and mask workspaces require an instrument definition, and do not have their own instrument definition
+        saved with them.  Therefore, for these workspace types, it is necessary to match them to
+        the proper instrument definition *with the proper instrument state parameters*.
         This uses the run number (and lite mode) to locate the proper state, and from that the proper instrument
         definition with instrument params for that state.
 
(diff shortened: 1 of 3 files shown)


walshmm:
# Testing

## Loading Previous Calibrations for Assessment on `next`
image
CTD as expected :heavy_check_mark: 


## Loading Previous Calibrations for Assessment on Branch
image
Success :heavy_check_mark: 

# Result

Looks to fix this issue, approved :+1:

walshmm (approved):

PR #676, opened 2026-04-27
Added ability to reduce legacy SNAP runs

dlcaballero16:
## Description of work

This adds the ability to reduce older SNAP experiment runs and to set PVs if not found in older runs.

## Explanation of work

LocalDataService was updated to look for legacy SNAP experiment paths. DetectorState was updated so that if a PV is missing in the experiment data, the user can manually set the PV in the SNAPInstPrm by setting a new "default" field. Also, if any required PVs are missing, all missing PVs are logged as an error to the user.

### Dev testing

See CIS testing

#### Setup
Make sure you copy what is currently in the instrument.calibration.home folder in analysis to make sure you are up to date. Mainly make sure that your SNAPInstPrm folder looks like the one on analysis and has the 4 version folders inside too. Then update your application.yml so that the instrument.minimumRunNumber is either 0 or 36992 for this test. 

For devs, it is recommended that you do not use lite mode as you will probably run into permissions errors when trying to create the lite version of the run.

#### Testing

Attempt to run DiffCal on run number 36992. If using lite mode, you will need to turn on Skip Pixel Cablibration or else you will run into the issue of too many pixels failing calibration. You should see an error about missing PVs; they should be BL3:Chop:Skf1:WavelengthUserReq and BL3:Det:TH:BL:Frequency, and it should be in one error message box. Now, go to the v_0003 folder in SNAPInstPrm and open the SNAPInstPrm.json. Look at stateIdSchema->properties and see that the PVs are listed there. Add a new default field to the 2 PVs and a value so they look like the following:

```json
"BL3:Chop:Skf1:WavelengthUserReq": {"type": "number", "resolution": 0.1, "alias": "WavelengthUserReq", "default": 2.1},
"BL3:Det:TH:BL:Frequency": {
                "type": "integer",
                "alias": "Frequency",
                "default": 60
            },
```
I recommend using those values as they worked for me. Now when you try to run diffcal on run 36992, you will be able to get to the tweak peaks tab. I recommend setting grouping to all so that you don't need to worry about too many bad peaks and not having enough peaks.

Now run normalization with 36996 and 36998 as the background run. This takes a little while as you can't use lite mode.

Finally run reduction with the following numbers: 37005, 37008, 37011, 37013, 37021, 37026. You will need to modify InstrumentParameterIndex.json in the SNAPInstPrm folder so that the "appliesTo" field for version 3 allows 37026.

## Link to EWM item

[EWM#15112](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=15112)

### Verification

### Acceptance Criteria

- [ ] acceptance criterion 1
- [ ] acceptance criterion 2

### DIFF
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -244,7 +244,9 @@ class LocalDataService:
             filePath = IPTS / (Config[pre] + str(runNumber) + Config[ext])
             if not filePath.exists():
                 if not useLiteMode:
-                    legacyPath = IPTS / ("data/SNAP_" + str(runNumber) + "_event.nxs")
+                    legacyPath = IPTS / (
+                        Config["nexus.legacy.prefix"] + str(runNumber) + Config["nexus.legacy.extension"]
+                    )
                     if legacyPath.exists():
                         filePath = legacyPath
                 else:
--- a/src/snapred/backend/data/LocalDataService.py
+++ b/src/snapred/backend/data/LocalDataService.py
@@ -243,9 +243,12 @@ class LocalDataService:
             ext = instr + ".extension"
             filePath = IPTS / (Config[pre] + str(runNumber) + Config[ext])
             if not filePath.exists():
-                legacyPath = IPTS / ("data/SNAP_" + str(runNumber) + "_event.nxs")
-                if legacyPath.exists():
-                    filePath = legacyPath
+                if not useLiteMode:
+                    legacyPath = IPTS / ("data/SNAP_" + str(runNumber) + "_event.nxs")
+                    if legacyPath.exists():
+                        filePath = legacyPath
+                else:
+                    filePath = None
         return filePath
 
     def stateExists(self, runId: str) -> bool:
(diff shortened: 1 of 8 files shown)


walshmm:
Managed to get all the way through reduction
image
Malcolm will need to tweak things so that he can get useful results out of this, but at least its possible now.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -242,6 +242,13 @@ def createNeutronFilePath(self, runNumber: str, useLiteMode: bool) -> Path | Non
             pre = instr + ".prefix"
             ext = instr + ".extension"
             filePath = IPTS / (Config[pre] + str(runNumber) + Config[ext])
+            if not filePath.exists():
+                if not useLiteMode:
+                    legacyPath = IPTS / ("data/SNAP_" + str(runNumber) + "_event.nxs")
```
walshmm: can we extract this out to the application.yml as a str template?

walshmm (commented):
I only had one comment.

walshmm (approved):

PR #679, opened 2026-05-07
Add filtering based on pixel grouping

dlcaballero16:
## Description of work

This adds the ability to filter out "bad" records meaning records that have pixel grouping that overlap East and West bank detectors. 

## Explanation of work

A new algorithm was added called DetectorPanelStraddleCheck that checks if a given GroupingWorkspace is "invalid" meaning that if any banks are both in the East and West panel. This saves on performance as opposed to checking each individual pixel. Two filter functions were then added to a new GroupingService where the inputs are a list of records, run number, if using liteMode or not, that then returns a list of invalid groupings for both calibration and normalization records. 

### Dev testing

See CIS testing.

#### Setup
Make sure to run a few diffcals and normalizations of both lite and non-lite mode and that have groupings of All and either Column or Bank. So at a minimum you should have 8 records. I tested using run 58882.

#### Testing
After generating all the records, run the new CIS test script, detector_panel_straddle_check.py. There are three user settings near the top: runNumber, useLiteMode, and recordType. Mess around with the inputs and make sure the outputs make sense. If you only generated the bare minimum, you should get output saying that the All grouping is invalid for whatever input you set for the script. It should only be one grouping since it only checks the lite mode setting and the specific record type.

## Link to EWM item

[EWM#16082](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=16082)

### Verification

### Acceptance Criteria

- [ ] acceptance criterion 1
- [ ] acceptance criterion 2


src/snapred/backend/data/DataFactoryService.py
```diff
@@ -162,6 +168,76 @@ def getCalibrationDataWorkspace(self, useLiteMode: bool, version: Version, name:
...
+
+            if checked[groupName]:
+                kept.append(record)
+            else:
+                logger.info(
+                    f"Excluding calibration record (run={record.runNumber}, version={record.version}) "
+                    f"because grouping '{groupName}' straddles East/West panels"
+                )
+
+        return kept
```
walshmm: I dont think the actual filtering should be implemented in this interface class.
This is really more something that should go in a Service.

It would also be good to store a cache of whether a group is valid or not, since there is a small number of groupings actually used.

src/snapred/backend/data/DataFactoryService.py
```diff
@@ -162,6 +168,76 @@ def getCalibrationDataWorkspace(self, useLiteMode: bool, version: Version, name:
...
+            if checked[groupName]:
+                kept.append(record)
+            else:
+                logger.info(
+                    f"Excluding normalization record (run={record.runNumber}, version={record.version}) "
+                    f"because grouping '{groupName}' straddles East/West panels"
+                )
+
+        return kept
+
```
walshmm: same as above, also these are very similar, is there a way the overlapping code can go in a helper?

walshmm: or if you make a dedicated service for this functionality, I guess it can go there.

src/snapred/backend/service/GroupingService.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def isGroupNonStraddling(self, groupName: str, runId: str, useLiteMode: bool) -> bool:
+        """
+        Return True if the named grouping does NOT straddle both the East and
+        West detector panels.  Results are cached by group name.
+        """
+        if groupName not in self._straddleCache:
+            item = GroceryListItem.builder().grouping(groupName).fromRun(runId).useLiteMode(useLiteMode).build()
+            wsName = self.groceryService.fetchGroupingDefinition(item)["workspace"]
+            straddlingPtr = self.mantidSnapper.DetectorPanelStraddleCheck(
```
walshmm, line 32: is it required to do it by inspecting a workspace or could we filter based on the values pull from pgp's?

walshmm (commented):
Image
Seems to function correctly

src/snapred/backend/service/GroupingService.py
walshmm: the souschef class already loads pixel group info this way, I guess they would pull from the same workspace cache.  I just wanna make sure this isnt redundant

dlcaballero16: It's not redundant, PGP's don't store any info about detector ID's or panel info

walshmm: alrighty, sounds good

walshmm (approved):

PR #682, opened 2026-06-08
Add better error handling

dlcaballero16:
## Description of work

Originally this story was to fix an issue with reducing livedata, but I was not able to recreate the defect. These changes should hopefully prevent the error from happening, and at the very least improve the error message that gets reported by SNAPRed. Also, a CIS script was added that will go through the calibration folder and add cycle info as needed. Cycle info is required for reduction, but if no cycle data is present, the reduction result will be marked as diagnostic. 

## Explanation of work

In LocalDataService.py, attempting to read live metadata is now run in a try/except block and an error message about missing PV info is reported. 
In StateValidationException.py, the actual exception string is added to the error message in the case that the exception type is not a FileNotFoundError or a PermissionError.
A check was added in the ReductionService so that if no cycle data is present, then a warning box pops up if the user wants to continue. If they chose to continue, then the resulting reduction gets marked as diagnostic.

### Dev testing

Make sure unit tests pass and also see CIS testing.

### CIS testing

Open mantid workbench and load the new update_SNAPInstPrm_cycle.py script. There are 3 user settings in the script defined near the top:
1. DRY_RUN - leave this set to true unless you actually want the script to update the files!
2. OVERWRITE_EXISTING - leave this to false unless you want to overwrite previous cycle data
3. CYCLE_METADATA - this is where you define the cycle data and how the SNAPInstPrm.json's get updated. It is a dictionary of dictionaries where each dictionary is keyed on the first run of the cycle, and the value is the rest of the info required to fill out a cycle object.

You should be able to run the script as is and see output showing how the SNAPInstPrm for each version folder found will be updated. Try adding more cycles to the CYCLE_METADATA and make sure the script still works.

To test the cycle continue warning, attempt to run reduction with no cycle data present. You should see a continue warning pop up saying you are missing cycle info. If you click continue and finish reduction, your reduction output workspaces should be marked as diagnostic. Now, try reduction again with valid cycle info, and you should not see a pop up about cycle info and now see that the workspace will not be marked as diagnostic (given that running reduction does not already lead to diagnostic output).

## Link to EWM item

[EWM#16396](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=16396)

### Verification

### Acceptance Criteria

- [ ] acceptance criterion 1
- [ ] acceptance criterion 2


src/snapred/backend/data/DataFactoryService.py
```diff
@@ -56,15 +59,20 @@ def getRunConfig(self, runId: str) -> RunConfig:
     def getInstrumentConfig(self, runId: str) -> InstrumentConfig:
         return self.lookupService.readInstrumentConfig(runId)
 
-    def getCycleID(self, runNumber: str) -> str:
+    def cycleInfoExists(self, runNumber: str) -> bool:
```
ekapadi, line 59: SOP:  these need to call `LocalDataService` methods.  They shouldn't be _ever_ fully implemented in `DataFactoryService`.   Sometimes the backend might need to access these methods directly, and in that case you now have no way to do that from `LocalDataService`.  Look at the other surrounding methods, and you'll see what the pattern is.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -1477,7 +1477,16 @@ def readRunMetadata(self, runNumber: str) -> RunMetadata:
                 raise  # the existing exception is sufficient
 
         if not success:
-            metadata = self.readLiveMetadata()
+            # No local PVFile for `runNumber`, but a live-data connection exists: fall back to the live run.
+            #   If the fallback itself fails, make clear that (1) there's no local data for the requested run,
```
ekapadi, line 1508: This took me a while to figure out, but in the end, I agree with what you've done.  Gosh all of this is complicated!  :)

src/snapred/backend/error/StateValidationException.py
```diff
@@ -26,7 +26,11 @@ def __init__(self, exception: Exception):
         if isinstance(exception, (FileNotFoundError, PermissionError)):
             self.message = f"The following error occurred: {exceptionStr}\n\n" + "Please contact your IS or CIS."
         else:
-            self.message = "Instrument State for given Run Number is invalid! (See logs for details.)"
+            self.message = (
+                "Instrument State for given Run Number is invalid!\n\n"
+                + f"{exceptionStr}\n\n"
+                + "(See logs for details.)"
```
ekapadi, line 32: This "see logs for details" statement needs to go away.  Can we just put all of the information here?  I'm especially fond of the case "see logs for details", **and then there's no additional information in the logs -- this is the usual case!  :(**

src/snapred/backend/service/ReductionService.py
```diff
@@ -250,6 +261,8 @@ def reduction(self, request: ReductionRequest):
         )
         isDiagnostic = workspaceMetadata.diffcalState != DiffcalStateMetadata.EXISTS
```
ekapadi: Can't all of this just be *one* short-circuit logical statement?  I'm not sure that's really better, but try it and see?

tests/cis_tests/util/migration/update_SNAPInstPrm_cycle.py
```diff
@@ -0,0 +1,119 @@
+# CIS script: add/update `cycle` info in every SNAPInstPrm.json under the SNAPInstPrm folder.
+#
+# Each instrument-parameter version lives in its own folder:
+#     <instrument.parameters.home>/v_XXXX/SNAPInstPrm.json
+# This walks those folders and, for each file, derives the cycle from the file's own
+# `indexEntry.appliesTo` run range:
```
ekapadi, line 6: Please justify why we didn't just extend / fix the original instrument-parameters migration script.  This *is* after all just an instrument-parameters migration?!

tests/unit/backend/data/test_DataFactoryService.py
```diff
@@ -149,21 +149,35 @@ def test_getCycleID(self):
         self.instance.lookupService.readInstrumentParameters.assert_called_once_with("200")
 
     def test_getCycleID_no_cycle(self):
+        # Missing cycle info is not an error: a fallback cycle ID is returned so reduction
```
ekapadi: Same issue as mentioned at `DataFactoryService`.  You are testing what should be `LocalDataService` methods.  For `DataFactoryService` these would be captured via the *generic* testing procedure, which automatically triggers the sub-call.  (Take a lookk and you'll see what I mean, although it's a bit abstract, and I don't think we use it consistently.)

tests/unit/backend/data/test_LocalDataService.py
```diff
@@ -631,6 +631,50 @@ def test_readRunMetadata_live_data_fallback_no_active_run(mockRunMetadata):
         actual = instance.readRunMetadata(runNumber)  # noqa: F841
 
 
+def test_readRunMetadata_other_file_not_found_reraises():
+    # A `FileNotFoundError` that isn't a "No PVFile exists" error should be re-raised as-is.
+    runNumber = "12345"
+    instance = LocalDataService()
+    instance._readPVFile = mock.Mock(side_effect=FileNotFoundError("some other file is missing"))
```
ekapadi, line 687: Wouldn't it be _incorrect_ to raise a `FileNotFoundError` when you're unable to construct the instrument-parameters due to an Indexer error?  This is an example of the exception-flow issue that I mentioned previously (in my slack message)!

tests/unit/meta/test_Decorators.py
```diff
@@ -88,11 +88,12 @@ def test_stateValidationExceptionWithInvalidState(mockLogger):  # noqa: ARG001
     testMessage = "Here I will tell you, the end user, the reason why the state is invalid."
     mock_exception = generateMockExceptionWithTraceback(RuntimeError, testMessage)
 
-    with pytest.raises(
-        StateValidationException, match=r"Instrument State for given Run Number is invalid! \(See logs for details\.\)"
-    ):
+    with pytest.raises(StateValidationException, match=r"Instrument State for given Run Number is invalid!") as excinfo:
```
ekapadi, line 91: YES!  Get rid of the stupid "See logs for details".  Could we get rid of it everywhere else as well?  :)

ekapadi (commented):
**So far, there are only some minor issues, with putting stuff in `DataFactoryService` that should actually be routed from `LocalDataService`.**  I also would like you to take a _deeper_ look at the exception flow in these two cases: (Indexer -> InstrumentParameters -> ...) and (\<live-data fallback\> -> PV-file -> \<run metadata\>, and etc.).  I still am not entirely satisfied that we've completely fixed the misleading messages issue.

**For "testing", it would be a huge help for you to please explain how a developer (not a CIS) should test this?**  I'm probably not going to run the migration script, so I'm at a bit of a loss.  I think for this, just explain what I need to do to set up my directories so that they will trigger or not-trigger these changes?

tests/cis_tests/util/migration/update_SNAPInstPrm_cycle.py
dlcaballero16: The migration script distributes central `*Index.json` entries into per-version folders, while this script really just adds/populates the cycle field in `SNAPInstPrm.json`. Also, the cycle script reads `indexEntry.appliesTo`, which only exists after the migration script runs.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -87,6 +87,9 @@ class LocalDataService:
     # -- in its use below, this is used to convert 1.0e-10 * m^2 to <time in microsecond>.
     CONVERSION_FACTOR = Config["constants.m2cm"] * PhysicalConstants.h / PhysicalConstants.NeutronMass
 
+    # Sentinel cycle ID used when cycle info is absent/invalid; reduction continues as "diagnostic".
+    FALLBACK_CYCLE_ID: str = "unknown"
```
ekapadi: I'm skeptical about using this as a sentinel:  is cycle-id really *any* *string*?  Shouldn't this be in a `class` declaration as part of a Pydantic model somewhere.   (That's usually how things work in SNAPRed!)

Please do a quick assessment:  I'm not asking for a huge code change to accomplish this, but it reminds me a lot of how the unit tests used to work:  e.g. `runNumber = 'dummy_run'` (i.e. completely invalid as an actual \<run number\>)>

src/snapred/backend/data/LocalDataService.py
```diff
@@ -294,7 +297,7 @@ def _readPVFile(self, runId: str):
 
     # NOTE `lru_cache` decorator needs to be on the outside
     @lru_cache
-    @ExceptionHandler(StateValidationException)
+    @ExceptionHandler(StateValidationException, passthrough=BUG_EXCEPTIONS)
```
ekapadi: If I'm understanding the meaning of the `passthrough` arg here: I think I want this to work the _other_ way.  That is, unless an exception class is _explicitly_ in a list-of-exceptions that we know about, it should not be converted to a `StateValidationException`.

src/snapred/backend/data/LocalDataService.py
```diff
@@ -634,6 +637,22 @@ def readInstrumentParameters(self, runNumber: str):
...
+    def cycleInfoExists(self, runNumber: str) -> bool:
+        # Cycle info is considered present only when a cycle is defined *and* the run
+        #   falls within it. Absent/incomplete/invalid cycle info is not an error: like a
+        #   missing calibration, reduction proceeds with output labelled "diagnostic".
+        instrumentConfig = self.readInstrumentParameters(runNumber)
+        if instrumentConfig.cycle is None:
+            return False
+        return int(runNumber) >= instrumentConfig.cycle.firstRun
+
+    def getCycleID(self, runNumber: str) -> str:
```
ekapadi: Why is \<cycle id\> a *string*?!

src/snapred/meta/decorators/ExceptionHandler.py
```diff
@@ -1,13 +1,27 @@
...
 
 logger = snapredLogger.getLogger(__name__)
 
 
+# Exceptions that indicate a programming error (a bug), not a recoverable domain condition.
+# Passing these as `ExceptionHandler(..., passthrough=BUG_EXCEPTIONS)` lets a genuine bug
+# propagate as itself instead of being mislabeled as a domain exception (e.g. a plain
+# `list.remove(x): x not in list` should not be reported as "Instrument State ... is invalid!").
+BUG_EXCEPTIONS: Tuple[Type[BaseException], ...] = (
+    TypeError,
```
ekapadi: Please *NO*!  Please do the _inverse_ of this.  That is: a list of `Exceptions` that *do* *not* indicate a programming error!

tests/cis_tests/util/migration/update_SNAPInstPrm_cycle.py
```diff
@@ -0,0 +1,119 @@
...
+print(f"SNAPInstPrm home: {home}")
+if not home.is_dir():
+    raise FileNotFoundError(f"SNAPInstPrm home directory does not exist: {home}")
+
+jsonFiles = sorted(home.glob("v_*/SNAPInstPrm.json"))
+if not jsonFiles:
+    raise FileNotFoundError(f"No 'v_*/SNAPInstPrm.json' files found under {home}")
+
+print(f"Found {len(jsonFiles)} SNAPInstPrm.json file(s).\n")
+
```
ekapadi, line 76: **CALIBRATION DATA IS BEING CHANGED.  WHERE IS MY BACKUP?!**  :)

ekapadi (commented):
Mostly nit-picky comments (except for the **BACKUP** comment).

src/snapred/backend/service/ReductionService.py
```diff
@@ -248,9 +259,12 @@ def reduction(self, request: ReductionRequest):
...
             groceries["inputWorkspace"]
         )
-        isDiagnostic = workspaceMetadata.diffcalState != DiffcalStateMetadata.EXISTS
-        isDiagnostic = isDiagnostic or workspaceMetadata.normalizationState != NormalizationStateMetadata.EXISTS
-        ingredients.isDiagnostic = isDiagnostic
+        # Absent/invalid cycle info forces diagnostic output, just like a missing calibration.
+        ingredients.isDiagnostic = (
+            workspaceMetadata.diffcalState != DiffcalStateMetadata.EXISTS
+            or workspaceMetadata.normalizationState != NormalizationStateMetadata.EXISTS
+            or not self.dataFactoryService.cycleInfoExists(request.runNumber)
```
ekapadi, line 266: **Notice the asymmetry here:**  does this mean that `CycleInfo` should actually be part of `WorkspaceMetadata`?  **Probably.**
Let's trace the structure and see, and if so, it's OK to leave this, but create a new *defect* (or story?) to pull `CycleInfo` into `WorkspaceMetadata`.  (I'm fine with this being "outside the scope" of this current PR!)

src/snapred/meta/decorators/ExceptionHandler.py
```diff
@@ -22,15 +35,43 @@ def extractTrueStacktrace() -> str:
     return "no exception has occurred"
 
 
-def ExceptionHandler(exceptionType: Type[Exception]):
+def ExceptionHandler(
+    exceptionType: Type[Exception],
+    convert: Tuple[Type[BaseException], ...] = (),
```
ekapadi: Consider: `rewrap` instead of `convert`?

src/snapred/meta/decorators/ExceptionHandler.py
```diff
@@ -1,13 +1,26 @@
...
+# programming bug or an unrelated failure). Passed as `ExceptionHandler(StateValidationException,
+# convert=STATE_EXCEPTIONS)`, these are the ONLY exceptions re-routed into the domain type; anything
+# else propagates as itself.  This way a plain bug (`list.remove(x): x not in list`, a `TypeError`,
+# ...) or an unrelated failure (e.g. a live-data read `RuntimeError`) is never mislabeled as
+# "Instrument State ... is invalid!".
+#
+# `OSError` is the umbrella for the file/IO conditions that mean "can't read the state data": it
+# covers `FileNotFoundError` and `PermissionError` (both subclasses, and special-cased in
+# `StateValidationException`) as well as the plain `OSError` that h5py raises on an unreadable file.
+STATE_EXCEPTIONS: Tuple[Type[BaseException], ...] = (OSError,)
```
ekapadi: Wait a second:  shouldn't there really be more exceptions here than `OSError`?!  You also shouldn't be using base classes here, when you *know* that specific exception types are actually what will be `rewrapped`.  So that means you need to list `FileNotFoundError` and `PermissionError` and etc. *explicitly*.  **There is no harm in missing an exception type, and not rewrapping it!**  As things stand, this is very little change from how things were before.

Not doing this explicit listing is exactly what went wrong before -- every exception raised just became a `StateValidationException`.

tests/cis_tests/util/migration/update_SNAPInstPrm_cycle.py
```diff
@@ -0,0 +1,136 @@
+# CIS script: add/update `cycle` info in every SNAPInstPrm.json under the SNAPInstPrm folder.
+#
+# Each instrument-parameter version lives in its own folder:
```
ekapadi, line 3: This goes in `tests/cis_tests/util/migration/...` !

ekapadi (commented):
**Thanks for all of the changes so far!**

**I need you to:**
1. List exceptions *explicitly* in the `convert` arg list (i.e. no base classes as catch-alls -- that defeats the purpose of this change).  Also, I suggest changing `convert` to `rewrap`?
2. Do one small structure analysis (use "opus") and then spin off another "refactoring story".  You'll see what the issue is in the comment.

src/snapred/backend/service/ReductionService.py
dlcaballero16: I will make a new story for this

ekapadi (approved):
There was a lot of back-and-forth in this review, in order to rework `ExceptionHandler` to be more _accurate_ in terms of its intended usage, and to not mask *necessary* exceptions.  The previous version would (basically, as applied with `StateValidationException`) rewrap any exception type that wasn't explicitly mentioned, and then present it as a `StateValidationException`, which is the _opposite_ of the required behavior, which should only present a very few exception types as `StateValidationException`.  I did not actually run the migration script, but I did carefully examine its structure.
I approve these changes.

PR #685, opened 2026-08-07
Fix live reduction when Mantid default facility is not SNS

darshdinger:
## Description of work

Live reduction failed outright for any user whose Mantid default instrument/facility was not
SNAP/SNS. The reported failure came through SNAPWrap on an analysis node:

```
Invalid value for property Instrument (string) from string "SNAP": When setting value of
property "Instrument": The value "SNAP" is not in the list of allowed values
snapred.backend.error.StateValidationException: Instrument State for given Run Number is invalid!
```

Mantid's live-data algorithms (`LoadLiveData`, `StartLiveData`, `MonitorLiveData`) build the
allowed-values list for their `Instrument` property from the **default facility**, restricted to
instruments that have a live-data listener. Under a non-SNS default that list comes back *empty*, so
`Instrument="SNAP"` is rejected before anything else happens.

SNAPRed already set the default facility — but only in `SNAPRedGUI.__init__`. Any consumer that
imports `snapred.backend` directly and never constructs the GUI (SNAPWrap, scripts, tests) inherited
the user's own facility setting. SNAPWrap never touches Mantid's `ConfigService` at all, and reaches
this through `DataFactoryService.constructStateId`, which is exactly the reported traceback.

This makes SNAPRed reduce SNAP data regardless of the user's Mantid configuration, without SNAPRed
permanently changing that configuration out from under them.

## Explanation of work

The fix has to work around two properties of Mantid's validator, both verified against 6.16:

1. **The allowed-values list is frozen when the algorithm is initialized** and is never re-evaluated.
   `AlgorithmManager.create` initializes, and a second `initialize()` is a no-op — so the facility
   must already be correct *at the point of creation*. Setting it later has no effect, and there is
   no Python-exposed way to replace the validator.
2. **Naming the listener explicitly doesn't help.** Setting `Listener`/`Address` with `Instrument`
   empty still resolves from the default instrument, failing with "Attempted to access live listener
   for `<instrument>` instrument, which has no listeners."

Two placements were considered and rejected. Setting the facility at **module load** would change it
for the whole workbench session, and Python has no dependable module unload, so there'd be no correct
point to restore it — quite apart from doing that work as an import side effect. Setting it **only at
GUI startup** is the status quo, and is what leaves every non-GUI entry point broken.

So the override is **scoped**: applied only for as long as it takes to create and run a live-data
algorithm, then reverted. New helper `snapred/meta/mantid/LiveDataFacility.py` wraps
`mantid.kernel.amend_config`, driven by `Config["liveData.facility.name"]` /
`Config["liveData.instrument.name"]` rather than hard-coded `"SNS"`/`"SNAP"`, so the
`TEST_LIVE`/`ADARA_FileReader` mock-listener setup keeps working.

It is applied centrally in `MantidSnapper.executeAlgorithm` rather than at the individual call sites.
That was the more obvious option, but it doesn't work: `MantidSnapper` defers execution to
`executeQueue()`, so wrapping `LocalDataService._readLiveData` and friends would cover enqueue and
miss the actual failure point. Centralizing also means one place instead of three, and it
automatically covers every live-data path.

`ConfigService` is a process-wide singleton, so while the scope is active the default facility is not
what the user set. Two consequences:

- The scope is entered **inside** the existing `MantidSnapper._liveDataLock`, which already serializes
  `LoadLiveData` and `LoadLiveDataInterval`. That keeps the interval in which another thread could
  observe the modified configuration as short as the live-data call itself.
- The scope is **non-reentrant** — nesting would run an outer call under an inner call's facility, so
  it raises `LiveDataFacilityReentranceError` instead.

Two details about that guard are worth calling out, since both are easy to get wrong:

``` python
# The check runs BEFORE the mutex is acquired.  Nesting live-data calls through MantidSnapper
# already deadlocks on _liveDataLock -- silently, with no diagnostic -- so a check placed inside
# the scope would never be reached.  Confirmed with a debugger stack dump.
if needsLiveDataFacility(name):
    assertNoLiveDataFacility(name)
```

It is also **conditional** on the algorithm needing the facility. An unconditional check breaks
working code: a live-data algorithm legitimately calls ordinary algorithms through `MantidSnapper`
from inside its own `PyExec` — `LoadLiveDataInterval` calls `CloneWorkspace`, `Plus`, `FilterByTime`
and `DeleteWorkspace`, all from inside the scope.

One side effect worth noting for review: algorithm creation moved from *before* the mutex to *inside*
it. That closes a window in which two threads could each construct a `LoadLiveData` — and so risk a
second listener — before either acquired the lock. The aggressive `_removeAlgorithm` cleanup is
preserved; it is skipped only when creation itself threw, i.e. when there is nothing to remove.

**No startup hook or consumer change is needed.** Because the scope is applied at the point of use,
there is nothing for SNAPWrap or any other consumer to call. This holds only as long as nothing
outside a `MantidSnapper` call depends on the Mantid *default* facility; the one such dependency,
`LoadLiveDataInterval.validateInputs`, runs during `execute` and is therefore already inside the
scope. Everything else resolves the facility or instrument *by name*. This assumption is written down
in the new implementation note, along with the fallback if it ever breaks.

Two smaller fixes ride along:

- `SNAPRedGUI` now saves and restores `default.instrument` alongside `default.facility`. Setting the
  facility alone never yields the right instrument: `setString` leaves the previous one in place,
  while `ConfigService.setFacility` resets it to the facility's *first* instrument.
- `CheckIPTS.getValidInstruments` compared an `InstrumentInfo` against the string `"DAS"`, which is
  never equal, so the exclusion silently did nothing. `"DAS"` matters precisely because it is what
  `ConfigService.setFacility("SNS")` selects as the default instrument.

New implementation note at `docs/source/developer/implementation_notes/mantid_config_ownership.rst`
documents who owns Mantid's `ConfigService`, why the change is scoped, and the validator behaviour
above — so the next person doesn't have to rediscover it.

### Dev testing

45 tests added across four files, plus 5 added to an existing one.

- **`tests/unit/backend/data/test_liveDataFacilityConfig.py`** (9) — characterizes the Mantid
  behaviour the fix depends on: the rejection itself under a non-SNS default; facility+instrument
  together making `Instrument="SNAP"` work; `amend_config` restoring the caller's settings;
  allowed-values being frozen at `initialize()`; and `getInstrument(name)` searching all facilities.
  Deliberately written to hold under any of the candidate fixes, so it doesn't encode this particular
  implementation.
- **`tests/unit/meta/mantid/test_LiveDataFacility.py`** (21) — the helper: applies both keys, restores
  them (including after an exception), is a no-op for non-live-data algorithms, is non-reentrant, is
  reusable after a failure, and permits ordinary algorithms to nest inside a live-data scope.
- **`tests/unit/backend/recipe/algorithm/test_MantidSnapper.py`** (+5) — that a live-data algorithm is
  *created* under SNAPRed's facility (the timing that matters), that the facility is restored
  afterwards and on failure, and that non-live-data algorithms don't touch the configuration at all.
- **`tests/unit/backend/recipe/algorithm/test_CheckIPTS.py`** (5) — `DAS` excluded, SNAP present,
  empty string still allowed.
- **`tests/unit/ui/test_mainLiveDataConfig.py`** (4) — the GUI save/override/restore, exercised
  against a stub so no Qt widgets are constructed.

I verified the two most important tests genuinely fail without their fix, rather than trusting that
they pass: the `MantidSnapper` creation-timing tests and the `CheckIPTS` `DAS` test.

**Things reviewers should look at closely:**

- `tests/unit/ui/test_mainLiveDataConfig.py` is `importorskip`-guarded, because `snapred.ui.main`
  needs the pinned Mantid version and my local env is behind (see below). It will **skip** locally and
  run in CI. I verified the underlying logic separately by shimming the one missing symbol, so I
  expect it to pass rather than merely skip — but CI is the first place it actually executes.
- My local `.pixi/envs/dev` is on Mantid 6.15 against the `>=6.16` pin, because `pixi` can't read
  lockfile v7 without an upgrade. Consequence: all of `tests/unit/ui` fails to collect locally, so
  the full suite was run as `pytest tests/unit --ignore=tests/unit/ui`:
  **1518 passed, 31 skipped, 1 failed**. The one failure
  (`test_RunMetadata::test_defaults_log_warning_fromNeXusLogs`) is pre-existing — confirmed by
  stashing on a clean tree.
- The re-entrance guard is a tripwire, not a fix for a live path. Searching the repo, there are only
  two live-data invocation paths and neither can reach the other through `MantidSnapper`
  (`LoadLiveDataInterval` creates its `LoadLiveData` child via `_createChildAlgorithm`). If that
  changes, the guard fires instead of deadlocking.

### CIS testing

The interesting case needs a real listener, so this is best run on an analysis node. It sets a
deliberately wrong default facility, exercises the path from the original traceback, and checks both
that the validator no longer rejects SNAP and that your own Mantid configuration is left untouched.

``` python
#
# EWM 15513: SNAPRed must reduce SNAP data even when the Mantid default instrument is not SNAP.
#
# Run from workbench.  Substitute a run number that exists for you.
#
from mantid.kernel import ConfigService

RUN_NUMBER = "58810"   # <-- substitute a run number available to you

FACILITY_KEY, INSTRUMENT_KEY = "default.facility", "default.instrument"
config = ConfigService.Instance()
saved = (config[FACILITY_KEY], config[INSTRUMENT_KEY])
print(f"your Mantid defaults: {saved[0]} / {saved[1]}")

try:
    # 1. Deliberately set a default facility that is NOT SNS.  Before this fix, this alone was
    #    enough to make live reduction fail.
    config.setString(FACILITY_KEY, "ILL")
    config.setString(INSTRUMENT_KEY, "IN5")
    print(f"forced defaults to: {config[FACILITY_KEY]} / {config[INSTRUMENT_KEY]}")

    # 2. Confirm the underlying Mantid validator really would reject SNAP right now.
    alg = AlgorithmFactory.create("LoadLiveData", -1)
    alg.initialize()
    print(f"LoadLiveData allowed instruments under ILL: {list(alg.getProperty('Instrument').allowedValues)}")

    # 3. Exercise the path from the reported traceback.
    #    EXPECTED: no "not in the list of allowed values" error.
    localDataService = LocalDataService()
    stateId, detectorState = localDataService.generateStateId(RUN_NUMBER)
    print(f"PASS: generateStateId('{RUN_NUMBER}') -> {stateId}")

    # 4. And the live-data metadata read itself, if a live run is active.
    if localDataService.hasLiveDataConnection():
        metadata = localDataService.readLiveMetadata()
        print(f"PASS: live run number = {metadata.runNumber}, active = {metadata.hasActiveRun()}")
    else:
        print("SKIP: no live-data connection from this node")

    # 5. SNAPRed must not have left its own facility behind.
    assert config[FACILITY_KEY] == "ILL", f"facility leaked: {config[FACILITY_KEY]}"
    assert config[INSTRUMENT_KEY] == "IN5", f"instrument leaked: {config[INSTRUMENT_KEY]}"
    print("PASS: SNAPRed restored the default facility and instrument")

finally:
    config.setString(FACILITY_KEY, saved[0])
    config.setString(INSTRUMENT_KEY, saved[1])
    print(f"restored your Mantid defaults: {config[FACILITY_KEY]} / {config[INSTRUMENT_KEY]}")
```

**Failure looks like:** step 3 raises `StateValidationException`, with
`The value "SNAP" is not in the list of allowed values` logged immediately above it.

**Success looks like:** `PASS:` on steps 3 and 5, and step 2 printing an empty allowed-instruments
list — which is the point: the list is empty, and SNAPRed works anyway.

**GUI check (optional).** Set a non-SNAP default instrument in *File → Settings → General* in
workbench, restart, then run a live reduction from the SNAPRed reduction panel. It should proceed
normally. Afterwards, confirm your default instrument in workbench settings is still what you set —
SNAPRed should not have changed it.

## Link to EWM item

[EWM#15513](https://ornlrse.clm.ibmcloud.com/ccm/web/projects/Neutron%20Data%20Project%20(Change%20Management)#action=com.ibm.team.workitem.viewWorkItem&id=15513)

### Verification

### Acceptance Criteria

This list is for ease of reference, and does not replace reading the EWM story as part of the review.
Verify this list matches the EWM story before reviewing.

- [ ] SNAPRed reduces SNAP data irrespective of the user's Mantid default instrument setting
- [ ] SNAPRed does not permanently alter the user's Mantid default facility or instrument

### Not covered by this PR

- **Why run 69405's PV file was missing** (EWM#17188). That absence is what routed the code into the
  live-data fallback in the first place, and it still wants an answer from the reporter/IS. This PR
  fixes the failure that resulted, not the missing file.
- **Upstream Mantid change.** The cleaner fix is a `Facility` argument on the live-data algorithms,
  with the check moved from a `StringListValidator` into `validateInputs` — that needs no
  configuration mutation at all and gives far better error messages. Kort and I prototyped and agreed
  on this; it is a separate PR against Mantid on a different timeline, and it deliberately does not
  touch `LoadLiveData` itself. Once it lands, the scoped override here can be retired.

### DIFF
--- a/docs/source/developer/implementation_notes/index.rst
+++ b/docs/source/developer/implementation_notes/index.rst
@@ -13,4 +13,5 @@ provide additional information that will be useful during SNAPRed development.
    :caption: Index
 
    grouping_workspaces.rst
+   mantid_config_ownership.rst
    profiling_and_progress_recording.rst
--- a/docs/source/developer/implementation_notes/mantid_config_ownership.rst
+++ b/docs/source/developer/implementation_notes/mantid_config_ownership.rst
@@ -1,0 +1,125 @@
+Mantid ``ConfigService`` ownership
+----------------------------------
+
+Mantid's ``ConfigService`` is a process-wide singleton.  When SNAPRed runs inside ``mantid_workbench``
+it therefore shares that configuration with the workbench itself and with anything else the user is
+doing in that session.  Any SNAPRed change to it is visible everywhere, immediately.
+
+For this reason SNAPRed treats the Mantid configuration as *borrowed*, never *owned*: every entry
+that SNAPRed modifies is saved, overridden, and then restored.
+
+Default facility and instrument
+===============================
+
+The awkward case is ``default.facility`` / ``default.instrument``.
+
+Mantid's live-data algorithms (``LoadLiveData``, ``StartLiveData``, ``MonitorLiveData``) build the
+allowed-values list for their ``Instrument`` property from the *default facility* -- that is, from
+``ConfigService.getFacility()`` with no argument -- restricted to those instruments which have a
+live-data listener.  If the default facility is not ``SNS``, that list comes back *empty*, and so
+``Instrument="SNAP"`` is rejected:
+
+.. code-block:: text
+
+   Invalid value for property Instrument (string) from string "SNAP": When setting value of
+   property "Instrument": The value "SNAP" is not in the list of allowed values
+
+SNAPRed consequently cannot reduce SNAP data at all when the user's Mantid default instrument or
+facility is set to anything else.
+
+Two properties of this validator constrain any workaround:
+
+1. **The allowed-values list is fixed at algorithm initialization** and never re-evaluated.
+   ``AlgorithmManager.create`` initializes the algorithm, so the default facility must *already* be
+   correct at the point of creation -- setting it afterwards has no effect.  This is also why simply
+   adding a ``Facility`` *property* to these algorithms would not help: property values are only set
+   after initialization, by which time the validator is frozen.  An upstream fix would instead have
+   to relax the ``StringListValidator`` on ``Instrument`` and perform the check in
+   ``validateInputs``, where both properties are available.
+
+2. **Naming the listener explicitly does not avoid the lookup.**  Setting ``Listener`` and
+   ``Address`` while leaving ``Instrument`` empty still resolves the listener from the default
+   instrument, and fails with "Attempted to access live listener for <instrument> instrument, which
+   has no listeners."
+
+Why the change is scoped
+========================
+
+Two placements were rejected:
+
+* **At SNAPRed module load.**  This would change the default facility for the entire workbench
+  session, and Python offers no dependable module unload, so there would be no correct point at
+  which to restore it.  Independently of that, doing resource-affecting work as an import side
+  effect is an antipattern.
+* **Only at GUI startup.**  This is what SNAPRed did historically (see
+  ``SNAPRedGUI._addLiveDataMantidConfigEntries``).  It leaves every non-GUI entry point broken:
+  consumers which import ``snapred.backend`` directly -- ``SNAPWrap``, scripts, tests -- never
+  construct the GUI, and so inherit the user's own facility setting.
+
+Instead, the default facility and instrument are overridden only for as long as it takes to create
+and run a live-data algorithm.  This is implemented by ``snapred.meta.mantid.LiveDataFacility`` and
+applied centrally in ``MantidSnapper.executeAlgorithm``, which is the single point through which all
+SNAPRed live-data algorithm calls pass.
+
+Because the override *is* still process-wide while it is in effect, it is deliberately entered
+*inside* ``MantidSnapper._liveDataLock`` -- the mutex which already serializes ``LoadLiveData`` and
+``LoadLiveDataInterval``.  This keeps the interval during which another thread could observe the
+modified configuration as short as the live-data call itself.
+
+The scope covers ``execute``, not merely initialization: ``LoadLiveData`` resolves its listener and
+address from the facility configuration at execution time, as can be observed from the resulting
+connection attempt to the SNAP listener address.
+
+No startup hook is required
+===========================
+
+Because the scope is applied at the point of use, there is deliberately *no* SNAPRed startup or
+shutdown call for a consumer to remember.  This holds only as long as nothing outside a
+``MantidSnapper`` algorithm call depends on the Mantid *default* facility or instrument.  At the time
+of writing, the only such dependency is ``LoadLiveDataInterval.validateInputs``, which runs during
+``execute`` and is therefore already inside the scope.  Everything else resolves the facility or
+instrument *by name* (see "Practical notes" below).
+
+If a future code path needs the default facility outside a live-data algorithm call, that assumption
+breaks, and an explicit start/shutdown pair -- which every consumer, including ``SNAPWrap``, would
+then have to call -- becomes the fallback.
+
+Non-reentrance
+==============
+
+While the scope is active the default facility is not what the user set, so nesting it would mean an
+outer live-data call silently running under an inner call's facility.  Re-entry is therefore treated
+as a defect: ``liveDataFacility`` raises ``LiveDataFacilityReentranceError`` rather than
+accommodating it.  Mantid runs algorithms one at a time, so there is no legitimate reason for these
+scopes to nest.
+
+Two details matter here:
+
+* **The check must happen before the live-data mutex is acquired.**  Nesting live-data calls through
+  ``MantidSnapper`` already deadlocks on ``_liveDataLock``, silently and with no diagnostic -- both
+  ``LoadLiveData`` and ``LoadLiveDataInterval`` map to the same non-reentrant lock.  A check placed
+  inside the scope would never be reached.
+* **The check must be conditional on the algorithm needing the facility.**  A live-data algorithm
+  legitimately calls *ordinary* algorithms through ``MantidSnapper`` from within its own ``PyExec``,
+  and hence from inside the scope: ``LoadLiveDataInterval`` calls ``CloneWorkspace``, ``Plus``,
+  ``FilterByTime`` and ``DeleteWorkspace``.  Rejecting those would break working code.
+
+Note also that algorithm *creation* now happens inside the mutex rather than before it.  This is
+deliberate: creating a second ``LoadLiveData`` instance while one is running risks creating a second
+listener, and the previous ordering left a window in which two threads could each construct one
+before either acquired the lock.
+
+Practical notes
+===============
+
+* Both keys must always be set together.  Setting the facility alone never yields the right
+  instrument: ``ConfigService.setString("default.facility", ...)`` leaves the *previous* instrument in
+  place, while ``ConfigService.setFacility(...)`` resets it to the facility's *first* instrument
+  (``DAS``, for SNS -- which is not a real instrument).
+* Look up instruments *by name* wherever possible.  ``ConfigService.getInstrument(<name>)`` and
+  ``ConfigService.getFacility(<name>)`` search all facilities and are unaffected by the default, as
+  used by ``LocalDataService.hasLiveDataConnection`` and ``CheckIPTS.getValidInstruments``.  Only the
+  property *validator* is facility-bound.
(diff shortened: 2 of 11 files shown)


ekapadi:
"""Naming the listener explicitly doesn't help. Setting Listener/Address with Instrument
empty still resolves from the default instrument, failing with "Attempted to access live listener
for <instrument> instrument, which has no listeners."
"""
Did you mean "setting address with instrument empty"?  Otherwise I'm not understanding what you mean here!  :)

ekapadi:
""" One side effect worth noting for review: algorithm creation moved from before the mutex to inside
it. That closes a window in which two threads could each construct a LoadLiveData — and so risk a
second listener — before either acquired the lock. """ -- I'm not sure about this change.  Previously the lock protected *execution*, not constructing listeners (basically: those are just socket connections).  So we need to be *very* careful about this!

ekapadi:
I'm very skeptical about these changes!  Arguably you've selected an optimal path, but I think we're adding lots of complexity, and future potential defects to work around Mantid flaws.  We should probably just fix the latter.  (And any users of SNAPwrap, in the mean time [i.e. waiting for the new Mantid version] can just set their facility!)

Here are what I think are the key points:

- No consumer algorithm in SNAPRed should be using the _default_ instrument;  In my own code changes, I've always tried to pass the instrument as an arg;
- Any algorithm that accepts an _instrument_ arg should also be accepting a _facility_ arg -- these two things just go together;  either one might be optional, but it should be possible to override both;
- We had discussed the issue of changing the validation in [those specific Mantid algorithms] so that it is _dynamic_ -- i.e. then this could be solved by just adding an optional _facility_ arg to the live-data algorithms;
- I don't like putting anything into `MantidSnapper`.  I also am "suspicious" about changes to the mutex -- it _might_ be OK, but ...

ekapadi:
I think I'm understanding the current state of the PR, and I like it much better.  I only really have two issues:

**1.** Since SNAPRed Algorithm are all "proposals" for general Mantid algorithms (i.e. at least that's how I view them); probably LoadLiveDataInterval's facility should be defaulting to the default Mantid facility.  Then the transfer from SNAPRed's Config[liveData.facility.name] would only happen when that param is set at the call.  That is, it's specific to how we use LoadLiveDataInterval in SNAPRed,  and then the algorithm itself becomes independent of SNAPRed specifically.  (And yes, we do have a lot of algorithms that depend implicitly on SNAPRed's Config, but we probably should not!)  If this change doesn't work for some reason, this is a fine point, so don't worry too much about it!   NOTE: this also means that I'm skeptical about what you're doing at 'CheckIPTS` -- that algorithm needs an optional "facility" arg.  ANY PLACE IN SNAPRED that we pass an instrument, now we should also be passing an optional facility -- this shouldn't be a big deal to fix!

**2.** [Mantid PR should be required?]: You should add the arg to LoadLiveData as part of this defect.   Then add a TODO (referring to the EWM item) to change the SNAPRed code once the Mantid PR is merged.  This then gets us (what I believe to be) the correct fix to this issue.  With respect to this issue: is LoadLiveData the only algorithm that we have this issue with?  For example, does GetIPTS need an optional facility arg as well?  Probably.  See "1" RE this issue.

Lastly, and a definite possibility that we should think about, might be that SNAPWrap itself should be called as a ConfigManager?  This might be important depending on how many settings from Mantid's Config we actually need.  If nothing else, it gives definite enter / exit hook behavior.  (Don't worry about this for this PR, but in thinking over all of this, this was the "other" solution I came up with.)

ekapadi (approved):
I have reviewed the changes, and performed the CIS test.  Everything works as it should.  I approve these changes.

**Note: please make sure to start the follow-on PR, to remove the "TODO" comments, and pass the 'facility' arg, once the mantid PR is landed!**

